PR #2, opened 2017-04-05
Script update

julien-tierny:
Dear Julien, could you add that in please?


julien-tierny:
Dear Julien, of course!

PR #4, opened 2017-04-06
Add missing cmake stuff

pavolzetor:


julien-tierny:
pavol,
I can't merge these requests without further explanations from you since ttk builds fine in Paris without these edits. The function calls you added should in principle be automatically pulled by ttk (see ttk.cmake lines 330 and 334-336).
please open an issue such that we can discuss this there. it seems to be that either your build chain is exotic or your removed something else somewhere else.

julien-tierny:
Closing this request without merge (see discussion in the corresponding issue).

PR #11, opened 2017-04-27
New TriangulationRequest paraview module and minor bug fix

GuillaumeFavelier:
Dear Julien, I think that it would be great if we had a TTK module able to display in ParaView different requests on the connectivity (e.g. simplex, facet, cofacet, star and link) of the dataset based on the ttk::Triangulation baseCode package? 

I also fixed a compilation warning related to a comparison of cells in the ttk::MorseSmaleComplex3D module.


julien-tierny:
Great stuff.

PR #16, opened 2017-05-04
Contour Forests Improvement V2

CharlesGueunet:
Dear Julien,
Now, the Contour Forests allows to compute one of the merge tree without computing the other.
Also I fixed a bug in the wrapper were the command line could only works on the first scalar field and removed warning at compile time.

Best regards,
Charles


julien-tierny:
Great stuff. Thanks.

PR #22, opened 2017-06-16
Patches for ParaView 5.4.0

jalevine:
Updated the patch to work with ParaView 5.4.0.  Not fully tested yet, but appears to work on OSX.


CharlesGueunet:
Also seems to work on Linux.

julien-tierny:
Great stuff.
May that be the first of many :)

PR #30, opened 2017-07-18
MorseSmaleComplex - smoothing on 1-separatrices

GuillaumeFavelier:
Dear Julien,

I added a SmoothingMask scalar field on the points of the 1-separatrices of the MorseSmaleComplex so that it can be smoothed by the GeometrySmoother module.

I also ran all the ttk-data tests and everything works fine.


julien-tierny:
Awesome!

PR #31, opened 2017-07-19
Discrete gradient  - dynamic removal method

GuillaumeFavelier:
Dear Julien,

I modified the simplification method in DiscreteGradient baseCode module so that the pairs of critical points can be simplified dynamically.


julien-tierny:
Great stuff!

PR #33, opened 2017-07-20
Persistence Diagram - persistence pairs inside domain

GuillaumeFavelier:
Dear Julien,

I added an option in ParaView to show the persistence pairs directly inside the input domain. I also added a field named 'Coordinate' which stores the coordinates of the vertices in the pairs, it is another way to show the pairs inside the domain thanks to the ParaView Calculator filter.

Guillaume


julien-tierny:
Awesome. Thanks.

PR #37, opened 2017-07-27
Scalar fields

GuillaumeFavelier:
Dear Julien,

I added new scalar fields on the critical points of the MorseSmaleComplex module : 
- ManifoldSize corresponds to the number of vertices in the basins/mountains originating at the extrema
- PLVertexIdentifier contains the vertex identifiers associated to the critical points


julien-tierny:
Great stuff. Thanks.

PR #38, opened 2017-07-27
Rename a scalar field of the MorseSmaleComplex module

GuillaumeFavelier:
Dear Julien,

I changed the name of the scalar field on the critical points of the MorseSmaleComplex module so that it will be compliant to PersistenceDiagram.


julien-tierny:
Thanks. Appreciated.

PR #39, opened 2017-07-27
DistanceField - bugfix

GuillaumeFavelier:
Dear Julien,

I fixed an inconsistency in the seed identifiers given by the DistanceField module. I'll send a pull request updating the ttk-data states.


julien-tierny:
Great thanks!

PR #40, opened 2017-07-28
Fix re-computation of persistence pairs

GuillaumeFavelier:
Dear Julien,

I fixed the bug in PersistenceDiagram that causes the re-computation of the whole module when the state of the checkbox 'Embed in Domain' is changed.


julien-tierny:
Great, thanks.

PR #43, opened 2017-07-29
Distance : fix abs for GCC >= 6

CharlesGueunet:
Dear Julien,

Here is a Fix for the new distances plugins as the did not compile to GCC >= 6 due to issue  [#2192](http://cplusplus.github.io/LWG/lwg-defects.html#2192) on these compilers.
GCC does not want to compile an std::abs on unsigned type, which is problematic when used with templates like here.
So i have re-implemented and used a naive version of the abs function.


julien-tierny:
Awesome. Thanks.

PR #44, opened 2017-08-01
Remove bad references to Contour Forests

CharlesGueunet:
Dear Julien,

These two files contained unwanted references to the ContourForests plugin that needed to be removed.

### BUG:

The *core/baseCode/bottleneckDistance/DataTypes.h* file is a copy of the *core/baseCode/contourForestsTree/DataType.h* with the same sentinels. This can leads to annoying bugs as described on the former file.


julien-tierny:
OK.
Let's discuss the overall typing thing when I'm back.

PR #45, opened 2017-08-04
MorseSmaleComplex2D - bugfix

GuillaumeFavelier:
Dear Julien,

I fixed a bug in the MorseSmaleComplex2D baseCode module that causes some 1-separatrices created from 1-saddles on the boundary to be inconsistent.


julien-tierny:
Cool. Thanks.

PR #46, opened 2017-08-23
fix(Cmake): for version btwn 2.8 and 3.1

CharlesGueunet:
Dear TTK Team,
this pull request is a fix for CMake.

On the "_CMakeLists.txt_", the **cmake\_minimum\_required** is set to 2.8,
but the "_baseCode.cmake_" uses the policy [CMP0054](https://cmake.org/cmake/help/v3.1/policy/CMP0054.html) introduced in cmake v3.1.
Here is a check to avoid using this policy for earlier version of CMake.

NOTE: as this policy is set to **NEW**, undefined behavior may appear for cmake < 3.1 if
this policy is indeed used.

Charles


julien-tierny:
Thanks!

PR #47, opened 2017-09-05
Improve CMake Build System to build base code and vtk wrapper libraries

Twinklebear:
This fixes the pain points of using TTK as a normal CMake library. Instead of going through the wrappers and having to build the TTK source with the project and jump through hoops to get it working nicely with other libraries or build system (like I did for topo-vol) TTK will now compile into a set of base code libraries and vtk wrapper libraries which can be easily found and used from CMake. For example, I've updated topo-vol's [CMakeLists.txt](https://github.com/Twinklebear/topo-vol/blob/master/topo-volume/CMakeLists.txt) to go through this new build path using my fork.

We can also migrate TTK to build and link like this for the standalone apps as well, as I've started to do for some of them.

**However** I'm having difficulty getting the ParaView plugins to work migrating to this build system. What I'd like to do is have the ParaView plugin be some source file (potentially empty) that does whatever possible extra junk is needed and that simply links in the existing TTK filters from the VTK wrappers libraries. It seems like when I do this though, ParaView can't find any of the TTK filters to run when you try to apply them, and poking around with readelf I don't see any `ttkContourForests` or such filter names in the paraview plugin library. I started trying to do this in the [ContourForests plugin](https://github.com/Twinklebear/ttk/tree/master/paraview/ContourForests) to test, but haven't gotten it working properly. This is likely fixable with some build system work, we need to get the filter symbols re-exported in the paraview plugin library that gets compiled. *Edit:* This is now resolved

@julien-tierny if you have some tips on the ParaView stuff it'd be helpful, or if you want to continue then let me know and you can merge the WIP PR. I was hoping to get this done for VIS quickly to help the library be easier to use for standalone applications but I don't know how much time I'll have.


Twinklebear:
Actually I have a different thought on implementing the paraview plugins that should work.

Twinklebear:
@julien-tierny I've actually got this working now, and can build and use the ParaView plugins. I've done this by merging the paraview plugin XML and building into the vtkWrappers directories, for an example of how this works see the [ContourForests wrapper & plugin](https://github.com/Twinklebear/ttk/tree/master/core/vtkWrappers/ttkContourForests)

julien-tierny:
hi will, 
this looks fantastic.
I've got this proposal due on Friday but I'll sure have a look at it over the week end.

quick question:
have you tried the bash scripts located in the "scripts/" directory? 
these create (or delete) a new TTK module from scratch and also enable users to release a buildable tarball of their module with the right amount of TTK dependencies. 
is that still working? (that was one of the reasons we had this 'original' cmake structure).
thanks for letting me know.
and thanks a lot for your work and help. much appreciated!

Twinklebear:
The scripts would likely be broken by these changes, I'll look at updating them as well.

Twinklebear:
@julien-tierny I've updated the create/delete module scripts, however I think with the improved build system in place the release on may not be necessary. The user can do a `make install` of ttk and package up the libraries as desired from the install.

I've also added the doxygen generation to be runnable in CMake, now if doxygen is found you can run `make doc` to build documentation.

julien-tierny:
hi will,

I've just had a look. this is great stuff. 
I'll probably tweak a few minor things afterwards, but overall this is looking great.
I think people will appreciate that a lot.

I ran into several issues though, that I believe we need to fix before we put that in. 

1) I applied TTK's standard installation procedure (https://topology-tool-kit.github.io/installation.html). I enabled your new cmake flags TTK_BUILD_PARAVIEW_PLUGINS, TTK_BUILD_STANDALONE_APPS and TTK_BUILD_VTK_WRAPPERS

everything seemed to build properly. however the "make install" does not seem to be fully functional.

a) I couldn't find where the standalone binaries were installed (I have the impression they are not). /usr/local/bin looks like a reasonable default location (current TTK's default)

b) on my test system here, TTK plugins for ParaView got installed in /usr/local/plugin while cmake should automatically retrieve the proper directory for ParaView's installation (on my test system: /usr/loca/lib/paraview-5.4/plugins/). 
ParaView need these plugins there in order to automatically load them at startup. because of this, the builtin examples are broken (section 7 of the installation instructions: https://topology-tool-kit.github.io/installation.html) 

you may want to test that out in a virtual environment to make sure the installation procedure works fine (I use a ubuntu 17.04 virtualbox image). 

you may want to have a look as well at the current cmake setup of TTK, in particular at how the following variables are set: TTK_BINARY_INSTALL_DIR and TTK_PLUGIN_INSTALL_DIR

2) it seems that the installation procedure installs static libraries (*.a files). 
it seems like it'd be better to switch to dynamically libs (*so files). 
that should be a tiny fix in cmake.

3) in the current version of TTK, baseCode.cmake setups a number of low-level compiler flags (Kamikaze build options, OpenMP, -03, -native, etc.). I don't have the impression that these are accessible anymore from cmake-gui. that'd be great if that could be fixed.

if you think you know where to fix these things, please feel free to go ahead.
on my side of things, I'll try to have a look at that this week end (that'd be great if you could add me as a collaborator on that branch such that I can push things).

but again, thanks for the effort. this is much appreciated and that'll be timely indeed for vis.

cheers,

julien-tierny:
also, unless you're strongly against it, I think I will move back all of your ParaView plugin CMake code (currently under vtkWrappers) back to the paraview plugin (along withe corresponding XML files).

julien-tierny:
edit: paraview paths look to be fixed now.

Twinklebear:
Cool, thanks Julien! I wasn't sure what the standard install location was for ParaView plugins so I was running paraview and setting the `PV_PLUGIN_PATH` to load them.

For your other comments:

> a) I couldn't find where the standalone binaries were installed (I have the impression they are not). /usr/local/bin looks like a reasonable default location (current TTK's default)

They aren't installed, I'll switch to install them. I wasn't sure if they were sample programs that you'd have for developers to use and play with (and thus shouldn't be installed) or actual utilities as part of TTK (and thus should be installed).

> It seems that the installation procedure installs static libraries (*.a files). it seems like it'd be better to switch to dynamically libs (*so files). that should be a tiny fix in cmake.

Yep, will do.

> in the current version of TTK, baseCode.cmake setups a number of low-level compiler flags (Kamikaze build options, OpenMP, -03, -native, etc.). I don't have the impression that these are accessible anymore from cmake-gui. that'd be great if that could be fixed.

I most likely missed these flags, I'll go through and check. The optimization flags should be added by building with `CMAKE_BUILD_MODE=Release`, though it won't setup Kamikaze specific things, OpenMP or native optimizations.

> also, unless you're strongly against it, I think I will move back all of your ParaView plugin CMake code (currently under vtkWrappers) back to the paraview directory (along with the corresponding XML files).

My initial plan was to keep them separate as well, and have the ParaView plugin just link the built vtkWrapper providing the filter; however, in this case the plugin library doesn't re-export the symbols from the vtkWrapper so ParaView is unable to find the TTK filters at runtime. I'm not sure if there's a better way or some linker flags to re-export stuff, let me know. For now I merged them so we would just have a the `ttk_build_vtk_wrapper` also build the ParaView plugin from the same source files so the symbols are exported.

julien-tierny:
hi will,
I added a todo file with some notes.

it seems there's been a regression since yesterday and things don't link anymore (I suspect something's wrong with the dynamic linkage).
builtin examples in paraview are currently broken, ttk apps too.
I'll have a look at that on Monday.

cheers

Twinklebear:
The todo file looks good. The `ParaView_DIR` and other `_DIR` are some CMake built-in stuff to point it to the corresponding libraries `XConfig.cmake`, but we can display those in the ccmake if we want.

What link error are you running into? I was having an issue with OpenMP but I think that's fixed in b015646

Twinklebear:
What might also be cool to look at (I'll put on the TODO with a question mark) is [Travis CI](https://travis-ci.org/). It's free for open source projects, so we could use it to test building the code on supported platforms (Linux, Mac, Windows(?))

Twinklebear:
Ah to clarify, by link do you mean at runtime the apps don't work? You may need to set the `LD_LIBRARY_PATH` to add the ttk libraries to your path. I did move where they would install to, instead of installing to `lib/ttk/base` and `lib/ttk/vtk` they both install to just `lib/ttk/`

julien-tierny:
hi will,

regarding the linking error, it's indeed an ld issue. 
I wish users could install TTK without having to run ldconfig and edit system config files (as it is the case with the current main branch).

there's several solutions to that issue:
1) we deliver both dynamic and static libs (and link the apps and the plugins to the static version). 
that's kind of unfortunate but I found that most unix libs tend to have this dual delivery strategy anyways.

2) we further tweak cmake to try to setup ldconfig automatically (I assume that should be possible)

regarding option 2, I have been playing with cmake's rpath variable (https://cmake.org/Wiki/CMake_RPATH_handling) but with no luck so far.

what do you think?
cheers

Twinklebear:
For 1, I thought it'd be nice to install the libraries under some prefixed path, but if we install them to where `LD_LIBRARY_PATH` is already looking then it should work properly. Currently they're put under `lib/ttk/` which may not be picked up by default when looking for libraries. I think switching back to installing them to the same library path as master would be best, since user's won't need to tweak anything.

Twinklebear:
I can take a look at exporting the bundle targets mentioned in https://github.com/Twinklebear/topo-vol/commit/7c7f0fcd813f1a21d77e23ad119651908e654344 if that sounds good as well.

julien-tierny:
hi will,
thanks for the bundle.

I just committed some fixes regarding the RPATH variable. it seems to work now: standalone apps (and ParaView plugins) find automatically their way to the shared libraries without having to configure the system linker. that makes end users' life easier.

however, there's still something off regarding how find_package(VTK) and find_package(ParaView) are handled. right now, either (option A) everything links against the system's VTK, or, if the build of the ParaView plugins is enabled (option B), everything links against the ParaView's VTK, which comes with several issues.
for instance, in option B, standalone apps try (and currently fail) at linking against some ParaView's shared libs while there's no reason for such links.
do you have any idea on how to fix this?

I have the impression that, for ParaView plugins, a static version of TTK should be built independently of the rest, and linked against ParaView's VTK (and extras), in order to build each TTK/ParaView plugin.
I believe that would mimic what TTK's master branch is currently doing (right now, the TTK's master links standalone apps against the system's VTK and ParaView plugins against the ParaView's VTK).

what do you think?

Twinklebear:
I agree, when building TTK with ParaView and then using the VTK wrappers in an app that doesn't link ParaView, you're then required to link say libQt or such, which should not be necessary. The VTK wrappers should just depend on VTK, not the version bundled with ParaView.

It might be possible to just link the vtk wrappers against regular VTK found with find VTK and ParaView against ParaView's VTK, but I noticed the ParaView cmake sets `${VTK_LIBRARIES}` as well. I'm not sure if this would conflict, we could work around it I think by saving the `${VTK_LIBRARIES}` gotten when finding VTK first, then letting ParaView potentially trample the `${VTK_LIBRARIES}` variable.

Twinklebear:
I'm looking into this a bit now, and it looks like it may be a bit trickier than I initially though. VTK incorrectly uses the `link_directories` command instead of absolute paths for their libraries.

Twinklebear:
Seems like the easiest solution is just to re-split the ParaView plugins and VTK wrappers, since the use files set link directories and other directory-scoped CMake vars that conflict with each other. I've done this in eaa3998

Twinklebear:
After playing with some tweaks I found online for trying to speed up the travis build I can't seem to get it under the time limit still. It looks like they migrated their host, and as a result all builds are significantly slower. Here (https://github.com/snowleopard/hadrian/issues/229) they mentioned switching back to the VM infrastructure, though I didn't seem to have much luck. Another option we could try to still have CI for the VTK and ParaView builds is to compile them in an Ubuntu 14.04 environment that matches the Travis one, and then download the binaries when we build.

Twinklebear:
I'm pushing a commit that merges in the changes added to include the FTM tree and fixes the merge conflicts. It seems like the repo has gotten a lot bigger after this merge though, was some binary data accidentally committed? A fresh git clone of the TTK repo is 131MB, which seems kind of big? A clone with `--depth is 30MB`. If some binary or other stuff was pushed and then removed it can be purged from the repo history using [bfg](https://rtyley.github.io/bfg-repo-cleaner/) to lighten up the repo.

Twinklebear:
dfb21b1 should now fix the segfaults in the paraview plugins, we would run into issues where the `ttkTriangulation` library was linked against the system VTK, but the paraview plugins link against paraview's included VTK, so we would be using the wrong libraries at some point in the call chain.

julien-tierny:
hi will, 
I confirm things build and run fine here too.
great effort, awesome outcome!

I'll double check things carefully over the week end but things look in a pretty good shape to me now and nearly ready to ship.

quick question, it seems that vtkWrappers are now installed as *.so files and baseCode libs as *.a files.
am I correct? does it mean that apps using the baseCode only have to link statically for now?

second quick question, is the main CMakeLists.txt of topo-vol updated to this branch of TTK?

cheers,

Twinklebear:
I switched the base code libraries over to static ones so the paraview plugins wouldn't run into RPATH or library path issues. We could probably switch the vtkwrappers to be static libs as well actually.

The root cmakelists should be up to date, [here's the root one on my repo](https://github.com/Twinklebear/ttk/blob/master/CMakeLists.txt)

julien-tierny:
I guess it's fine for now. I was just curious.
I'll let you know how my pass goes this week end.
cheers

julien-tierny:
hi will, 
I've been pretty optimistic on this last commit regarding your cmake bundles ("Automatically generated bundles", 0211d49). thanks for rolling it back if it breaks topo-vol's build.
I figured it'd be easier to maintain if the list of libs in the bundle would be automatically generated from the source tree layout.

julien-tierny:
alright will, things look ready to ship for me now (I'll resolve conflicts this afternoon).
there's still one thing I'm kind of not satisfied with. it's the module creation script.

right now, if developers want to create a new module, they'll have to call the appropriate scripts from the main ttk source tree (which is OK so far). however, if they want to release it, their only option is to ship the entire ttk source tree altogether, which does not look sustainable in the long run.
the current ttk master is slightly better in that regard (the releaseTTKmodule script will only add in the release source tree the required code dependencies).

I've been playing with the idea of a module creation script that would work outside of the main ttk source tree, that could be called anywhere and create a small source tree in the local directory, only with the new module.
for this, I first tried to create by hand a module outside of the main TTK source tree and things become difficult when it comes to ParaView plugins (ttkTriangulation is statically built and linked against those). 

any thoughts on how to make such an off-source-tree script work?
cheers

julien-tierny:
alright will, that's merged in.
thanks again and congrats for this awesome work.

next step: support for ms-winpoop

(if you have some ideas about ouf-of-tree createTTKmodule, let me know)

cheers

Twinklebear:
0211d49 looks great, and the rest of the changes. It's great to have it merged in!

I was thinking, for bundling packages we could probably do something with cmake, since we can get a list of all the targets dependencies (libraries and headers) through the different base code or vtkWrappers it links. Maybe with a cmake custom command, like `cmake bundle ttkMyCoolThing` which would then create some subdirectory with all the headers and libraries needed for `ttkMyCoolThing`.

I haven't tested, but I wonder if you just `make install` a specific target if you'd get that behavior (along with a reduced list of exports in the TTKConfig.cmake and TTKVTKConfig.cmake)?

PR #49, opened 2017-10-09
Handle NaN in FTM Tree

CharlesGueunet:
Dear Julien,

FTM is now able to deal with NaN values on data set where the scalar type can represent such a value.
In that case, a parallel pass in done in the base code to replace all NaN by a 0 value, as Paraview does in the GUI.


julien-tierny:
Sweet. Thanks.

PR #50, opened 2017-10-12
FTM Tree Improvements

CharlesGueunet:
Dear Julien,

This pull request brings several improvements to the FTM Tree computation:
* A bug fix for the segmentation where a vertex might be missing in rare cases
* Ability to add the region size/span to the node (can be used as a new measure for the simplification)
* GUI Change: add region size/span in Advanced stats checkbox + put Normalized Id in output
* Wrapper: memory and computation improvement: one alloc for each array + parallelism
* Wrapper: Architecture improvement: data in structure

Charles


julien-tierny:
Sweet. Thanks!

PR #51, opened 2017-10-16
Update Blank.xml

jalevine:
DoubleArgument should be specified as 'DoubleVectorProperty' and 'DoubleRangeDomain' to pass the type correctly.


julien-tierny:
weird, it's been there the whole time.
thanks for catching that up!

PR #52, opened 2017-10-18
Filter saddle-connectors in the Morse-Smale Complex module

GuillaumeFavelier:
Dear Julien,

I added a way to apply a threshold on the saddle-connectors by their persistence values as a post-process pass on the gradient.


julien-tierny:
Awesome. Thanks.

PR #53, opened 2017-10-19
Fix the smoothingMask values of the saddle-connectors

GuillaumeFavelier:
Dear Julien,

I fixed the values of the smoothingMask scalar field on the saddle-connectors of the MorseSmaleComplex module.


julien-tierny:
Cool. Thanks.

PR #55, opened 2017-10-20
Scalar field on arc skeleton for FTM

CharlesGueunet:
Dear Julien,
This pull request allows the user to access the scalar field used for the tree computation on the resulting arc skeleton. 

Charles


julien-tierny:
Thanks, that's gonna be useful for Jonas.

PR #57, opened 2017-10-24
Adapt FTM To deal with multiple Connected Components data sets

CharlesGueunet:
Dear julien,

I have adapted the Wrapper of the FTM Tree algorithm to works on disconnected data sets.
It now compute one tree by connected component and reconstruct the output with consistent identifiers on the global data set.

(also a minor change on how FTM deal with the offset array)

Charles


julien-tierny:
ok! neat

julien-tierny:
ok, neat!

PR #58, opened 2017-10-24
Remove legacy file

CharlesGueunet:
Dear Julien,

During the previous migration, we missed this file that had not been deleted.

Charles


julien-tierny:
Cool. thanks

PR #59, opened 2017-10-24
Compile on macOS with minimalistic setup

dsakurai:
For a macOS compile TTK without VTK wrappers, without Paraview plugins, without OpenMP, and without standalone apps.


julien-tierny:
domo arigato

dsakurai:
Je vous en prie. And your merge 149c537 compiles on my Mac, I confirm.

PR #60, opened 2017-10-25
Allow const for copy constructor of Separatrix

dsakurai:
Dear Julien,

Some collaborator using a Mac needed this fix to compile TTK successfully.

Sincerely,
Daisuke


julien-tierny:
Ok. Thanks.

PR #61, opened 2017-10-26
FTM Tree stats about tasks in advanced

CharlesGueunet:
Dear Julien,

I have fixed some stuff about the tasks stats in the FTM Tree. Also, I have added an advanced option in the CMake for enabling such stats.

Charles


julien-tierny:
Thanks!

PR #64, opened 2017-11-03
Race condition bugfix - LDistance

GuillaumeFavelier:
Dear Julien,

I fixed a concurrency bug in the `LDistance` base module by changing the scope of some variables that were shared instead of being private in each thread.
I also made the output matrix optional by checking its pointer value.


julien-tierny:
Thanks!

PR #66, opened 2017-11-22
refactor(bottleneck): clarified structure

madblade:
Features:
- enhanced structure and lisibility
- enhanced cost matrix representation
- fixed matching of last maximum
- fixed matching when nb saddles = 0
- fixed geometrical matching


julien-tierny:
Thanks a lot!

PR #67, opened 2017-11-22
FTMTree: Consistent id when only one CC

CharlesGueunet:
Dear julien,

This pull request is aimed to bring consistency on simply connected mesh.
Coming from the recent changes on the Wrapper, the connectivity and threshold filters both modify the ids of the mesh for VTU files.
We now directly copy the original input in case the mesh is already simply connected.

Regards,
Charles


julien-tierny:
Thanks!
(you may want to make sure that vtk indeed deletes these objects on exit, see lines 263 and 267, with a message in the destructor for instance)

PR #68, opened 2017-11-23
PersistenceDiagram - BugFix

GuillaumeFavelier:
Dear Julien,

I fixed a bug causing inconsistency of the pairType values in the `PersistenceDiagram` base module.


julien-tierny:
Ok, thanks!

PR #69, opened 2017-11-26
check(FTM): add light check on !Kamikaze

CharlesGueunet:
Dear Julien,
With this pull request,
when the normalizeId function is used (for consistent id btwn executions) and Kamikaze is deactivated, we check that the normalize function has seen all arcs. The guarantee the
tree structure to be valid.

Also, fix a legacy withKamikaze in ttkFTMStructures

Charles


julien-tierny:
Thanks!

PR #70, opened 2017-12-02
Avoid false negative on tree verification

CharlesGueunet:
Dear Julien,
I have recently added a new check on the tree structure after the Unification of the ids.
Their is some valid tree that could trigger this check and raise a warning even on a valid tree configuration. This commit fix this issue.

Charles


julien-tierny:
Thanks!

PR #73, opened 2017-12-15
ttkFieldSelector - select scalar fields

GuillaumeFavelier:
Dear Julien,

I think that it would be great if we could be able to select the scalar fields we are interested in and exclude the rest inside ParaView with simple checkboxes not only during data loading but anywhere in the pipeline. Of course, this solution should use shallow copy of the selected data only and avoid unnecessary operations.


julien-tierny:
Thanks Guillaume.
Great initiative.
This is super useful. I bet people will love it.

Can you please add CellData too though?

GuillaumeFavelier:
Sure, I'm working on it.

PR #75, opened 2017-12-16
Add cells scalar in OFF Reader

### LINKED ISSUE #74: OFF Reader does not import all point data
hi @CharlesGueunet ,

the OFF reader does not seem to read all point data fields.
bug scenario: 
1) open a surface in paraview
2) compute several point data fields 
3) save it to OFF format (new writer)
4) load it back into Paraview => only the last field is present (name ScalarField_n)

you may want to implement the same things for cell data too.

thanks

CharlesGueunet:
Dear Julien,

This pull request fix #74  and add the requested feature:
OFF Reader is now able to deal with Cells data.

Best,
Charles


julien-tierny:
Great. Thanks a lot for the quick fix!

PR #78, opened 2018-01-05
Improved field selectors

GuillaumeFavelier:
Dear Julien,

I improved #73 and now there are PointDataSelector and CellDataSelector plugins to select arrays associated on point or cell respectively. If the checkboxes are empty, no array will be selected as output.


julien-tierny:
Thanks a lot!

PR #80, opened 2018-01-14
Improve TTK Field/Cell Selector using regexp

CharlesGueunet:
Dear Julien,

this little pull request add a new feature to the Point/Cell data selector filters.
A string property is now available in advanced mode, corresponding to a regexp. Only the fields matching this regexp are selected by the filter. (The default value of this string is ".*")

This is useful on data set having a huge number of scalars, to select only a part of them quickly.

Charles


julien-tierny:
Ok, thanks!

PR #82, opened 2018-02-12
FTM Tree: fix when openmp disable

CharlesGueunet:
Dear Julien,

I have noticed a bug in FTM Tree when OpenMP is disabled.
There is a loop where the variable is initialized by the openmp clause, leading to wrong values when openmp is disabled. This pr fix this.

Charles


julien-tierny:
OK, thanks!

PR #84, opened 2018-02-22
Fix CRLF on paraview patch

CharlesGueunet:
Dear julien,

During the paraview patch procedure on windows, some conflict may appear between the patch files generated on Windows and the Paraview files on linux.
This is the everlasting problem of the CRLF VS LF.
This patch allows to ignore these end of line symbols when patching files on Windows, thus avoiding the problem.

Charles


julien-tierny:
Thanks!

PR #85, opened 2018-02-23
Improve CMake build type option

CharlesGueunet:
Dear Julien,

This pull request set the default build type of TTK to Release.
Moreover, we can now switch between Release and Debug mode by pressing enter on gui / ccmake.
This change is a piece of code coming from the Paraview CMakeLists.txt.

Charles


julien-tierny:
neat :)

PR #86, opened 2018-02-26
FTMTree memory leaks

CharlesGueunet:
Dear Julien,

This pull request fix the memory leaks given by Valgrind for the FTM Tree plugin.
It should not induce any performances cost.

Charles


julien-tierny:
Thanks!

PR #89, opened 2018-04-05
Allow build directory to be separated from source directory

GuillaumeFavelier:
Dear Julien,

I am well aware that in the [standard TTK installation procedure](https://topology-tool-kit.github.io/installation.html), the build directory is to be in the source directory but personally I'm used to separate those directories from each other and in my case, building TTK somewhere else works just fine except for one little thing, the documentation. 

This patch address an issue occuring when one generates the TTK documentation with the build's directory outside the source directory.

I thought that it would be great to share it.


julien-tierny:
Great fix, thanks!

PR #102, opened 2018-05-03
Morse-Smale complex - bugfix

GuillaumeFavelier:
Dear Julien,

I fixed a bug on the Morse-Smale complex module causing an unnecessary shift of the identifiers of the separatrices and improved consistency in general of these identifers.

I'll modify the ParaView state files accordingly to follow these modifications.


julien-tierny:
Awesome, thanks.

PR #104, opened 2018-05-07
MSVC fix when UNICODE is defined

martinfalk:
Hi,
here is a small fix for builds on Windows. This will fix an issue in `OsCall::listFilesInDirectory()` when `UNICODE` is defined and `windows.h` is included since this will use wchar instead of char.
Furthermore, I move the implementations to `Os.cpp` to reduce the number of include files in the header.

Martin


julien-tierny:
that looks great, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #106, opened 2018-05-11
Morse-Smale complex - new feature

GuillaumeFavelier:
Dear Julien,

I added a new advanced option in the Morse-Smale complex module allowing, with some compromises on memory, the computation of the geometry of the 2-separatrices more efficiently if OpenMP acceleration is available. In ParaView's GUI, it's name is "_Prioritize Speed Over Memory_".


julien-tierny:
Awesome, thanks a lot!

PR #107, opened 2018-05-11
Windows - bugfix

GuillaumeFavelier:
Dear Julien,

I fixed errors on Windows related to the use of C++ alternative tokens in the modules CellDataConverter and PointDataConverter.


julien-tierny:
Awesome. Thanks!

PR #108, opened 2018-05-14
micro improvement on critical points

CharlesGueunet:
Dear Julien,

This is a really small PR improving a bit the expected performances on Scalar Field critical points by adding a reserve on a vector.

Charles


julien-tierny:
Thanks!

PR #110, opened 2018-05-19
Two updates for OpenMP

jalevine:
One fixes the build w/ OpenMP in the CMakeLists.txt and the second fixes the build w/out OpenMP


julien-tierny:
Awesome. Thanks!

PR #111, opened 2018-05-24
ComponentSize - bugfix

### LINKED ISSUE #97: Not able to run viscousFingering.pvsm
viscousFingering.pvsm with viscousFingering.vti doesn't seem to work on windows. It crashes.

Jonathan

GuillaumeFavelier:
Dear Julien,

I fixed a bug in the `ComponentSize` module causing inconsistency on the output scalar fields in a configuration where OpenMP is disabled, in Windows for instance. I think this PR could help fix #97 
but this needs more investigation.


julien-tierny:
Thanks!

PR #112, opened 2018-05-26
FTM Default scalar arrays

CharlesGueunet:
Dear Julien,

this pull request set some scalar arrays as active by default:
* NodeType for nodes
* Segmentation Ids for arcs
* Segmentation Ids for Vertices
Instead of the solid color.

Also, some cleans on the code have been performed.

Charles


julien-tierny:
thanks!
I'm unsure about defaulting the display to a given scalar instead of a solid 
color, but otherwise that's all good.
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #114, opened 2018-05-29
Improve Debug mode compile flags

CharlesGueunet:
Dear Julien,

I added some flags for the Debug mode compilation of TTK, completely preventing every optimizations and enabling every code annotations. This allows a complete evaluation of the code by tools like GDB, even when templates are used.

Note, we could use -g3 for more expressiveness but I am not sure this would be compliant with other tools than GDB. 

Charles

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -19,6 +19,10 @@ function(ttk_set_compile_options library)
     endif()
   endif()
 
+  if (uppercase_CMAKE_BUILD_TYPE MATCHES DEBUG)
+     target_compile_options(${library} PRIVATE -O0 -g -pg)
+  endif()
+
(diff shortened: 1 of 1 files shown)


julien-tierny:
Awesome. Thanks!

petersteneteg:
The will not work when for example the Visual Studio compiler is used. This should be done in a platform/compiler independent way.

Also one can not use uppercase_CMAKE_BUILD_TYPE  or CMAKE_BUILD_TYPE in cmake generally is is only valid for single-configuration generators like make and will not work for multi configurations generators like Visual Studio see
https://cmake.org/cmake/help/v3.11/variable/CMAKE_BUILD_TYPE.html

One should use generator expressions for this see:
https://cmake.org/cmake/help/v3.11/manual/cmake-generator-expressions.7.html

PR #115, opened 2018-05-29
Use generic syntax in CMake as advised

CharlesGueunet:
Dear julien,

This is a fix concerning the @petersteneteg comments on #114 for the CMake options.
The behavior should be the same for single-configurations generators but also suitable for multi-configurations generator.

Charles

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -11,17 +11,11 @@ function(ttk_set_compile_options library)
 		target_compile_definitions(${library} PUBLIC TTK_ENABLE_KAMIKAZE)
 	endif()
 
+   if (TTK_ENABLE_CPU_OPTIMIZATION)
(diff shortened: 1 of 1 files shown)


julien-tierny:
hi charles,

thanks for the quick fix.

@petersteneteg:
is that ok for you guys with MSVC?
(feel free to pull request any fix if needed)

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

petersteneteg:
thanks, yes that is better, still don't think MSVC accepts those flags though. You probably only want to add those flags when using clang or gcc... 
-Wfatal-errors would be \WX on MSVC for example

julien-tierny:
> still don't think MSVC accepts those flags
> though.
thanks for your feedback.
the commit 1369508 should address that. please don't hesitate to send a pull 
request if it still does not do the trick.

best,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #116, opened 2018-06-02
New plugin to compare two data sets

CharlesGueunet:
Dear Julien,

This pull request bring a new plugin: TTK Compare.
This plugin is aimed to help automatic tests and other robustness checking.

## Principle

This plugins is aimed to compare two data sets using triangulation.
It will compare these data sets independently of the internal VTK ids and fill the "Compare" field data. It can focus solely on the mesh (vertices and cells) or also take care of scalar arrays on vertices and cells.

The main idea is to identify vertices using their position in space to create a mapper between both data set. Then, we can use this mapper for every comparison.

## Compare field

This is a bit field describing differences between data sets, added as a Filed Data on the data set. 
* 0 means both data sets are equals
* 1<<0 means vertices are not equals
* 1<<1 means cells are not equals
* 1<<2 means vertices scalars are not equals [if computed]
* 1<<3 means cells scalars are not equals [if computed]

## Output array

* VerticesDiff: vertices of the first data set not present in the second one have the value 1
* CellsDiff: cells of the first data set not present in the second one have the value 1

We note here than these two arrays can be 0 filled if the second data set contains the first one (uses the Compare field data for data sets discrimination).

If "Only Compare Mesh" is not checked, each scalar array having the same name in both data set is also checked. It will replace each scalar value on the first data set by 0 if this value is the same on the other data set (and 1 otherwise).

## Limitations

* For now, only 1D PointData / CellData are considered.
* When scalar array are considered, the output mesh is a deep copy, not to ovewrite arrays in the input data set.

## Bonus

* This plugins is parallel.
* A command line version is available (the "Comparable" field can be the return status)

Charles


julien-tierny:
hi Charles,

I'll wait for the commits we discussed before merging that one in.

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
hi charles,

if you don't mind, I'm going to close that one (it's going to be a year old soon). feel free to re-open if the situation evolves.

cheers,

PR #117, opened 2018-06-04
fix(FTM): When trunk detected by last task on a join saddle

CharlesGueunet:
Dear Julien,

Here is a fix for the FTM plugin. Sometime the computation could raise a segmentation fault when the trunk procedure was detected by the last task on a saddle.
This PR should solve this issue.

Charles


julien-tierny:
awesome, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #118, opened 2018-06-05
Morse-Smale complex - bugfix

GuillaumeFavelier:
Dear Julien,

From a developer perspective, it could be bothersome to set all the input and output pointers of the Morse-Smale complex base module when only a subset of the features is needed.

This PR adds optional arguments to the corresponding function calls (`nullptr` means unused argument), adds default values to the module flags and fixes some minor bugs.

A user of the Morse-Smale complex ParaView plugin will now see a new checkbox called "_Critical Points_" in the interface, it commands the computation of the geometric embedding of the critical points of the MSC.


GuillaumeFavelier:
My bad, I forgot to mention that I also updated the documentation of the MSC's flag setting functions to specify which arguments are optional and which are not.

GuillaumeFavelier:
I fixed some compatibility issues with the [examples](https://github.com/topology-tool-kit/ttk/blob/master/examples/c%2B%2B/main.cpp). I think this PR is ready to be reviewed.

julien-tierny:
Thanks!

PR #121, opened 2018-06-14
Morse-Smale complex - number of threads

GuillaumeFavelier:
Dear Julien,

I fixed a broken link between `DiscreteGradient` and `MorseSmaleComplex` which set the number of threads to be used for the computation.


julien-tierny:
thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #122, opened 2018-06-18
DiscreteGradient - bugfix

GuillaumeFavelier:
Dear Julien,

I fixed the first pass of the discrete gradient assignment so now it should be faster.


julien-tierny:
thanks for that first fix!
julien
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #123, opened 2018-06-19
DiscreteGradient - bugfix

GuillaumeFavelier:
Dear Julien,

I fixed the second and third pass of the discrete gradient assignment so now it should be faster.


julien-tierny:
awesome!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #124, opened 2018-06-25
DiscreteGradient - bugfix

GuillaumeFavelier:
Dear Julien,

I fixed the values of the `PLVertexIdentifier` field and selected the identifier of the vertex with the highest scalar value instead of just using `-1` as default value.


julien-tierny:
awesome, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #129, opened 2018-07-06
New data types for simplices and threads

GuillaumeFavelier:
Dear Julien,

This PR updates the modules so that they can use user-defined data types _i.e._ `SimplexId` and `ThreadId` in the file [`DataTypes.h`](https://github.com/GuillaumeFavelier/ttk/blob/dataTypes_migration/core/base/common/DataTypes.h). This should allow the processing of much bigger data sets.


julien-tierny:
Awesome! (needed to update one header file but that was an easy fix)
Thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #135, opened 2018-07-30
TriangulationRequest - bugfix

GuillaumeFavelier:
Dear Julien,

I fixed a bug occuring when requesting the link of a vertex in 2D.


julien-tierny:
awesome :)
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #145, opened 2018-10-01
Minor include in FTM

CharlesGueunet:
Dear Julien,

I have added a missing include in the FTM Wrapper. This include is required for the new ttkIdType system. The missing include was reported by Clang linter but did not prevent compilation in my experiments.

Charles


julien-tierny:
awesome.
thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #151, opened 2018-10-12
Cinema Filters; Depth Image Based Geometry Approximation; Sqlite3 Dep.

JonasLukasczyk:
Cinema Filters; Depth Image Based Geometry Approximation; Sqlite3 Dependency


julien-tierny:
awesome! 
great stuff. thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #154, opened 2018-10-28
Cinema and DIBGA Update

JonasLukasczyk:
New Features:
* CinemaImaging can now produce value images
* DIBGA can now map input value images to approximations
* improved console output


julien-tierny:
awesome, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #156, opened 2018-11-05
DataSetInterpolator - bugfix

GuillaumeFavelier:
Dear Julien,

I fixed a bug in the ttkDataSetInterpolator plugin causing the original data arrays in the target of the probe to disappear in the final result.


julien-tierny:
awesome!
thanks a lot
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #159, opened 2018-11-14
fix(ttkBottleneckDistance): indexing; fix(BottleneckDistance): lw metric

madblade:


julien-tierny:
awesome, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

Issue #158, opened 2018-11-09
Paraview crashes when loading state with missing file


kamakshidasan:
Tried loading https://raw.githubusercontent.com/topology-tool-kit/ttk-data/master/states/persistenceClustering0.pvsm without the necessary files.

ERROR: In /Users/adhitya/ttk/ParaView-v5.5.0/VTK/IO/Infovis/vtkDelimitedTextReader.cxx, line 734
vtkDelimitedTextReader (0x7fd59c6b0e80): caught exception: Unable to open input file clustering0.csv

Warning: In /Users/adhitya/ttk/ParaView-v5.5.0/VTK/Imaging/Hybrid/vtkGaussianSplatter.cxx, line 209
vtkGaussianSplatter (0x7fd59f5d1590): No POINTS!!

[ttkPersistenceDiagram] Starting computation on field `SplatterValues'...
Segmentation fault: 11

julien-tierny:
hi @Kamakshidasan,
@GuillaumeFavelier's PR #163 should have fixed this.
I'm going to go ahead and close this one.
please feel free to re-open if you run into the same issue again.
cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

Issue #152, opened 2018-10-15
Paraview crashes on renaming variable


kamakshidasan:
Steps to reproduce this:

- Load the attached VTK file
- Add calculator plugin with sqrt((U * U) + (V * V)) [Do not rename the default scalar 'Result']
- Perform FTM on the VTK file
- Rename calculator plugin from 'Result' to 'magnitude'

Paraview/TTK crashes with the message: 

"[ttkFTMTree] Error : input scalar 0 field pointer is null.
Launch on field : Result
Segmentation fault: 11"

[tv_0.vtk.zip](https://github.com/topology-tool-kit/ttk/files/2478738/tv_0.vtk.zip)

kamakshidasan:
I updated the above comment:

It should be read as, "calculator plugin with sqrt((U * U) + (V * V))"

For some reason, Github needs a space before and after the multiplication sign.

julien-tierny:
hi @Kamakshidasan,
@GuillaumeFavelier's PR #163 should have fixed this.
I'm going to go ahead and close this one.
please feel free to re-open if you run into the same issue again.
cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #163, opened 2018-11-15
Check input - bugfix

GuillaumeFavelier:
Dear Julien,

I implemented some more input sanity check in the code to avoid the crashes of ParaView encountered in #152 and #158.


julien-tierny:
awesome, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

Issue #147, opened 2018-10-04
ttkPersistanceCurve's parent class adjustment


HollowSun0:
I've been trying to integrate ttkPersistanceCurve into another filter by doing

`

		vtkInformation *inInfo0 = inputVector[0]->GetInformationObject(0);
		vtkDataSet *input = vtkDataSet::SafeDownCast(
			inInfo0->Get(vtkDataObject::DATA_OBJECT()));
		
		vtkInformation* outInfo0 = outputVector->GetInformationObject(0);
		vtkTable* output0 = vtkTable::SafeDownCast(
			outInfo0->Get(vtkDataObject::DATA_OBJECT()));
			
		vtkInformation* outInfo1 = outputVector->GetInformationObject(1);
		vtkTable* output1 = vtkTable::SafeDownCast(
			outInfo1->Get(vtkDataObject::DATA_OBJECT()));
			
		vtkInformation* outInfo2 = outputVector->GetInformationObject(2);
		vtkTable* output2 = vtkTable::SafeDownCast(
			outInfo2->Get(vtkDataObject::DATA_OBJECT()));
			
		vtkInformation* outInfo3 = outputVector->GetInformationObject(3);
		vtkTable* output3 = vtkTable::SafeDownCast(
			outInfo3->Get(vtkDataObject::DATA_OBJECT()));

		vtkSmartPointer<ttkPersistenceCurve> persistenceCurveFilter
			= vtkSmartPointer<ttkPersistenceCurve>::New();
		persistenceCurveFilter->SetInputData(input);
		persistenceCurveFilter->SetUseAllCores(this->UseAllCores);
		persistenceCurveFilter->SetThreadNumber(this->ThreadNumber);
		persistenceCurveFilter->SetdebugLevel_(this->debugLevel_);
		persistenceCurveFilter->SetScalarField(this->ScalarField);
		persistenceCurveFilter->SetInputOffsetScalarFieldName(this->InputOffsetScalarFieldName);
		persistenceCurveFilter->SetForceInputOffsetScalarField(this->ForceInputOffsetScalarField);
		persistenceCurveFilter->SetComputeSaddleConnectors(this->SaddleConnectors);
		persistenceCurveFilter->Update();

		output0->ShallowCopy(persistenceCurveFilter->GetOutput(0));
		output1->ShallowCopy(persistenceCurveFilter->GetOutput(1));
		output2->ShallowCopy(persistenceCurveFilter->GetOutput(2));
[...]

PR #164, opened 2018-11-15
PersistenceCurve - SafeDownCast

GuillaumeFavelier:
Dear Julien,

I implemented the feature requested in #147 so now the outputs of the ttkPersistenceCurve filter are valid vtkTable objects.


julien-tierny:
awesome!
thanks
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #165, opened 2018-11-16
Debug::dMsg now always flushes stream

JonasLukasczyk:


julien-tierny:
neat, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #167, opened 2018-11-20
New plugin to sort cells by scalar value

GuillaumeFavelier:
Dear Julien,

I implemented a plugin to sort the cells of a data set according to their scalar field values. The plugin creates an identifier scalar field associated to the cells as an output. The user can choose the order (increasing or decreasing) and if it starts with zero or one.


julien-tierny:
great, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #169, opened 2018-11-22
Improved i/o in ttkUncertainDataEstimator

julesvidal:
Hello Julien, 

As we discussed I modified the UncertainDataEstimator plugin to enforce the following properties :
- The plugin can take one or several inputs
- All the fields of each input are included in the computation
(the plugin is meant to be used with PointDataSelector to clean the inputs of unwanted fields beforehand)
- The names of the input fields can be different (thus I removed the corresponding property in the paraview gui)
- The only outputs are the resulting fields : lower/upper bounds, probabilities and mean of the ensemble data

see you soon,

Jules


julien-tierny:
neat, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

Issue #170, opened 2018-11-26
TTKBaseConfig.cmake contain boost absolute path 


mv740:
Hi, 

TTKBaseConfig.cmake and TTKVTKConfig.cmake contain the boost directory absolute path, thus we can't share the install folder of TTK with another computer.  

Is it possible to change it for  a variable ${BOOST_HOME}, then when we used FindPackage(TTKBase), it would then ask us for TTKBASE_DIR and BOOST_HOME

julien-tierny:
hi michal,

> Is it possible to change it for  a variable ${BOOST_HOME}, then when we used FindPackage(TTKBase), it would then ask us for TTKBASE_DIR and BOOST_HOME 
can you provide an example showing how to implement that in cmake?
(as a reminder, these files are generated automatically upon the install calls lines 141 of core/base/CMakeLists.txt and 72 of core/vtk/CMakeLists.txt)
@Twinklebear: any suggestion?

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

Twinklebear:
I think it should be possible to update the cmake config files to search for the dependencies when you include TTK, [based on this mailing list thread](https://cmake.org/pipermail/cmake/2014-October/058804.html). Then we can provide an input CMake config template that searches for the various dependencies when you import TTK so @mv740 's use case would work (as long as the Boost it finds is still compatible w/ the built one).

julien-tierny:
hi there,
@Twinklebear's PR #171 should have fixed that for boost. I'll go ahead and close this one. please feel free to re-open if needed.
cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #171, opened 2018-11-27
WIP: Have TTK's CMake config files find dependencies

Twinklebear:
Hey Julien, I later found https://cliutils.gitlab.io/modern-cmake/chapters/install/installing.html which I think is the cleanest way to have the the config files we make for TTK's packages find their corresponding dependencies. This also lets us find other dependencies for the user depending on the build configuration (e.g., OpenMP, MPI, VTK, etc.). We can also have the TTKVTK CMake file automatically import the TTKBase config, so that people using TTKVTK don't need to also search for TTKBase, we just handle it for them.

This is still a bit of a work in progress, I haven't tested if this has any effect on the ParaView plugin side of things or how things work with the VTK wrappers when building just VTK, or with ParaView for the dependent applications. But this is probably the path to go for resolving #170 . Feel free to edit this as well

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -182,10 +182,7 @@ if(TTK_BUILD_DOCUMENTATION)
     endif()
 endif()
 
-find_package(Boost COMPONENTS system)
-if(NOT Boost_FOUND)
-  find_package(Boost REQUIRED)
-endif()
+find_package(Boost REQUIRED COMPONENTS system)
 
 find_package(ZLIB)
 if(NOT ZLIB_INCLUDE_DIR)
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -8,8 +8,7 @@ function(ttk_set_compile_options library)
     target_compile_options(${library} PRIVATE -Wall)
 
     if(Boost_FOUND)
-    target_include_directories(${library} PUBLIC ${Boost_INCLUDE_DIR})
-    target_link_libraries(${library} PUBLIC ${Boost_LIBRARIES})
+		target_link_libraries(${library} PUBLIC Boost::boost Boost::system)
     endif()
 
     if (TTK_ENABLE_KAMIKAZE)
@@ -196,6 +195,7 @@ else()
   option(TTK_ENABLE_ZLIB "Enable Zlib support" ON)
 endif()
 
+# TODO: This should be in its own findpackage.cmake file!
 # START_FINDSQLITE3
 find_path(SQLITE3_INCLUDE_DIR sqlite3.h
     PATHS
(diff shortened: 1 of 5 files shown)


Twinklebear:
For example in topo-vol, this cleans up the CMake work needed to import TTK pretty nicely https://github.com/Twinklebear/topo-vol/commit/2a5d969fc20a246090935200869f78c6aebdd307 , now we just do `find_package(TTKVTK)` and TTKBase and any dependencies will automatically be looked for. Though, if they can't be easily found users will still need to pass `-DLIBNAME_DIR` to find them.

Twinklebear:
For the failing appveyor, it may be running an older CMake where the Boost targets hadn't been added yet. It should still be do-able to support the older cmake following a similar thing to what I put in for VTK (but that needs some testing). It looks like the Boost:: targets where added around CMake 3.5

Twinklebear:
@mv740 would you mind testing this as well?

julien-tierny:
hi will,

all of this looks awesome.

I like a lot the fact that TTKBase gets automatically pulled in by TTKVTK, as well as their dependencies.

can you try to commit an alternate fix for windows such that appveyor gets happy?

also, I'm not sure how you plan to handle the remaining dependencies in general (sqlite3, zfp, numpy). for instance zfp does not install any config cmake file (thus there's no find_package available)

thanks for your input :)

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

Twinklebear:
For those packages we'd probably need to write our own FindX.cmake file and include it with the installation, though I think recently ZFP has added a ZFPConfig.cmake file. For the others we'd probably need to make one ourselves, essentially taking the code that's in the CMakeLists now to find them and moving it to FindX.cmake, and modifying the CMake module path to find the modules.

I'm not sure on the windows build, it seems like it has an up to date enough CMake version, so I'll look at it a bit more later (likely after eurovis)

mv740:
Hey, 
In TTT..Targets.cmake I can see  " INTERFACE_LINK_LIBRARIES "ttk::base::triangulation;Boost::boost;Boost::system".   No more absolute path !

I have buit it on a Windows desktop.  The latest version of cmake should support more boost version.

Thanks a lot for this patch

julien-tierny:
hi there,

I'll merge that PR if you don't mind and mark the other dependencies as work in progress.
I'm not a fan of the idea of providing our own install CMake files for third-party dependencies though (that looks like an upstream job).

fyi, appveyor is not happy because it does not build the build-able components of boost (we don't have time for that with the free plan) so it fails when trying to consider Boost::system for linking. I'll post a tiny fix.

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #173, opened 2018-12-03
fix(tracking): removed trackingFromPersistenceDiagrams vtk dependency

madblade:
Removed ttkBottleneckDistance as a dependency of ttkTrackingFromPersistenceDiagrams


julien-tierny:
alright, looks great!
thanks
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #177, opened 2018-12-17
CMake: generate only on library for all ParaView filters

CharlesGueunet:
Dear Julien,

This pull request allows to generate a single plugin containing all filters for ParaView.
Additionally:
* On all *CMakeLists.txt*, functions have been moved into a separate file
* VTK is not required anymore if ParaView is included, as it already contains a VTK
* The *PLUGIN_VERSION* is now deprecated on a per filter basis and the single filter uses the TTK version
* Some fix for Windows (troubles with lambda expressions)

This work has been tested on Linux and Windows, but I had trouble with boost on Mac (not linked to this PR)

Charles


julien-tierny:
hi charles,

thanks for the PR.

here's a few remarks/comments:

1. the functions.cmake files make sense. thanks for organizing that!

2. can you clarify why you had to edit core/base/meshGraph/MeshGraph.h? (build issues?)

3. could you provide an advanced option TTK_BUILD_STANDALONE_PARAVIEW_PLUGINS (disabled by default) that would switch to the current behavior (1 lib per filter) when enabled? 
this is useful for developers when they want to get rid of a buggy filter or when they want to focus on a single one. that'd be neat!

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

CharlesGueunet:
> 2. can you clarify why you had to edit core/base/meshGraph/MeshGraph.h? (build issues?)

Absolutely, as briefly explained in d942c4c, MSVC had troubles compiling the code because of a rule about ternary operations: they are required to have the same type for second and third operand. On another hand, even if the two lambda on the ternary had the same signature, lambda functions are not required to have the same type (even with a same signature) so MSVC failed to compile that.
I have just moved the ternary inside the lambda to fix this issue.

> 3. could you provide an advanced option TTK_BUILD_STANDALONE_PARAVIEW_PLUGINS (disabled by default) that would switch to the current behavior (1 lib per filter) when enabled?
this is useful for developers when they want to get rid of a buggy filter or when they want to focus on a single one. that'd be neat!

Working on it !

julien-tierny:
> * VTK is not required anymore if ParaView is included, as it already contains a VTK
I'm not so sure about that.
what if users don't want to build paraview plugins but still want the vtk layer? (vtk becomes required in that case)

the current version is fine with me (I'm using a system-wide install of vtk that differs in version from paraview's)
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
> Absolutely, as briefly explained in d942c4c, MSVC had troubles compiling the code because of a
app-veyor didn't report anything about that. 
what error did you get? what version of MSVC were you using?
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

CharlesGueunet:
The use of the VTK shipped with ParaView is only a fallback in the case VTK was not already provided by the user. The behavior stay the same if `VTK_DIR` is set or if a system-wide VTK is available.

core/vtk/CMakeLists.txt
```diff
@@ -1,20 +1,18 @@
 include(functions.cmake)
 
-option(TTK_FORCE_VTK_PARAVIEW "Force to use the VTK bundled into ParaView" OFF)
```
julien-tierny: I think you wanted to keep that one in (instead of the one in standalone/CMakeLists.txt, which is less general)

CharlesGueunet:
As a complementary information,
I have tested this work on a Windows environnement, following the TTK's website instructions and it works.
I had to build VTK and install it on my system though, while the `TTK_FORCE_VTK_PARAVIEW` can be used to avoid having to build a separate VTK. That could be a good thing for TTK users, is it worth a word on the website ?

PR #178, opened 2018-12-19
Fix for mandatory critical points

CharlesGueunet:
Dear Julien,

The last commit (b19fb5b9d6ec4fd69274693a7bf30a1dcb5f80f0) introduced a bug in the mandatory critical point filter. This bug was due to a conflict between the classe `Node` in ***ContourTree.h*** and the class `Node` in ***ContourForestsTree.h*** (idem for the class `SuperArc`).
I have put ContourForests in its own namespace, as I did for FTM, and the issue is fixed.

Idea, could we add a namespace in the blank example so new filter could easily live in their own namespaces ?

Charles

### DIFF
--- a/core/base/contourForests/ContourForests.cpp
+++ b/core/base/contourForests/ContourForests.cpp
@@ -12,6 +12,7 @@
 
 using namespace std;
 using namespace ttk;
+using namespace cf;
 
 Interface::Interface(const SimplexId &seed) : seed_(seed)
 {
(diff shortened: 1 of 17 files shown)


julien-tierny:
Awesome Charles!

> Idea, could we add a namespace in the blank example so new filter could easily live in their own namespaces ?
with the new unified plugin library, that may be a good idea indeed.

thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #181, opened 2019-01-09
FTM Tree minor fix for scalar arrays

CharlesGueunet:
Dear Julien,

The use of SetScalars to set the active scalar array for the segmentation output implied the last active scalar was removed from the output data set. I have replaced it by an AddArray + SetActiveScalar to avoid this drawback.

Charles


julien-tierny:
alright, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #182, opened 2019-01-09
refactoring CMakeLists.txt, use proper list syntax

CharlesGueunet:
Dear Julien,

I have made a minor change in the CMakeLists.txt to handle dependencies using list(append) instead of custom set(variable).

Charles


julien-tierny:
looks good, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #183, opened 2019-01-10
fix in the CMake when using list

CharlesGueunet:
Dear Julien,

The previous PR: #182 introduced a bug where the list was not propagated to the global level for some new builds. This Pull Request fix this bug.

Charles


julien-tierny:
ok, can you please double-check everything is back to normal?
thanks
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #187, opened 2019-02-08
core/CMakeLists: Force generation of a compilation database

pierre-guillou:
A compilation database is a small file usually called
compile_commands.json, located at the root of the build directory.

It contains compilation rules for every source file, including
definitions as passed by CMake, those being specific to the current
build directory.

Once generated, this file can be used by third-party tools like
clang-tidy [1] (a C/C++ code linter which detects code defects) or
language servers [2], such as RTags (for Emacs) and
clangd/cquery/ccls. Those can be used by code editors to provide them
IDE-like features (improved code completion, better warnings,
goto-definition).

[1] https://clang.llvm.org/extra/clang-tidy/
[2] https://langserver.org/

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -9,6 +9,9 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
    set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release")
 endif()
 
+# Generate a compilation database
+SET(CMAKE_EXPORT_COMPILE_COMMANDS TRUE)
+
 option(TTK_ENABLE_64BIT_IDS "Enable processing on large datasets" OFF)
 mark_as_advanced(TTK_ENABLE_64BIT_IDS)
 


julien-tierny:
sure, why not.
thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
now that I'm double-checking this, I'm not sure that forcing everybody to set this flag on by default really makes sense. only for (a portion of) the devs I guess and those already have tailored cmake configurations anyways.
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

CharlesGueunet:
As a developer, you can use the `-C` option of `cmake` to give an initial configuration with for example the`CMAKE_EXPORT_COMPILE_COMMANDS` set to `ON`. This would have the same effect without impacting all users.
Also, most of the tools used by IDE/Editor can also force this option for your cmake.

I would join Julien in this and I think this option may be set to ON only by users that known what they do, I would not force it for everyone.

pierre-guillou:
I see that my proposition has met some resistance, let me try to raise more arguments in favor of it :)

* While I agree that activating this flag has no direct benefit to the end user, the generated file can be used as a debugging tool. For example, in case of a failing build, the compilation database can be sent to the developer team alongside the CMakeCache.txt to better understand the user environnment. However, note that link commands do not appear in this file :(
* Developers can have an easier way 1. to discover and 2. to use compilation database-based tools on TTK codebase.
* Some people advocate for enabling this behavior by default in CMake [1].
* The same thread shows that enabling this flag has no consequence on configuration time.
* LLVM has enabled it [2]

If these arguments fail to convince you, don't hesitate to revert this commit. It will still live somewhere in my private repo :)

@CharlesGueunet I did not know about this ```-C``` flag. How do you suggest I use it? Should it be set for every ```cmake``` invocation? (not very practical IMO) Or, as proposed in [1] (at the very end), should I use a shell alias ? (kind of hackish). Is there a configuration file (or an environment variable) I can edit that will be automatically called by CMake ?

[1] https://gitlab.kitware.com/cmake/cmake/issues/18386
[2] https://github.com/llvm-mirror/llvm/blob/838fcc69d62c420ae74885208e0de5c01f361236/CMakeLists.txt#L218 (last commit on llvm/CMakeLists.txt as of writing this comment)

CharlesGueunet:
> * While I agree that activating this flag has no direct benefit to the end user, the generated file can be used as a debugging tool. For example, in case of a failing build, the compilation database can be sent to the developer team alongside the CMakeCache.txt to better understand the user environnment. However, note that link commands do not appear in this file :(

I am not sure what you mean here. What informations would be given by the `compile_commands.json` not already present in the `CMakeCache.txt` ?

> * Developers can have an easier way 1. to discover and 2. to use compilation database-based tools on TTK codebase.

Agreed, but I am not advocating for not using this file, only for letting the developer setting the corresponding option by himself.

> * The same thread shows that enabling this flag has no consequence on configuration time.

Brad King himself seems to tell otherwise in the thread you mention (and he knows CMake quite well).
In my tests, ~22 seconds against ~24 seconds for TTK depending on this flag.

> @CharlesGueunet I did not know about this `-C` flag. How do you suggest I use it? Should it be set for every `cmake` invocation? (not very practical IMO) 

You can if you want, but this is not my way of doing things. I use zsh so I can use post alias. I just have to write `CCD` in my `cmake` command, the alias extend to `-C ~/init.cmake`.

> Or, as proposed in [1] (at the very end), should I use a shell alias ? (kind of hackish). Is there a configuration file (or an environment variable) I can edit that will be automatically called by CMake ?

Sadly, there is no environment variable I am aware of yet. Still, your IDE od Editor should take care of this and force cmake to use this option by default, hence avoiding the need of an alias. Anyway, the `EXPORT_COMPILE_COMMANDS` only make sens on projects you are developing :)

pierre-guillou:
@CharlesGueunet thanks for your help!

> I am not sure what you mean here. What informations would be given by the `compile_commands.json` not already present in the `CMakeCache.txt` ?

The `compile_commands.json` is a lower-level representation of the build process. It contains the name of every source file, and also preprocessor definitions set by CMake. I cannot find this information in `CMakeCache.txt` but I assume it is encoded in the generated `Makefiles`. I suppose it is possible to extract manually it from the `CMakeCache.txt`, but not in a convenient way.

> You can if you want, but this is not my way of doing things. I use zsh so I can use post alias. I just have to write `CCD` in my `cmake` command, the alias extend to `-C ~/init.cmake`.

I am also a zsh user, but I think I will rely on basic aliases to override the `cmake` and the `ccmake` commands. I tend to use CMake in a terminal, and I don't like typing more letters :)

> Sadly, there is no environment variable I am aware of yet. Still, your IDE od Editor should take care of this and force cmake to use this option by default, hence avoiding the need of an alias. Anyway, the `EXPORT_COMPILE_COMMANDS` only make sens on projects you are developing :)

As an Emacs user, I know there is a `cmake-project` package [1] to call CMake directly from my editor, but I have not tried to use it yet. I find it quicker to use a terminal to build than to type `M-x compile` :)

[1] https://github.com/alamaison/emacs-cmake-project

julien-tierny:
alright, rolling back for now.
as usual, will reconsider if a concrete situation justifies it. 
in the meantime, individual config files will do the trick to enable this convenience without having to force it for all.
thanks for the discussion.
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

pierre-guillou:
Well, to be more precise, using ```cmake -C``` to set ```CMAKE_EXPORT_COMPILE_COMMANDS``` in a separate file doesn't work, since this is not a cache variable.
I tried that yesterday, and I did not get any ```compile_commands.json``` generated :(

Nonetheless, setting shell aliases:

```sh
alias cmake='cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=TRUE'
alias ccmake='ccmake -DCMAKE_EXPORT_COMPILE_COMMANDS=TRUE'
```
works for me.

CharlesGueunet:
> Well, to be more precise, using `cmake -C` to set `CMAKE_EXPORT_COMPILE_COMMANDS` in a separate file doesn't work, since this is not a cache variable.
I tried that yesterday, and I did not get any `compile_commands.json` generated :(

Why don't you make it a cached variable ?
for me using:
```cmake
set(CMAKE_EXPORT_COMPILE_COMMANDS ON CACHE BOOL "")
```
in the `init.cmake` file works as expected.

pierre-guillou:
@CharlesGueunet Because I'm still a newbie at CMake :)

PR #189, opened 2019-02-22
FTM: Perf, single leaf search for both merge trees

CharlesGueunet:
Dear Julien,

This is the first commit of the FTC series.
This commit allows to preform a single leaf search when 2 merge trees are computed, thus avoiding a traversal on the data.

Charles


julien-tierny:
ok, looks good.
thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

Issue #190, opened 2019-02-22
Automatic indendation & code formatting


CharlesGueunet:
Dear Julien,

As there is more and more people contributing to TTK (which is great btw !) we may want to introduce common rules for code formatting. This would improve code readability and ease contributions.

I propose to use [clang-format](https://clang.llvm.org/docs/ClangFormat.html) as this is a robust tool, almost established as a standard with a wonderful git integration. Doing so, we can provide a *.clang-format* file for the contributor so that they can respect the rules established for TTK, but we can even automatically format with git any new commit so that any new contributions will be correctly formatted and even the code-base already established is likely to be indented little by little as we improve it.

The choice of the best indentation is not mine, so I here is my *.clang-format* file, [here](http://charles.gueunet.fr/doc/.clang-format). I have already used it on FTM and any other plugins I worked on in TTK as an example. Alternatively, there is a really good [configurator](https://zed0.co.uk/clang-format-configurator/) if you prefer to create your file form scratch.

Best,
Charles

julien-tierny:
hi charles, 

thanks for pointing that out.
ttk now has a .clang-format. 

the (currently minimalist) dev wiki points to scripts for automatically calling clang-format on git pre-commits, something we may formally encourage in the future (to further discuss): 
https://github.com/topology-tool-kit/ttk/wiki/Dev-wiki

I'm going to close that issue. please re-open if needed.

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #191, opened 2019-02-28
Geometry: Template functions to extract a float version

pierre-guillou:
The Geometry library contains functions whose parameters are only doubles. This conflicts with the Triangulation class, which operates on float values.

The current branch aims to solve this issue by providing a template version of the Geometry library, with template instantiations for double and float types. No direct user of the library should be impacted by this change; the function call syntax is the same.

Also in the same branch, two *slightly controversial* changes:
1. I have applied clang-format onto Geometry.h and Geometry.cpp prior to working on those files to improve their readability on my editor (using default formatting, waiting on #190 to provide specific options)
2. I have switched out the Geometry class in favor of a Geometry namespace, since all methods were static and no state was used. That change should be easy to revert.


julien-tierny:
Looks good, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #192, opened 2019-03-01
Replace omp_lock_t usage with critical sections

pierre-guillou:
An OpenMP critical section is executed by only one thread at a given
time.

Since this is apparently the intended usage of those omp_lock_t
variables, using critical sections improves code readability and lets
the OpenMP run-time handle the low-level
bits (c.f. https://stackoverflow.com/questions/29587124/openmp-critical-section-vs-locks).


julien-tierny:
ok, why not.
thanks
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #196, opened 2019-03-03
Fix overflow in vertex count calculation

klacansky:


julien-tierny:
hi pavol,

thanks for your PR.

I'm not so sure about that fix though.
your previous commit (klacansky@cc6d6b7) which changed the prototype of the function ImplicitTriangulation::setInputGrid (by replacing int by SimplexId) looked better to me. can you PR that instead (without the printfs)?

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #197, opened 2019-03-04
Minor fix: Unused variable with GCC on FTR

CharlesGueunet:
Dear Julien,

This PR remove a warning when compiling FTR with GCC.
Also, I have updated the warning message on Contour Forests to also mention FTR.

Charles


julien-tierny:
awesome, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #200, opened 2019-03-04
FTR: SoS offsets use SimplexId type

CharlesGueunet:
Dear Julien,

the SoS array in FTR is now in SimplexId type instead of vertexId (both are the same under the hood, this is a semantic improvment).
Should this type be changed in favor of a template ?

Charles

### DIFF
--- a/core/base/ftrGraph/Scalars.h
+++ b/core/base/ftrGraph/Scalars.h
@@ -28,7 +28,7 @@ namespace ttk {
       idVertex size_;
 
       ScalarType *values_;
-      idVertex *offsets_;
+      SimplexId *offsets_;
 
       bool externalOffsets_;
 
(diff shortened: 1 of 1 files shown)


julien-tierny:
hi charles,

> the SoS array in FTR is now in SimplexId type instead of vertexId (both are the same under the hood, this is a semantic improvment).
thanks for this PR.

> Should this type be changed in favor of a template ?
honestly, I don't believe so. as you know very well, even vtk is not templating their ids (which are configured at build time). I think we should stick to that too and remove the double templating in the future from the other modules (ftm, topological simplification, morse-smale complex, etc.). this echoes a discussion we had last week with @GuillaumeFavelier. 

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #201, opened 2019-03-04
FTR fix empty nodes output

CharlesGueunet:
Dear Julien,

It seems I forgot the shallowCopy on the getSkeletonNode function... This is fixed here!

Charles


julien-tierny:
awesome, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #202, opened 2019-03-05
FTR: Fix wrapper for data sets with multiple CC

CharlesGueunet:
Dear Julien,

This is a fix for FTR, where the number of nodes in the arc structure was wrongly deduced from the number of arcs. This could lead to memory corruption on data sets with several connected components.

Charles


julien-tierny:
awesome, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #203, opened 2019-03-05
Refix for FTR: the number of node in the arc when sampling

CharlesGueunet:
Dear Julien,

The last fix for the reserve on the number of nodes on the arc skeleton was not enough as this number can be higher than the number of node in the RG if sampling is used. A pessimistic reserve is now used, where the number of nodes is twice the number of arcs.

Charles


julien-tierny:
alright, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #204, opened 2019-03-05
Add harmonic field computation plugin

pierre-guillou:
Add a new plugin solving harmonic equations on meshes.
This module takes a mesh and a set of vertices with scalar constraints on them, and computes an harmonic scalar field that should meet those constraints on those vertices.
Two methods of computing the laplacian matrix of the mesh are proposed:
* the direct laplacian matrix of the graph
* the cotangent weights matrix

Similarly, two methods of solving the harmonic equation can be used:
* for small meshes, a direct resolution by a Cholesky factorization,
* for larger meshes, an iterative method based on conjugate gradients.

The solver is externalized to the Eigen C++ library.


julien-tierny:
that looks awesome!
thanks
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #205, opened 2019-03-06
Factor textures in standalone folder

pierre-guillou:
Some GUI apps in the standalone folder were embedding the same set of 11 PNG textures. Some of those were installed in the same ```${TTK_INSTALL_BINARY_DIR}``` directory by the CMake ```install``` command, overwriting the same image from another directory.

The current set of commits aims at fixing this problem by:
* removing the duplicates CMake ```install``` commands
* removing the duplicates PNG textures
* keeping only one copy in a top-level directory, here standalone/textures, with only one install command


julien-tierny:
looks good, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #206, opened 2019-03-06
Fix int overflow in vertex number calculation

klacansky:


julien-tierny:
thanks pavol for this PR.
may that be the first of many :)
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #208, opened 2019-03-08
Standalone textures: Manage install path in executables

pierre-guillou:
The GUI executables in the ```standalone``` folders were searching PNG textures into their local directory.
The current set of commits make them look for textures into the installation directory when built by CMake, or locally otherwise.

The textures installation directory, since containing no executable files, was also changed from ```$prefix/bin``` to ```$prefix/share/ttk```, to be more consistent with the standard filesystem hierarchy.


julien-tierny:
looks good, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #209, opened 2019-03-10
FTR: Great performance improvement

CharlesGueunet:
Dear Julien,

This PR brings new performance improvements to FTR throughout two main contributions:

* Improve the laziness mechanism so a single list of operation is used, on which deletion are directly applied.
* Use the scalarFieldCritical point filter to extract critical points before the sweep. 

For this second contribution, the scalarFieldCriticalPoint filter has been refactored a bit:
The API now expose the getNumberOfLowerUpperComponent function, wich return a pair with the number of CC in the lower / upper star of a vertex.
Also, the shortcut for extrema avoiding the computation of the lower/upper star can be discarded if the mesh is not manifold
Finally, the filter has been re-indented with the new clang-format style

Charles

### DIFF
--- a/core/base/ftrGraph/CMakeLists.txt
+++ b/core/base/ftrGraph/CMakeLists.txt
@@ -34,7 +34,7 @@ ttk_add_base_library(ftrGraph
    HEADERS AtomicVector.h FTRGraph.h FTRGraph_Template.h
            FTRGraphPrint_Template.h FTRGraphPrivate_Template.h DynamicGraph.h
            Lazy.h Mesh.h Propagation.h Propagations.h Scalars.h Segmentation.h
-           LINK triangulation ${profiler_lib})
+           LINK triangulation scalarFieldCriticalPoints ${profiler_lib})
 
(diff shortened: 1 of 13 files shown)


julien-tierny:
hi charles,

thanks a lot for the cool work.

ttk::ScalarFieldCriticalPoints<dataType>::getCriticalType() does not seem to return extremum types anymore (CriticalType::Local_minimum, CriticalType::Local_maximum).
can you fix that please?

also, in the future, can you restrict your clang-format to the lines you actually commit? that will make the PR much more readable. you can use scripts such as those: https://github.com/barisione/clang-format-hooks/

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

CharlesGueunet:
I have changed the getCriticalType so now extrema are correctly handled. Moreover, in the case of non manifold mesh a vertex which is both an extremum and a saddle is considerate degenerate.

julien-tierny:
alright, looks good.
thanks
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #210, opened 2019-03-11
Fix for ScalarField critial points: Monkey Saddle

CharlesGueunet:
Dear Julien,

In the last P-R, monkey saddles in 2D have been classified as degenerate saddles. This was wrong so this pull request re-classify them as normal saddles.

Charles


julien-tierny:
thanks for the quick fix!
julien
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #207, opened 2019-03-07
ftrGraph/ftmTree: Rename installed headers

pierre-guillou:
The ftrGraph and ftmTree folders inside ```core/base``` were containing two sets of headers with the same name:
* AtomicVector.h
* Segmentation.h

The CMake installation script was overwriting one by the other.
In order to prevent issues for user applications including TTK headers, the offending headers were renamed by prefixing them with their parent folder name.

Every other installed header of those two folders has been also renamed following the same pattern.

PR #211, opened 2019-03-13
File name conflict between FTM and FTR

CharlesGueunet:
Dear Julien,
As already mentioned in #207 a file name collision between FTM and FTR could led to files being overwritten during the installation process.
This has been fixed by renaming the files.

Charles


julien-tierny:
looks good, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #212, opened 2019-03-15
Fix two buffer overflows using AddressSanitizer

pierre-guillou:
I was playing with the [AddressSanitizer](https://github.com/google/sanitizers/wiki/AddressSanitizer) and it founds those two buffer overflows.


julien-tierny:
ok, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #213, opened 2019-03-17
CMake, warn for ZFP only once

CharlesGueunet:
Dear Julien,

This PR is only a proposition. I do not like having a warning about ZFP each time CMake is run, with this change the Warning is raised the first time, but not recalled for each following cmake.

Charles


julien-tierny:
a priori, I have several arguments against.
let's discuss this... after the vis deadline :)
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #215, opened 2019-04-04
Add new dijkstra module computing shortest paths on a mesh

pierre-guillou:
This new module depends on the geometry and triangulation modules, and exports one unique function, shortestPath, to compute for a given mesh vertex the shortest path to every other mesh vertex.

A vector of vertices can be set as boundaries to limit the computation to a subset of the mesh.

This implementation has replaced the previous one located in the distanceField module. Using std::priority_queue instead of std::set has lead to a x1.5 speedup compared to the distanceField implementation.

### DIFF
--- a/core/base/dijkstra/CMakeLists.txt
+++ b/core/base/dijkstra/CMakeLists.txt
@@ -1,0 +1,6 @@
+ttk_add_base_library(
+  dijkstra
+  SOURCES Dijkstra.cpp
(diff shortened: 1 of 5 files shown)


julien-tierny:
alright, looks good thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #217, opened 2019-04-05
Fix HarmonicField behavior

pierre-guillou:
There were (at least) two bugs in my HarmonicField filter:
* a pre-processing of the constraints TTK identifiers that caused wrong constraint values on identifiers
* the filter was not reacting to the input constraints type.

This pull request fixes those two behaviors by:
* removing the pre-processing of the constraints identifiers
* linking the computation to the constraints input type
* removing the setting of the filter output type


julien-tierny:
thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #218, opened 2019-04-08
Fixes overflows in FTM

klacansky:


julien-tierny:
thanks a lot pavol!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #219, opened 2019-04-16
[HarmonicField] Fix race condition using thread synchronisation

pierre-guillou:
Should fix non-deterministic harmonic field output with Cholesky factorization and cotan weights method


julien-tierny:
looks good, thanks a lot!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #220, opened 2019-04-17
CMake Indentation

CharlesGueunet:
Dear Julien,

This PR is only about indentation and does not changes anything but codestyle.
As I will work on the TTK CMake on the following days, I prefer to clean the indentation of all CMake files a priori, so that futur diff will be easier to understand.

For the indentation, I have chosen to use spaces, accordingly to the CPP files.

Charles


julien-tierny:
alright, thanks.
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #221, opened 2019-04-26
MSVC build fix

mv740:
Building on Visual Studio 2017 or 2019  will fail because of these two ambiguous calls


julien-tierny:
thanks a lot michal!

julien-tierny:
thanks a lot michal!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #224, opened 2019-05-12
OpenMP Priority improvements

CharlesGueunet:
Dear Julien,

This PR unifies OpenMP priority handling between FTM and FTR. It also improves FTM by computing JT and ST simultaneously when the output is a contour tree.

Charles


julien-tierny:
looks good, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #225, opened 2019-05-14
New module for computing Eigenfunctions of a triangular surface

pierre-guillou:
The current pull-request adds a new module to compute the first eigenfunctions of a triangular mesh, i.e. the eigenvectors corresponding to the largest eigenvalues of the laplacian matrix of the mesh.

To this end, this new module uses the [Spectra](https://spectralib.org/) library to compute eigenvectors of sparse matrices.

Laplacian matrices of triangular meshes are computed inside a new utility module, aptly named Laplacian, refactored from the HarmonicField module. This Laplacian module depends on the [Eigen](http://eigen.tuxfamily.org/index.php?title=Main_Page), as does the Spectra library.

A small cleaning of the HarmonicField module also occured during the development of these new filters.


julien-tierny:
looks good, awesome!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #226, opened 2019-05-15
Quick fixes after playing with MSVC

pierre-guillou:
The current pull-request has small build fixes for MSVC compatibility:

1. Separating compilation flags between GCC/Clang and MSVC and reducing warning level from /Wall to /W4;
2. Fixing a link error in my new EigenField filter concerning some unimplemented methods;
3. Adding some missing pre-processor guards around compiler specific pragmas in the FTRGraph filter that caused a MSVC warning.

### DIFF
--- a/core/base/ftrGraph/FTRGraphPrivate_Template.h
+++ b/core/base/ftrGraph/FTRGraphPrivate_Template.h
@@ -894,8 +894,10 @@ namespace ttk {
         }
       }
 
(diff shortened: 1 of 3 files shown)


julien-tierny:
looks good, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #227, opened 2019-05-19
Temporary fix for FTM

CharlesGueunet:
Dear Julien,

It seems the simultaneous computation of ST and JT still bring some bug on certain systems.
I can't investigate this bug today, but I will come back to it later.
Hope this will fix the issue.

Charles

### DIFF
--- a/core/base/ftmTree/FTMTree_CT.cpp
+++ b/core/base/ftmTree/FTMTree_CT.cpp
@@ -86,15 +86,15 @@ void FTMTree_CT::build(TreeType tt)
 #endif
        {
           if (tt == TreeType::Join || bothMT) {
-#ifdef TTK_ENABLE_OPENMP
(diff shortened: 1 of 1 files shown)


julien-tierny:
thanks charles,
it seems that the issue comes from the openmp priorities activated by default. I will simply restrict the test (for openmp 5 only). I'm using gcc 8.2.0. cheers,
julien
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
gcc 8.2.0 seems to report the version 5 of omp.
I'll disable priorities altogether for now but this needs to be investigated.
julien
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #228, opened 2019-05-24
Fix warnings found by the Clang Static Analyzer

pierre-guillou:
Static Analysis is a methodology for finding potential software bugs without executing the program. It is currently a research subject in computer science (c.f. the LIP6 [MOPSA Project](http://mopsa.lip6.fr/)). Static Analysis tools detect, in C or C++, bugs around memory use, such as:

- garbage or undefined values,
- uninitialized variables,
- NULL pointer de-references,
- memory leaks,
- dead stores (writes into variables that are never read).

Some static analyzers have reached maturity and can be easily used on a given codebase. One of them is the [Clang Static Analyser](https://clang.llvm.org/docs/ClangStaticAnalyzer.html) which is part of the Clang/LLVM tool suite (in the `clang-tools` package on Ubuntu). The Clang Static Analyzer can be invoked on a source file using the `clang-tidy` front-end or on the whole project with the `scan-build` front-end (in conjunction with CMake) with the following commands:

```sh
# go to the project sources and create a build directory
mkdir build && cd build
# set the configuration options
scan-build ccmake ..
# analyze the project according to the current configuration
scan-build make # -j$(nproc) to speed up the process
```
The output of the previous set of commands is a navigable HTML document that lists all detected defects, and for every one of them, the path (annotated source code) that lead to this bug.

Applied on TTK (on the default configuration), the Clang Static Analyzer raise ~100 warnings, a few of them (~5) coming from external libraries (Boost, Eigen). This pull request aims at fixing all the other ones, mostly by providing additional (guarded) null checks, variable initializations and removing dead stores.

### DIFF
--- a/core/base/contourForests/ContourForests.cpp
+++ b/core/base/contourForests/ContourForests.cpp
@@ -564,7 +564,6 @@ void ContourForests::unifyTree(const char treetype)
 
       for (idSuperArc upArcPos = 0; upArcPos < nbUpArc; upArcPos++) {
           // reset partition / node and tree
-         currentPartition = refPartition;
          currentTree      = getTreePart(refPartition);
          currentNodeId    = refNodeId;
          currentNode      = currentTree->getNode(currentNodeId);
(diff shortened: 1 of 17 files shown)


julien-tierny:
ok, thanks.

PR #229, opened 2019-05-24
Activate -Wextra flag for GCC and Clang, fix subsequent warnings

pierre-guillou:
In order to maintain the  current level of software quality in the TTK project, I propose to raise the compilation flags with GCC and Clang from `-Wall` to `-Wall -Wextra`. The `-Wextra` flag activates a set of compiler warnings stricter than `-Wall` alone.

In TTK, two kinds of warnings are affected by this flag:

- `-Wtype-limits`: unsigned integers are always greater than 0, so comparisons to 0 can be removed (2 commits),
- `-Wunused-parameters`: function parameters that are unused in the function body (the other commits).

Unused parameters warnings can be fixed through different solutions:

- using compiler attributes syntax, such as `__attribute__((__unused__))`, to mark the parameter declaration in the function definition. Those attributes are not portable and have to be factored into a preprocessor macro such as: 
```c
#if defined(__GNUC__) || defined(__clang__)
#define TTK_UNUSED __attribute__((unused))
#else // !__GNUC__ && !__clang__
#define TTK_UNUSED
#endif // __GNUC__ || __clang__
```

- raising the C++ standard to C++17 to use the new `[[maybe_unused]]` standard attribute,
- inserting no-op statements using the offending parameter, such as casting it to void:
```c
#define TTK_UNUSED(X) (void)(X)
```

- commenting out the parameter name in the function definition, which is surprisingly portable,
- silencing this particular warning with `-Wno-unused-parameter` (too easy!).

The solution that lead to minimal changes to the code without silencing totally this warning was commenting out parameters name using C-style block comments (`/* */`).  In case the parameter is  used in a `#ifdef ... #endif`, I chose to insert a no-op statement in the `#else ... #endif` using a `TTK_UNUSED` macro defined in `core/base/common/Os.h`.


julien-tierny:
thanks, but this does not look necessary (or even desirable) for ttk at the moment.
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #230, opened 2019-06-06
Adding #include<array> to fix a compilation bug on line 128

jalevine:
On MacOS the line:

std::array<float, 9> coords{};

Leads to the compiler complaining with:

/Users/josh/git/github/ttk/core/base/laplacian/Laplacian.cpp:128:28: fatal error: implicit instantiation of undefined template
      'std::__1::array<float, 9>'
      std::array<float, 9> coords{};

Which is fixed by #include<array>

### DIFF
--- a/core/base/laplacian/Laplacian.cpp
+++ b/core/base/laplacian/Laplacian.cpp
@@ -4,6 +4,8 @@
 #ifdef TTK_ENABLE_EIGEN
 #include <Eigen/Sparse>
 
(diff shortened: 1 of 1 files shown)


julien-tierny:
awesome, thanks!
julien
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #233, opened 2019-06-17
[core] CMake, include headers in target library

martinfalk:
* include header files in Visual Studio projects so they show up in the VS solution as well


julien-tierny:
thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

Issue #232, opened 2019-06-16
Constness in MorseSmaleComplex


martinfalk:
What is the reason behind the non-const arguments in some functions for initializing a `MorseSmaleComplex`? 

For example:
https://github.com/topology-tool-kit/ttk/blob/756d9fdc670d319e63509fd6643e7cbaadea5015/core/base/abstractMorseSmaleComplex/AbstractMorseSmaleComplex.h#L256

Using `void* const data` here breaks constness when trying to hand in a pointer to a const object.

GuillaumeFavelier:
The priority in this case is to ensure that the pointer itself called `data` in here cannot be modified, the intention is not to break const-correctness.
I agree with you, the use case with a pointer to const is not handled. Unfortunately, I can't evaluate myself right now the influence of a potential modification to `const void*`, what do you think @pierre-guillou?

martinfalk:
OK, I will give it a try tomorrow and check out the potential implications.

GuillaumeFavelier:
Thanks for your thorough investigation @pierre-guillou !

The second scenario looks like a good comprise but I wonder if it's okay for @julien-tierny  and if it suits @martinfalk 's use case. Anyway, the first scenario looks good to me. Who would take care of it?
Personally, I still have to catch up with the latest master on my config but once it's done, I could tackle it.

pierre-guillou:
I spent a few minutes this morning to add const qualifiers to the class members `ttk::AbstractMorseSmaleComplex::inputScalarField_` and `ttk::AbstractMorseSmaleComplex::inputOffsets_`. Fortunately, these two pointers are never written.

I'll still have some checks to perform before making a PR.

martinfalk:
> To enforce const-correctness of pointed data, we'll have to const-qualify the class member inputScalarField_ and also every function at which this pointer is passed.

I think that is the way to go. See also https://isocpp.org/wiki/faq/const-correctness

Small side note: I only used `setInputScalarField()` as one example. There are more functions like this with `* const` arguments. And not only in the Morse-Smale project...
My point being that the argument to the function should be a pointer to a const object. Making it a const pointer can happen within the function, i.e. saving the object as `const <type> * const`.

It appears that there might some confusion between `void * const` vs. `const void *` (or alternatively `void const *`).
https://isocpp.org/wiki/faq/const-correctness#const-ptr-vs-ptr-const

julien-tierny:
PR #234 should have fixed this issue.
thanks!
[...]

PR #234, opened 2019-06-21
[MorseSmaleComplex] Const-qualify class members

pierre-guillou:
Enforce const correctness for buffers pointed by the inputScalarField_ and inputOffsets_ class members.

See #232.


julien-tierny:
alright, looks good. 
thanks
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #236, opened 2019-06-24
Outline vtkTemplateMacro arguments into functions

pierre-guillou:
This PR aims at fixing some portability issues arising around `vtkTemplateMacro` calls by outlining the macro arguments into proper functions.

Using macro directives (`#ifdef`, `#define`…) inside macro calls is not portable although accepted by GCC and Clang (with a lot of -Wpedantic warning though).

On the contrary, MSVC seems not to support those kind of constructs. This case arises with the use of the `vtkTemplateMacro` macro and its derivatives throughout `core/vtk` in the TTK codebase. The previous workaround was to write three times the same macro argument, one for GCC and Clang, and two for MSVC, like the following example:
```c++
vtkTemplateMacro({
#ifndef _MSC_VER
  foo();
#ifdef TTK_ENABLE_KAMIKAZE
  bar();
#endif 
#else // _MSC_VER
#ifdef TTK_ENABLE_KAMIKAZE
  foo();
  bar();
#else 
  foo();
#endif 
#endif // _MSC_VER
});
```
Outlining the macro argument will reduce code duplication while increasing portability:

```c++
void baz() {
  foo();
#ifdef TTK_ENABLE_KAMIKAZE
  bar();
#endif 
}
...
vtkTemplateMacro(baz());
```

The proposed changes will also allow us to get rid of the `ttkTemplateMacro` and `TTK_COMMA` macros.

### DIFF
--- a/core/vtk/ttkBarycentricSubdivision/ttkBarycentricSubdivision.cpp
+++ b/core/vtk/ttkBarycentricSubdivision/ttkBarycentricSubdivision.cpp
@@ -117,7 +117,7 @@ int ttkBarycentricSubdivision::doIt(std::vector<vtkDataSet *> &inputs,
     outputScalarField->SetName(inputScalarField->GetName());
     output->GetPointData()->AddArray(outputScalarField);
     switch(inputScalarField->GetDataType()) {
-      ttkTemplateMacro(baseWorker_.interpolateCellDataField<VTK_TT>(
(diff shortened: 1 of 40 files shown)


julien-tierny:
looks ok!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #237, opened 2019-06-24
Better memory management

pierre-guillou:
This PR aims at replacing uses of manual memory management (raw pointers, `malloc`, `new`) by safe wrappers or workarounds, such as using RAII containers (`std::vector`), or simply transforming heap-allocated variables into stack-allocated ones.

NB: TTK still contains some occurrences of manual memory management, this PR only focuses on the most blatant ones.


GuillaumeFavelier:
This is awesome work, thanks for the memory overhaul @pierre-guillou ! RAII whenever possible in general can only help to avoid memory-related errors very tricky to debug. LGTM

julien-tierny:
ok.
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #238, opened 2019-06-24
Fix -Wshadow warnings in TTK

pierre-guillou:
Shadowing occurs when two variables sharing the same name have overlapping lifetime, i.e:

```c++
int a = 0;
{
  double a = 1.0;
}
```
This can lead to tricky bugs when using a variable that is not defined as the developer thinks it is.


julien-tierny:
why not. this could help debugging.
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #239, opened 2019-06-27
Fix memory leak in the FTM wrapper

CharlesGueunet:
Dear Julien,

This PR fix a memory leak in the FTM wrapper. This was introduced by the hack to deal with multiple connected components, one of the array did not clean its memory.

Charles


julien-tierny:
looks good, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #240, opened 2019-06-28
Missing include for gcc

martinfalk:
missing include for parallel sort in gcc (noticed in gcc 7.4 and 8.3)
```
06:03:25  .../ttk/core/base/ftmTree/FTMTree_MT_Template.h: In member function 'void ttk::ftm::FTMTree_MT::sortInput()':
06:03:25  .../ttk/core/base/ftmTree/FTMTree_MT_Template.h:59:4: error: '__gnu_parallel' has not been declared
06:03:25      __gnu_parallel::sort(sortedVect->begin(), sortedVect->end(), indirect_sort);
06:03:25      ^~~~~~~~~~~~~~
```

### DIFF
--- a/core/base/ftmTree/FTMTree_MT_Template.h
+++ b/core/base/ftmTree/FTMTree_MT_Template.h
@@ -17,6 +17,9 @@
 #define FTMTREE_MT_TPL_H
 
 #include <functional>
+#if (defined(__GNUC__) && !defined(__clang__))
+#include <parallel/algorithm>
+#endif
 
 #include "FTMTree_MT.h"
 


julien-tierny:
awesome, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

pierre-guillou:
I see that `parallel/algorithm` is already included in `FTMTree_MT.h`:
https://github.com/topology-tool-kit/ttk/blob/951eaf28a129c70e4d4afdafd00a308ee574388d/core/base/ftmTree/FTMTree_MT.h#L36

Maybe some conditions are missing in the surrounding nested `#ifdefs`? 

Or a bug with circular includes? `FTMTree_MT.h` is included in `FTMTree_MT_Template.h`, itself included at the end of `FTMTree_MT.h`.

martinfalk:
@pierre-guillou good point. I guess there are some conditions missing (which I've missed). We run the gcc compilation on MacOS... and there is no compiler distinction for `apple`

`ttk/core/base/ftmTree/FTMTree_MT.h`:
```c++
#ifdef __APPLE__
#include <algorithm>
#include <numeric>
#else
#ifdef _WIN32
#include <algorithm>
#include <numeric>
#else
#ifdef __clang__
#include <algorithm>
#include <numeric>
#else
#include <parallel/algorithm>
#endif
#endif
#endif
```

petersteneteg:
That block should be equivalent to 
```
#include <algorithm>
#include <numeric>
#ifdef __GNUC__
#include <parallel/algorithm>
#endif
```

Seems quite bad to mix checks for OS versions and compilers. When can use Clang/GCC on windows, or gcc on OSX and so on. better check for a compiler. especially when it is a compiler feature that is checked for.

PR #242, opened 2019-07-01
Fix build warnings/errors with Clang in ttkContourAroundPoint

pierre-guillou:
Building ttkContourAroundPoint with Clang has been failing with `'vtkPointData' is an incomplete type` error. 

To fix this, `vtkPointData.h` has been included.

Other minor warnings displayed by Clang have also been fixed for this new module.


julien-tierny:
ok.

PR #250, opened 2019-07-26
[Quadrangulation] Fix Travis CI with old VTK

pierre-guillou:
Use GetNumberOfTuples instead of GetNumberOfValues


julien-tierny:
perfect, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #251, opened 2019-07-30
Including VTK_USE_FILE only when it is not empty.

peterghristov:

- [x] Review our [Contributor Guidelines](https://github.com/topology-tool-kit/ttk/blob/master/CONTRIBUTING.md), in particular regarding code formatting (with clang-format) and continuous integration.

Your description here

Hi,

I had the following two cmake errors on MacOS Mojave 10.14.5.

CMake Error at standalone/CMakeLists.txt:15 (include): include called with wrong number of arguments.  include() only takes one file.

CMake Error at core/vtk/CMakeLists.txt:20 (include): include called with wrong number of arguments.  include() only takes one  file.

The relevant line in both files that causes these problems is include(${VTK_USE_FILE}), and the issue is that the cmake variable VTK_USE_FILE was empty.

I can confirm that ttk compiles fine with the changes from this commit on OSX Mojave 10.14.5.

### DIFF
--- a/core/vtk/CMakeLists.txt
+++ b/core/vtk/CMakeLists.txt
@@ -17,7 +17,11 @@ if(NOT DEFINED VTK_USE_FILE)
     message(STATUS "No VTK provided, fallback to ParaView")
   endif()
 endif()
-include(${VTK_USE_FILE})
+
+if(DEFINED VTK_USE_FILE)
+    include(${VTK_USE_FILE})
+endif()
+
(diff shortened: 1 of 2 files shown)


julien-tierny:
hi peter,
awesome, looks good!
thanks a lot 
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #243, opened 2019-07-03
Para view57

CharlesGueunet:
Dear Julien,
This PR  changes the CMake architecture in order to build TTK against the current versions of VTK & ParaView.

These changes does not impact how the user build TTK (except that he should have an up to date VTK/ParaView of course).

As you can see [here](https://markmail.org/thread/cu2wyqpvporqvps6), important changes have been made on the VTK build system, impacting ParaView and every VTK based applications. The way VTK is linked to a project is now more modular and relies on a `vtk.module` file in order to declare dependencies. I have used a similar syntax for TTK, using a `ttk.module` file.

I will update the Guideline to explain in detail the new TTK architecture.

Charles

PR #256, opened 2019-08-20
VTK Python module

lhofmann:
This PR adds automatically generated Python wrappers for the VTK layer of TTK.

A CMake option `TTK_BUILD_VTK_PYTHON_MODULE` is added, which when turned on builds a Python module `ttk`, which wraps all ttk vtk wrappers and links against them. The Python module is built in the subdirectory `lib/python-x.y/site-packages` of the build path and installed into the same subdirectory of the install prefix path.

It makes the following use-cases possible:
* Use ttk from python without the ParaView dependency (an example is provided in `examples/vtk-python/example.py`)
* Package ttk with conda-forge
  * A working conda build (linux only) can be found here: https://github.com/lhofmann/staged-recipes/tree/ttk (`.circleci/run_docker_build.sh` can be used to perform a local build)
* When installed into the ParaView tree, ttk can be used within Programmable Filters and [ParaView python plugins](https://gitlab.kitware.com/paraview/paraview/merge_requests/2516)

Issues:
* Supports only VTK 8 (tested with 8.2.0). Supporting versions 6 and 7 would require substantial adjustments in all ttk header files, since the VTK 6 parser does not support C++11 and neither support classes that do not derive from vtkObject. This could be worked around by adding a large number of `#if !defined(__WRAP__) || !defined(__WRAP_VTK__) .. #endif` directives around offending syntax in the ttk header files.
* In Python 2.7, `import ttk` for some reason imports `tkinter.ttk`, i.e. the ttk python module would need to be renamed to something else (e.g. `topologytoolkit`)
* The presence of `VTK_INSTALL_PREFIX` is currently assumed for finding VTK's python module path

The future VTK 9/ParaView 5.7 (#243) can be supported with much less CMake code (using the new `vtk_module_wrap_python` CMake function).


julien-tierny:
hi lutz,

thanks a lot for your contribution to ttk!

that looks fantastic.

a few remarks / suggestions:

  - thanks for inserting at core/functions.cmake:27 the following message line (for the configuration summary)
message(STATUS, "TTK_BUILD_VTK_PYTHON_MODULE: ${TTK_BUILD_VTK_PYTHON_MODULE}")

  - your pr made travis-ci hiccup. the vm they provide ships with an old ubuntu (16.04 I believe) that has only vtk 6. I'd suggest:
    - to only enable TTK_BUILD_VTK_PYTHON_MODULE for recent detected versions of vtk (8 at least?)
    - provide a binary install of a recent version of vtk in travis-ci's vm (you may want to put together a docker image or something like this)

  - regarding examples/vtk-python/, this is great that you put that together. did you run it to make sure it provided the exact same output as examples/python? (it should if all default values are consistent). if so, I wonder if it should not simply replace the existing directory examples/python/. what do you think? 

  - can you add a file examples/vtk-python/README (in essence a copy+adapt of examples/python/README)? also, is the conda build currently usable for general users? if so could you document how to use it in examples/vtk-python/README?

  - regarding the issues with python 2.7, I can live with it. I believe 2.7 is deprecated and we officially recommend 3.6 for scikit-learn support

  - regarding VTK_INSTALL_PREFIX, this is what we already use I believe, so I can live with it. that does not seem sustainable though (absent from pv5.7, btw: any idea for a workaround on pv5.7?)

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

lhofmann:
I moved to option `TTK_BUILD_VTK_PYTHON_MODULE ` to core/vtk. It is now enabled by default if VTK 8 with python wrapping is detected. In case the variable was manually defined in the cmake cache, a VTK version check is in place.
The python module is now installed in $CMAKE_INSTALL_PREFIX/$TTK_PYTHON_MODULE_DIR. `TTK_PYTHON_MODULE_DIR` is automatically detected from the VTK config, but can also be set in cmake cache.
Both variables are printed by `ttk_print_summary`.

>  regarding examples/vtk-python/, this is great that you put that together. did you run it to make sure it provided the exact same output as examples/python? (it should if all default values are consistent). if so, I wonder if it should not simply replace the existing directory examples/python/. what do you think?

I compared it against the vtk-c++ example and both produce identical outputs (matching md5 sums). The example is also run during the testing stage of the conda build.

I guess some users would still want to use ParaView's python interface, as it has a more concise interface and can be automatically generated using ParaView's trace functionality. Therefore, I'm keeping the vtk-python directory for now.

> can you add a file examples/vtk-python/README (in essence a copy+adapt of examples/python/README)? also, is the conda build currently usable for general users? if so could you document how to use it in examples/vtk-python/README?
 
I added a README. The conda build is not available yet. I would wait for this PR to be merged and then change the build to pull directly from the ttk repository (and not my fork). 
There is some sort of review process involved until the package will finally be available on conda-forge (https://conda-forge.org/#contribute). I would update the REAME once this is done.

> regarding the issues with python 2.7, I can live with it. I believe 2.7 is deprecated and we officially recommend 3.6 for scikit-learn support

I still renamed the python module to topologytoolkit, since conda-forge requires the package working on Python 2.7. It can be used as `import topologytoolkit as ttk`.

>  regarding VTK_INSTALL_PREFIX, this is what we already use I believe, so I can live with it. that does not seem sustainable though (absent from pv5.7, btw: any idea for a workaround on pv5.7?)

I believe this variable was renamed to `VTK_PREFIX_PATH` in 5.7. I have added some more logic, that uses `VTK_DIR` or `ParaView_DIR` in the cases, where VTK/ParaView's build trees are used.

julien-tierny:
hi lutz,

excellent. that looks pretty much ready to go.

> I moved to option `TTK_BUILD_VTK_PYTHON_MODULE ` to core/vtk. It is now enabled by default if VTK 8 with python wrapping is detected. In case the variable was manually defined in the cmake cache, a VTK version check is in place.
awesome.

> The python module is now installed in $CMAKE_INSTALL_PREFIX/$TTK_PYTHON_MODULE_DIR. `TTK_PYTHON_MODULE_DIR` is automatically detected from the VTK config, but can also be set in cmake cache.
> Both variables are printed by `ttk_print_summary`.
perfect.

> I compared it against the vtk-c++ example and both produce identical outputs (matching md5 sums). The example is also run during the testing stage of the conda build.
perfect.

> I added a README. 
perfect.

> I guess some users would still want to use ParaView's python interface, as it has a more concise interface and can be automatically generated using ParaView's trace functionality. 
that's a good point.

> Therefore, I'm keeping the vtk-python directory for now.
how about the following:
- renaming examples/python into examples/pvpython
- and renaming examples/vtk-python into examples/python

somehow, I feel that we should clearly identify one of the two APIs as the default one (otherwise it could be confusing for new users) and I believe your new implementation is a better candidate (no need for paraview and it could be pulled by conda). 
what do you think?

> The conda build is not available yet. I would wait for this PR to be merged and then change the build to pull directly from the ttk repository (and not my fork). 
> There is some sort of review process involved until the package will finally be available on conda-forge (https://conda-forge.org/#contribute). 
understood.

> I would update the REAME once this is done.
that'd be awesome.

> I still renamed the python module to topologytoolkit, since conda-forge requires the package working on Python 2.7. It can be used as `import topologytoolkit as ttk`.
sounds good.

-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
hi lutz,

excellent. that looks pretty much ready to go.

> I moved to option `TTK_BUILD_VTK_PYTHON_MODULE ` to core/vtk. It is now enabled by default if VTK 8 with python wrapping is detected. In case the variable was manually defined in the cmake cache, a VTK version check is in place.
awesome.

> The python module is now installed in $CMAKE_INSTALL_PREFIX/$TTK_PYTHON_MODULE_DIR. `TTK_PYTHON_MODULE_DIR` is automatically detected from the VTK config, but can also be set in cmake cache.
> Both variables are printed by `ttk_print_summary`.
perfect.

> I compared it against the vtk-c++ example and both produce identical outputs (matching md5 sums). The example is also run during the testing stage of the conda build.
perfect.

> I added a README. 
perfect.

> I guess some users would still want to use ParaView's python interface, as it has a more concise interface and can be automatically generated using ParaView's trace functionality. 
that's a good point.

> Therefore, I'm keeping the vtk-python directory for now.
how about the following:
- renaming examples/python into examples/pvpython
- and renaming examples/vtk-python into examples/python

somehow, I feel that we should clearly identify one of the two APIs as the default one (otherwise it could be confusing for new users) and I believe your new implementation is a better candidate (no need for paraview and it could be pulled by conda). 
what do you think?

> The conda build is not available yet. I would wait for this PR to be merged and then change the build to pull directly from the ttk repository (and not my fork). 
> There is some sort of review process involved until the package will finally be available on conda-forge (https://conda-forge.org/#contribute). 
understood.

> I would update the REAME once this is done.
that'd be awesome.

> I still renamed the python module to topologytoolkit, since conda-forge requires the package working on Python 2.7. It can be used as `import topologytoolkit as ttk`.
sounds good.

> I believe this variable was renamed to `VTK_PREFIX_PATH` in 5.7. 
I see (@CharlesGueunet you may want to give it a try)

> I have added some more logic, that uses `VTK_DIR` or `ParaView_DIR` in the cases, where 
> VTK/ParaView's build trees are used.
we've been discussing this with @CharlesGueunet for his pv57 branch. I have mixed feelings about maintaining both implementations and I would vote to only consider installed versions of vtk/paraview (see https://www.google.com/url?q=https://discourse.paraview.org/t/paraview-install-development-files-still-used-in-paraview-5-7/2456&sa=D&source=hangouts&ust=1566577025725000&usg=AFQjCNHdP24jCmks76spS0VnUrTzDZZKGg)

could you consider renaming the python example directories as suggested above?
other than that, it looks ready to go :)

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

lhofmann:
> somehow, I feel that we should clearly identify one of the two APIs as the default one (otherwise it could be confusing for new users) and I believe your new implementation is a better candidate (no need for paraview and it could be pulled by conda).
what do you think?

Agreed

> could you consider renaming the python example directories as suggested above?

done

julien-tierny:
awesome, thanks a lot!

thanks for letting me know about the next steps with conda-forge.
cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #258, opened 2019-08-22
Fix clang -Wsometimes-uninitialized warning

JochenJankowai:
Initialize ```startId``` in ```DiscreteGradient_Template.h:1666``` to fix clang ```-Wsometimes-uninitialized``` warning.


julien-tierny:
looks good too. thanks!

Issue #257, opened 2019-08-22
PR #236 introduced bugs in point data arrays copies


julien-tierny:
**Describe the bug**
The newly suggested management for the copy of the input point data arrays onto the generated outputs introduced new bugs. 
Some data arrays are no longer copied. This in particular verifiable with the default ttk example states/viscousFingering.pvsm from the ttk-data repository.
In that example, the list of output point data arrays for TTKScalarFieldCriticalPoints2 is:
- concentration
- CriticalType
- IsOnBoundary
- ttkVertexScalarField
- velocity
- vtkValidPointMask

While it should (and used to) be:
- concentration
- CriticalType
- height
- IsOnBoundary
- OutputDistanceField
- RegionId
- SeedIdentifier
- ttkVertexScalarField
- velocity
- VertexNumber
- vtkValidPointMask

An instance of consequence is that the "finger tips" (large spheres) are no longer colored by seed identifier. See the following screenshots.
before:
![before](https://user-images.githubusercontent.com/3018628/63497165-8aed8700-c4c3-11e9-9d81-273bb260095c.png)
after:
![after](https://user-images.githubusercontent.com/3018628/63497173-8f19a480-c4c3-11e9-9aa5-029c6b1c8b8c.png)

**To Reproduce**
Steps to reproduce the behavior (from the ttk-data directory):
$ paraview states/viscousFingering.pvsm

**Expected behavior**
Described above.

**Screenshots**
Given above

**System (please complete the following information):**
 - OS: [e.g. iOS] gentoo linux
 - Compiler [e.g. 22] gcc 8.2.0
 - Installed TTK's optional dependencies: all

**Additional context**
this one is for @pierre-guillou who authored PR #236

pierre-guillou:
I can reproduce this behavior on my laptop.
After some bisection, the culprit seems to be afdc4f2b.
Fix very soon.

julien-tierny:
PR #260 seems to solve the issue.
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #260, opened 2019-08-22
Fix control flow in ScalarFieldCriticalPoint

pierre-guillou:
Commit afdc4f2b in PR #236 replaced a `break` statement with a `return` one, skipping later statements to allow the outlining of the switch body instructions after this `return`. This `return` statement also skipped some valid statements, leading to issue #257.

This current PR fixes issue #257 by refactoring the outlined switch body into a lambda called in every case of the switch construct. It also replaces the faulty `return` by a `break`.

Fix issue #257.
Sorry for the inconvenience.

Pierre

### DIFF
--- a/core/vtk/ttkJacobiSet/ttkJacobiSet.cpp
+++ b/core/vtk/ttkJacobiSet/ttkJacobiSet.cpp
@@ -240,51 +240,60 @@ int ttkJacobiSet::doIt(vector<vtkDataSet *> &inputs,
       vtkDataArray *scalarField = input->GetPointData()->GetArray(i);
       vtkSmartPointer<vtkDataArray> scalarArray;
 
+      auto copyToScalarArray = [&]() {
+        scalarArray->SetNumberOfComponents(
+          scalarField->GetNumberOfComponents());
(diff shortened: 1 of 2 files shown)


julien-tierny:
hi pierre,

thanks for the quick fix.

this structure seems to be present in other modules too (at least core/vtk/ttkJacobiSet/ttkJacobiSet.cpp). can you please double-check everything is fine for those too?

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

pierre-guillou:
Hi Julien,

the exact same structure, before and after PR #236, was present in ttkJacobiSet.cpp. 

As a consequence, I applied the same fix as in ttkScalarFieldCriticalPoints.cpp.

Sorry again for the inconvenience,

Pierre

julien-tierny:
hi pierre,

thanks for the quick fix.

this looks good.

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #264, opened 2019-09-03
Fix building the Python module on Windows

lhofmann:
This PR contains minor fixes, that are needed for building on Windows:
* `VTK_PYTHONPATH` appears to be a relative path on Windows
* Python module needs suffix .pyd on Windows


julien-tierny:
looks good, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #265, opened 2019-09-05
Fix building/running MorseSmaleQuadrangulation filters with 64BIT_IDS

pierre-guillou:
After testing the quadrangulation with TTK_ENABLE_64BIT_IDS set, I noticed one build error and several run-time issues with data arrays.

This PR should fix those problems.


julien-tierny:
ok.

PR #272, opened 2019-09-27
Triangulation copy/move constructors

### LINKED ISSUE #271: Triangulation missing/wrong copy/move constructor and assignment operators
Triangulation.h has only have a default constructor 
https://github.com/topology-tool-kit/ttk/blob/0b0d78e91367e433cdb7b94c17b931a9e71c6ea7/core/base/triangulation/Triangulation.h#L49
which means that it gets the default member wise copy/move constructors and assignment operators. But since Triangulation keeps a pointer
https://github.com/topology-tool-kit/ttk/blob/0b0d78e91367e433cdb7b94c17b931a9e71c6ea7/core/base/triangulation/Triangulation.h#L3205 
that points inside of itself, those defaults will not work!

petersteneteg:
Fixes the copy/move constructors. Added override to overridden functions. Added final to classes to help the compiler devirtualize.

Closes #271

### DIFF
--- a/core/base/explicitTriangulation/ExplicitTriangulation.h
+++ b/core/base/explicitTriangulation/ExplicitTriangulation.h
@@ -19,7 +19,7 @@
 
 namespace ttk {
 
-  class ExplicitTriangulation : public AbstractTriangulation {
+  class ExplicitTriangulation final : public AbstractTriangulation {
 
   public:
     ExplicitTriangulation();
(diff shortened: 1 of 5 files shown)


julien-tierny:
alright peter. that looks good to me!
thanks a lot!
cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #249, opened 2019-07-26
Add surface quadrangulation filter

pierre-guillou:
This pull-request allows TTK to generate quadrangular surfaces (surfaces composed of quadrangles) from traditional triangular surfaces.
 
Three new filters have been developped:

- MorseSmaleQuadrangulation generates a coarse-grain quadrangulation from Morse-Smale critical points
- QuadrangulationSubdivision refines a coarse-grain quadrangulation by subdividing it and projecting it on the original surface
- BarycentricSubdivision is a utility filter providing a subdivided triangulation. Each triangle is divided into 6 new triangles whose vertices are those of the parent triangulation, the middles of the triangle edges, and the triangle barycenter.

Enjoy :)

PR #273, opened 2019-10-04
Barycentric Subdivision: Fixes, Subdivision Level

pierre-guillou:
The current PR focuses on the BarycentricSubdivision, which was introduced for the Morse-Smale Quadrangulation (#249).

This PR:
- tweaks the filter memory model to make the VTK layer own the data (instead of the base layer);
- adds support for a variable number of subdivision levels and interpolate the input data accordingly (using std::swap/move semantics to avoid copies);
- refactor methods in the VTK wrapper (and in the base layer too) to allow for said subdivision levels;
- fixes some issues with unsupported data types and incorrect number of triangles.

### DIFF
--- a/core/base/barycentricSubdivision/BarycentricSubdivision.cpp
+++ b/core/base/barycentricSubdivision/BarycentricSubdivision.cpp
@@ -15,11 +15,7 @@ int ttk::BarycentricSubdivision::subdiviseTriangulation() {
     return 1;
   }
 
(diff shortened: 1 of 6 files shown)


julien-tierny:
looks good, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #275, opened 2019-10-15
Moving GetCellIncenter to AbstractTriangulation

petersteneteg:
Hi
I propose moving the GetCellIncenter and related functions from DiscreteGradient into AbstractTriangulation. 
The reason being that the PeriodicImplicitTriangulation needs to customize the behavior, and DiscreteGradient only delegated the to the triangulation without using any of the DiscreteGradient  state.

The reason that the PeriodicImplicitTriangulation need to customize the functions can be understod if one considers an edge at the "boundary" of the grid, that edge will connect a vertices on both sides of the grid, and the GetCellIncenter will by default, return the mean position of the vertices, which will be in the middle of the grid. In the current proposal, that edge center point will be put in the higher side of the grid, after the last vertex. Same goes for the other types of cells.

### DIFF
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -13,6 +13,7 @@
 #define _ABSTRACTTRIANGULATION_H
 
 // base code includes
+#include <Geometry.h>
 #include <Wrapper.h>
 
 namespace ttk {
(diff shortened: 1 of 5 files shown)


julien-tierny:
hi peter,

thanks a lot for this pr!

> I propose moving the GetCellIncenter and related functions from DiscreteGradient into AbstractTriangulation. 
all of this makes a lot of sense.
thanks a lot for the generic fix.

cheers,
julien
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #277, opened 2019-10-21
Topological Compression Writer/Reader: Add scalar field name into compressed file

pierre-guillou:
This PR modifies the Topological Compression Writer and Reader to pass the scalar field name into the compressed file.

To do so, I edited the WriteMetaData and the ReadMetaData methods of the TopologicalCompression class to include the char array corresponding to the field name into the meta-data section of the compressed file.

Warning: this is a BREAKING CHANGE for the Reader. An old Reader won't be able to read new files (with the scalar field name). Segfaults expected.

To keep backward compatibility (new Reader reading old files), I also introduced two new meta-data in the file format: some magic bytes ("TTKCompressedFileFormat") and a file format version counter. Testing for the presence of the former let the Reader distinguish between the old and the new file format. The version number should be used to ensure backward compatibility in case of future file format modifications.

Enjoy!
Pierre

### DIFF
--- a/core/base/topologicalCompression/TopologicalCompression.cpp
+++ b/core/base/topologicalCompression/TopologicalCompression.cpp
@@ -14,6 +14,10 @@ ttk::TopologicalCompression::TopologicalCompression() {
 ttk::TopologicalCompression::~TopologicalCompression() {
 }
 
+const std::string ttk::TopologicalCompression::magicBytes_{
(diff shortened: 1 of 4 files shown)


julien-tierny:
alright, looks good!
thanks.
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #280, opened 2019-11-10
Make destructor of base class virtual

dsakurai:
The destructor of a base class should be declared virtual in order to avoid resource leaks.

See here:
https://stackoverflow.com/questions/461203/when-to-use-virtual-destructors

### DIFF
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -23,7 +23,7 @@ namespace ttk {
   public:
     AbstractTriangulation();
 
-    ~AbstractTriangulation();
+    virtual ~AbstractTriangulation();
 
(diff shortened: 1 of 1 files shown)


julien-tierny:
ok, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

CharlesGueunet:
Worry I come here a bit late, but wouldn't `override` be more suitable than `virtual` for a child class ?

dsakurai:
@CharlesGueunet I agree, Charles. Thanks for pointing that out.

It's not mandatory, though.
I only fixed the base class in this PR, while there are a few child classes in TTK...
I hope I will find the time to fix at least the triangulation classes this weekend.

PR #282, opened 2019-11-13
Use Topological Compression in CinemaWriter/CinemaProductReader

pierre-guillou:
This PR add the support of Topological Compression when reading and writing vtkImageData into cinema databases.

Here, the TopologicalCompressionWriter and the TopologicalCompressionReader modules are used as back-ends for the CinemaWriter and CinemaProductReader modules. Some Topological Compression options are re-exposed in the CinemaWriter GUI (under a checkbox).

Also included are a quick refactor of the log/debug message management logic in TopologicalCompression to let it be set from outside, and the support of TIFF images in CinemaProductReader.

Enjoy!

Pierre


julien-tierny:
alright, looks good.
thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

Issue #284, opened 2019-11-20
Compilation error on Windows: ttkPersistenceDiagramClustering.cpp


CharlesGueunet:
**Describe the bug**

Compilation error on Windows:
```
..\core\vtk\ttkPersistenceDiagramClustering\ttkPersistenceDiagramClustering.h(129): warning C4805: '!=': unsafe mix of type 'bool' and type 'int' in operation
..\core\vtk\ttkPersistenceDiagramClustering\ttkPersistenceDiagramClustering.cpp(221): error C2059: syntax error: '{'
..\core\vtk\ttkPersistenceDiagramClustering\ttkPersistenceDiagramClustering.cpp(221): error C2143: syntax error: missing ';' before '{'
..\core\vtk\ttkPersistenceDiagramClustering\ttkPersistenceDiagramClustering.cpp(221): error C2059: syntax error: ')'
..\core\vtk\ttkPersistenceDiagramClustering\ttkPersistenceDiagramClustering.cpp(95): error C2065: 'intermediateDiagrams': undeclared identifier
..\core\vtk\ttkPersistenceDiagramClustering\ttkPersistenceDiagramClustering.cpp(95): error C2046: illegal case
..\core\vtk\ttkPersistenceDiagramClustering\ttkPersistenceDiagramClustering.cpp(95): error C2043: illegal break
..\core\vtk\ttkPersistenceDiagramClustering\ttkPersistenceDiagramClustering.cpp(95): error C2059: syntax error: 'break'
..\core\vtk\ttkPersistenceDiagramClustering\ttkPersistenceDiagramClustering.cpp(95): error C2059: syntax error: 'case'
..\core\vtk\ttkPersistenceDiagramClustering\ttkPersistenceDiagramClustering.cpp(221): error C2447: '{': missing function header (old-style formal list?)
..\core\vtk\ttkPersistenceDiagramClustering\ttkPersistenceDiagramClustering.cpp(222): error C2059: syntax error: '}'
..\core\vtk\ttkPersistenceDiagramClustering\ttkPersistenceDiagramClustering.cpp(222): error C2143: syntax error: missing ';' before '}'
..\core\vtk\ttkPersistenceDiagramClustering\ttkPersistenceDiagramClustering.cpp(225): error C2059: syntax error: '}'
..\core\vtk\ttkPersistenceDiagramClustering\ttkPersistenceDiagramClustering.cpp(225): error C2143: syntax error: missing ';' before '}'
[...]

PR #285, opened 2019-11-20
PersistenceDiagramClustering: outline vtkTemplateMacro body

pierre-guillou:
Apply the logic of PR #236 to the PersistenceDiagramClustering module (which was in development at the time #236 landed).

Should fix issue #284.

The issue here is that the preprocessor logic is different between MSVC and GCC/Clang: the former can directly use blocks (statements inside braces) as macro arguments whereas the latter two expect them inside parentheses. We use to rely on a custom macro to wrap vtkTemplateMacro and add parentheses when needed.

The solution presented here also allows to keep using #ifdef constructs inside the function body (for example for KAMIKAZE checks) without resorting to *triplicated* code blocks (the initial rationale for #236).


julien-tierny:
looks good, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #286, opened 2019-11-21
DimensionReduction: Input is a distance matrix

pierre-guillou:
The DimensionReduction filter wraps several functions in the Scikit-Learn Python toolkit. Instead of passing raw point clouds as an input, two of the wrapped functions, Multi-Dimensional Scaling (MDS) and Spectral Embedding (SE), can work directly on pre-computed dissimilarity (or distance) matrices.

This PR factor this option for the two filters in a checkbox in the GUI.

Besides, this PR allows to select input Row Data columns with a regexp, as an alternative to the list of input columns with checkboxes. This change allows generated Python scripts/ParaView states to be more independent from the input data structure (especially with relation to the number and the names of the input columns).

### DIFF
--- a/core/base/dimensionReduction/DimensionReduction.h
+++ b/core/base/dimensionReduction/DimensionReduction.h
@@ -29,8 +29,13 @@ namespace ttk {
 
     inline int setSEParameters(std::string &Affinity,
                                float Gamma,
-                               std::string &EigenSolver) {
-      se_Affinity = Affinity;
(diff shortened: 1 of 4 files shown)


julien-tierny:
looks good, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #287, opened 2019-11-27
TopologicalCompressionReader: Display logs, fix segfault

pierre-guillou:
This PR aims to fix the ZFP dependency handling inside the TopologicalCompressionReader module: when ZFP is not installed, reading compressed files causes a segfault (a read into an empty vector).

This segfault has been fixed, and a vtkWarning is displayed instead.
I also displayed the TopologicalCompression logs when reading compressed files through CinemaProductReader.

For now, no logs are shown when reading directly .ttk files with ParaView. Should I change that too?

Pierre

### DIFF
--- a/core/base/topologicalCompression/TopologicalCompression.h
+++ b/core/base/topologicalCompression/TopologicalCompression.h
@@ -857,7 +857,7 @@ int ttk::TopologicalCompression::ReadFromFile(FILE *fp) {
     }
   }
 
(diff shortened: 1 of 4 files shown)


julien-tierny:
looks good, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #288, opened 2019-12-13
Support multiple inputs in CinemaQuery

pierre-guillou:
The current PR provides multiple inputs support to the CinemaQuery filter.

As a consequence, one can perform JOIN queries onto several vtkTables.

The input tables are identified from InputTable0 to InputTableN (where N is the total number of input tables minus 1).

The first non-null input FieldData is transferred to the output.

To keep backward compatibility with the current ttk-data states, a step to convert "InputTable" to "InputTable0" has been introduced.

Enjoy,
Pierre


julien-tierny:
looks good, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #292, opened 2019-12-18
Cmake variable and PointDataSelector default

CharlesGueunet:
Dear Julien,

This PR brings two small features:
* Improvement in CMake variables
* Set the default state for the ScalarField in the PointDataSelector to checked (cherry-pick the stable branch)

Cell / Field data selector will be improved in another PR soon.

Charles

PR #293, opened 2019-12-18
WIP: Data selectior unification

CharlesGueunet:
Dear Julien,

NOT Ready yet

This PR depends on #292 .
It brings some improvements on the *DataSelector filter by unifying their filters: CheckBox + Range + Regexp. ( The TableDataSelector still don't have a rename feature).
I have also fixed a bad behavior on the range that could have added more fields that asked by the user if the input is not continuous.

Charles.


julien-tierny:
looks good, thanks!
julien
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #294, opened 2019-12-18
fix(DataSelection): fill the available array in RequestData

CharlesGueunet:
Dear Julien,

This is the end of #293, ttk-data rolls again :)

Charles


julien-tierny:
thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #297, opened 2019-12-20
fix(CMake): hotfix for make condition with ttkAlgorithm

CharlesGueunet:
Dear Julien,

this is an ugly fix for the make race condition.
We still need to investigate why the dependency to this target is not solved correctly.

Charles


julien-tierny:
thanks for the quick fix!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #300, opened 2019-12-21
+ Debug Widgets + HelloWorld + Base Naming Convention + IcoSphere Modules + Azure Hotfix

JonasLukasczyk:
+ Debug Widgets: 'UseAllCores' flag, 'Execute' button
+ HelloWorld: new example and extensive documentation
+ Base Naming Convention: base methods start with lower case
+ IcoSphere Modules: IcoSphere, IcoSphereFromPoint, IcoSphereFromObject
+ Azure Hotfix Attempt: Force IcoSphere build hierarchy


JonasLukasczyk:
The build failed again despite the Azure hotfix.

@CharlesGueunet do you have an idea how to fix this? The problem here is that we have the following dependency hierarchies:
ttkAlgorithm -> ttkIcoSphere -> ttkIcoSphereFromPoint
ttkAlgorithm -> ttkIcoSphere -> ttkIcoSphereFromObject

CharlesGueunet:
The reason the build is broken on azure is because the fix should also be done on mac (the error is random and does not occur frequently on mac).

I am still investigating the *-hierarchy issue.

julien-tierny:
alright, looks good!
cheers
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #301, opened 2019-12-22
ParaViewFilter.cmake now executes preprocessor on XML files

JonasLukasczyk:
This commit enables developers to use preprocessor directives in the ParaView plugin XMLs.

In particular, there exists now a cmake variable called DEBUG_WIDGETS that can be included at the right location inside the XML file with the common syntax ${DEBUG_WIDGETS} (see HelloWorld.xml). This variable will provide the widgets for setting the UseAllCores flag, the ThreadNumber, the DebugLevel, and a button that (re)executes a filter even if parameters didn't change (handy for performance measurements and debugging).

### DIFF
--- a/CMake/ParaViewFilter.cmake
+++ b/CMake/ParaViewFilter.cmake
@@ -1,6 +1,34 @@
 # register a new filter to build in the TTK plugin
 # deduce the location of the corresonding vtk.module file
 # also register the xml file if given
+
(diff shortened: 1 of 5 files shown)


julien-tierny:
alright, looks good.
cheers
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #303, opened 2019-12-29
Debug Updates

JonasLukasczyk:
+ Puts all special output characters (colors, font modes, etc) into a separate namespace (debug::output); In the future this namespace could be renamed to debug::style or something similar.  
+ Adds the option to print the used number of threads in performance msgs:
    [ttkFilter] Computing ................... [0.2s|8T|100%]
+ reorganizes how debug msgs are actually inserted into an output stream.
   Originally, the debug::LineMode was always applied at the beginning of a debug msg, which made it possible to override error and warning msgs. Now, the debug::LineMode is applied at the beginning of a msg and the last used debug::LineMode is tracked internally to make sure that error and warning msgs are always shown correctly. So, in the new version the correct way to use replaceable debug msgs is as follows:

this->printMsg(
  "Computing Something", progress, threadNumber,
  debug::LineMode::REPLACE
);

for(...)
  this->printMsg(
    "Computing Something", progress, threadNumber,
    debug::LineMode::REPLACE
  );
  
this->printMsg(
  "Computing Something", progress, threadNumber,
);

So every time a debug msg is printed with debug::LineMode::REPLACE, one should be aware that this debug msg will be replaced with the next msg (except errors and warnings which will automatically be put in the next line).

Best
Jonas

### DIFF
--- a/core/base/common/Debug.cpp
+++ b/core/base/common/Debug.cpp
@@ -1,5 +1,7 @@
 #include <Debug.h>
 
+ttk::debug::LineMode ttk::Debug::lastLineMode = ttk::debug::LineMode::NEW;
+
 bool ttk::welcomeMsg_ = true;
 bool ttk::goodbyeMsg_ = true;
 int ttk::globalDebugLevel_ = 0;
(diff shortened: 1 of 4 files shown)


julien-tierny:
hi jonas,

looks good, thanks a lot! 

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #302, opened 2019-12-23
Overhauled Cinema Filters

JonasLukasczyk:
CinemaQuery: Multiple table support
CinemaReader: Now automatically prepends database path to FILE columns (removes the need to pass along the absolute database path as a field array and makes it possible to merge multiple tables in the CinemaQuery filter)
CinemaProductReader: Improved memory management
CinemaImaging: Automatic tuning of camera parameters and improved value image generation
CinemaWriter: Self-organizing Cinema databases (replaces existing products and appends new ones) + PNG output support
DepthImageBasedGeometryApproximation: Improved computation
CinemaLayout -> GridLayout: Memory efficient output (only translate and shallow copy where possible)

PR #304, opened 2020-01-03
TU-KL Fork (Step 1/2)

JonasLukasczyk:
Hi everyone,
happy new year!

This pull request contains the conversion of all CinemaFilter to the new architecture, as well as some base features that up to now only existed in the TU-KL fork. Here is a detailed list:

* New Modules:
    + Extract: filter to extract blocks of a multi-block object or geometry of an unstructured grid based on a logical expression 
    + ForEach: Option to iterate over points, cells, array values, blocks, and rows
    + ArrayEditor: A filter to add arrays parsed from a string, to add arrays from an optional second output, to remove arrays, or to edit array properties  
* Cinema: Conversion and update of all Cinema filters (including ttk topological compression support introduced in earlier commits)
* CinemaLayout has been replaced with GridLayout
* ForEachRow filter has been replaced with the more general ForEach filter
* Remaining modules of this pull request have been converted to the new architecture

Best
Jones


JonasLukasczyk:
Hi,
I checked the error logs and I will investigate why Windows can not find the ttkUtils dependency. On Mac the problem is the old race condition. Btw. I will also open a pull request on ttk-data dev.

julien-tierny:
hi jonas,

thanks a lot!

it seems that the remarks of PR #302 also apply here, so @pierre-guillou should also have a look at it (I guess next week).

cheers
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

JonasLukasczyk:
Hi Julien,
seems like the latest Windows fix worked.

As you suggested I incorporated @pierre-guillou features in the new code, i.e., support for topological compression I/O in the cinema filters (multiple tables are already supported). The only change I made is that the SQL tables are always referred to as InputTableX with X = 0...5. In @pierre-guillou code if the current query only has one table he skipped the X. To skip special cases I would argue for enforcing the X (any thoughts?).

Note: after the merge four state files of ttk-data will no longer work until you merged PR https://github.com/topology-tool-kit/ttk-data/pull/21. 

Best
Jonas

julien-tierny:
hi jonas,

alright I'll wait for @pierre-guillou's feedback before merging these 2 PRs.

cheers
julien
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

pierre-guillou:
Hi Jonas, Julien,

I checked out and tested the code of this PR on my ParaView states. Some features were missing from my #282, #287 and #288 PRs but I ported them and will PR Jonas' branch in a short while.

@JonasLukasczyk The backward-compatibility code in CinemaQuery consisted in adding a trailing '0' to queries that contained 'InputTable' without numeral. This way I avoided updating the ttk-data states that use CinemaQuery. Since you have to/have already updated them anyway to make them compatible with your new filters, I don't think this particular piece of my code is relevant any more.

Pierre

JonasLukasczyk:
Thank you Pierre!

I merged the changes into my branch, but I think before we merge this PR into ttk-dev, we have to revisit the multiple table thing for the CinemaQuery filter:
I also like to have multiple connections on a single port in contrast to one connection at multiple ports, but this way the user can only control the index of the table by making the input selection in the right order, correct? So lets say I have three tables as my input, I have to select them in the right order to guarantee which table will be InputTable{0,1,2,...}, right? Is it also possible to later change the input tables individually by clicking "change input", or would the user have to select all tables again? 

Best
Jonas

pierre-guillou:
Hi Jonas,

Your are right: with this multiple connections approach, the order of the selections is important. You can also change the existing connections afterwards using Change Input and here too the input filters selection order matters. The only drawback is that this order is not explicitly displayed (no numbers on GUI after selection).

However, after a round of testing about modifying the inputs of CinemaQuery, I noticed that I was missing a `clean_command="RemoveAllInputs"` statement inside the `InputProperty` XML tag:

`<InputProperty clean_command="RemoveAllInputs" name="InputTable" command="AddInputConnection" multiple_input="1">`

Without this statement, you can select several inputs with ctrl-click but, if you use Change Input afterwards, the new inputs will be added but the old ones won't be removed, even if they are unselected.

Do you want me to PR you this change?

Pierre

JonasLukasczyk:
Hi Pierre,
that sounds good and I think the multiple connection approach is not perfect because of the invisible order, but far better than the multiple port approach since there is no table limit and better user interaction. So please open a pull request for your fix, and I will also update the BlockAggregator to work in the same way.

Thanks!
Jonas

pierre-guillou:
Hi Jonas,

I just PR'd you the fix.

To get this straight, I have been also looking at the "Group Dataset" ParaView filter interface, which uses the same scheme of one port/multiple inputs.

Hope this helps,

Pierre

julien-tierny:
alright, then it looks like this one is ready to go.
thanks to the 2 of you!
cheers
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

JonasLukasczyk:
In contrast to OpenMP modules that use MPI will not work without it since there is no fallback. So instead of guarding every expression, I would put one guard around the source files. What do you think?

PR #308, opened 2020-01-15
WIP: Unlock OpenMP on windows

CharlesGueunet:
Dear Julien,

This work in progess, once finished, will unleash the full power of OpenMP.... 2.0 on windows.
Ok, this is not so impressive now that we are reaching the 5.0 standard but hey... this is better than nothing :)
Still need some tests on Azure which seems to be dead on my side (only the interface).

Charles

### DIFF
--- a/CMake/BaseCode.cmake
+++ b/CMake/BaseCode.cmake
@@ -188,9 +188,16 @@ function(ttk_set_compile_options library)
     target_compile_options(${library} PUBLIC ${OpenMP_CXX_FLAGS})
     target_link_libraries(${library} PUBLIC ${OpenMP_CXX_LIBRARIES})
 
-    if (TTK_ENABLE_OMP_PRIORITY)
-      target_compile_definitions(${library} PUBLIC TTK_ENABLE_OMP_PRIORITY)
+    target_compile_definitions(${library} PUBLIC TTK_OPENMP_VERSION_MAJOR=${TTK_OPENMP_VERSION_MAJOR})
+    target_compile_definitions(${library} PUBLIC TTK_OPENMP_VERSION_MINOR=${TTK_OPENMP_VERSION_MINOR})
+    # Most important features are explicit
+    if (TTK_OPENMP_VERSION_MAJOR GREATER_EQUAL 3)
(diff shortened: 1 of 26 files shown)


CharlesGueunet:
Will reopen later.

pierre-guillou:
I think there is an alternative way to use OpenMP on Windows that does not rely on such heavy code modifications.

I have been trying for the past week to use [clang-cl](https://clang.llvm.org/docs/MSVCCompatibility.html), a modified version of clang to serve as a front-end to Microsoft's `cl.exe` compiler, and so far it has been able to build TTK with `TTK_ENABLE_OPENMP=ON` and execute the TTK examples, seemingly on several cores, on my Windows VM and on GitHub Actions.

To do so, you'll need the `C++ Clang tools for Windows` optional component of Visual Studio, as stated in [this Visual Studio documentation](https://docs.microsoft.com/en-us/cpp/build/clang-support-msbuild?view=vs-2019), and configure & build TTK with:
```
cmake -DTTK_ENABLE_OPENMP=TRUE -G"Visual Studio 16 2019" -Tclangcl [project/dir]
cmake --build . --config Release --parallel
```
You might need to set the path to the directory containing the OpenMP runtime library (`libomp.lib`) ([somewhere around]( https://stackoverflow.com/questions/57209524/using-openmp-with-clang-and-cmake-in-visual-studio) `C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/Llvm/lib`.

In my packaging CI pipelines, I use [conda's openmp](https://anaconda.org/conda-forge/openmp) libraries though.

julien-tierny:
does paraview also need to be built with clang-cl?
I'm asking the question to see if we need to edit ttk-paraview's packaging too?

pierre-guillou:
No, I was using the standard Visual Studio build of ParaView. It seems that ParaView is not using OpenMP > 2.0 in its source code (or it may disable it on Windows like we do).

JonasLukasczyk:
Does this PR still have a future?

julien-tierny:
nope.
actually, openmp is enabled on windows builds thanks to clang :)

PR #312, opened 2020-01-20
Use protected inheritance for base layer to fix undefined symbols in Python wrapper libraries

pierre-guillou:
This PR switches all multiple inheritance to the base layer in the VTK layer from public to protected.

As a consequence, undefined symbols such as `PyvtkDataSetAlgorithm_ClassNew` should no more appear when executing Python scripts like examples/python/example.py.

Still, some undefined symbols (`PyvtkImageData_ClassNew`, `PyvtkPolyData_ClassNew` and `PyvtkUnstructuredGrid_ClassNew`) remain in ttkTriangulationAlgorithmPython.so. I don't know how to avoid this though. Hopefully, this library is not used through the Python wrapper?

I also noticed that the Python modules under the topologytoolkit package are not re-exported at the package level, meaning you have to use these kinds of imports

```py
from topologytoolkit.ttkMorseSmaleComplex import *
from topologytoolkit.ttkPersistenceCurve import *
from topologytoolkit.ttkPersistenceDiagram import *
from topologytoolkit.ttkTopologicalSimplification import *
```
instead of 

```py
from topologytoolkit import (
    ttkMorseSmaleComplex,
    ttkPersistenceCurve,
    ttkPersistenceDiagram,
    ttkTopologicalSimplification,
)
```
in examples/python/example.py.

No changes to ttk-data states have been noticed so far.

Pierre

### DIFF
--- a/core/vtk/CMakeLists.txt
+++ b/core/vtk/CMakeLists.txt
@@ -77,9 +77,15 @@ if(VTK_WRAP_PYTHON)
   endforeach()
   file(GENERATE
     OUTPUT
-    "${CMAKE_INSTALL_PREFIX}/${TTK_PYTHON_MODULE_DIR}/topologytoolkit/__init__.py"
+    "${CMAKE_BINARY_DIR}/${TTK_PYTHON_MODULE_DIR}/topologytoolkit/__init__.py"
     CONTENT
       "${InitContent}"
+    )
+  install(
+    FILES
+    "${CMAKE_BINARY_DIR}/${TTK_PYTHON_MODULE_DIR}/topologytoolkit/__init__.py"
+    DESTINATION
+    "${TTK_PYTHON_MODULE_DIR}/topologytoolkit/"
     )
   # Install TTK Python
   export(
--- a/core/vtk/CMakeLists.txt
+++ b/core/vtk/CMakeLists.txt
@@ -68,6 +68,20 @@ if(VTK_WRAP_PYTHON)
     TARGET
       TTK::Python
     )
+  # Generate __init__.py
+  set(InitContent "from __future__ import absolute_import\n")
+  foreach(MODULE ${TTK_ENABLED_MODULES})
+    if (NOT "${MODULE}" STREQUAL "ttkTriangulationAlgorithm")
+      string(APPEND InitContent "from .${MODULE} import *\n")
+    endif()
+  endforeach()
+  file(GENERATE
+    OUTPUT
+    "${CMAKE_INSTALL_PREFIX}/${TTK_PYTHON_MODULE_DIR}/topologytoolkit/__init__.py"
+    CONTENT
+      "${InitContent}"
+    )
+  # Install TTK Python
   export(
     EXPORT
       TTKPython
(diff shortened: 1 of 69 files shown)


pierre-guillou:
Hi @CharlesGueunet,

This PR should fix (kind of…) our undefined definitions errors when using the Python wrapper.

I have noticed a change in the Python imports though (see the updated example.py, last commit of the PR). 

When I try to instantiate ttkPersistenceCurve in example.py, I am greeted (on my Arch/Python 3.8 machine) with a 

> Traceback (most recent call last):
>   File "example.py", line 48, in `<module> `
>      curve = ttkPersistenceCurve()
> TypeError: 'module' object is not callable

I have to import the modules inside topologytoolkit as shown above to get access to the class.

I think we miss a (generated?) script that re-exports the Python sub-modules of the topologytoolkit package at the package level. There is a .py script and a .so for every module but no `__init__.py` to initialize the package.

Do you agree? Is there something missing?

Cheers,
Pierre

CharlesGueunet:
> I think we miss a (generated?) script that re-exports the Python sub-modules of the topologytoolkit package at the package level. There is a .py script and a .so for every module but no __init__.py to initialize the package.

Agreed, with a `__init__.py` manually written I am able to use the old import system again.
I am working on a way to generate this file automatically.

CharlesGueunet:
I have a fix for the missing symbols (`PyvtkImageData_ClassNew`, `PyvtkPolyData_ClassNew` and `PyvtkUnstructuredGrid_ClassNew`)
and I have added a way to generate the `__init__.py`. I wil PR this on your branch when I have finished testing.

CharlesGueunet:
You can cherry-pick [this commit](https://github.com/CharlesGueunet/ttk/commit/61cfed213ee2814e162660989311a49177c71241) from my repo to fix the problem with the import in python.
Unfortunately, the fix for the missing symbols in ttkTriangulationAlgorithm is not working, I will need to find another workaround. This should not be a problem for now as this class is not meant to be instantiated directly in python.

julien-tierny:
awesome team work you guys!
I'll wait for azure to be finished prior to merging.
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

pierre-guillou:
Thanks @CharlesGueunet for working on this issue!

However, when I try to re-configure TTK, CMake complains that it can't write the `__init__.py` file in my `${CMAKE_INSTALL_PREFIX}`, which is `/usr/local`.

Is there a way not to require super-user rights during a default configuration?

Pierre

CharlesGueunet:
> However, when I try to re-configure TTK, CMake complains that it can't write the __init__.py file in my ${CMAKE_INSTALL_PREFIX}, which is /usr/local.

The error is during the generate step, because it cannot generate this file correctly in the install folder.
I will improve this mechanism to first generate in the build folder then install it.

CharlesGueunet:
[this commit](https://github.com/CharlesGueunet/ttk/commit/4b0b37d92a1b94fedd7de5fcf1b606b9bb5f8b6f) split the creation of this file in two step, it is first generated in the build folder then installed with the remaining of TTK. No more error :)

Note: the indentation is not really good on my commit :/

pierre-guillou:
I cherry-pick your last commit, rebased everything on top of origin/dev and made some quick tests. Everything seems fine to me.

julien-tierny:
alright guys, looks good!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #313, opened 2020-01-20
Update PersistenceDiagramClustering for VESTEC

pierre-guillou:
This PR updates Jules' work on clustering persistence diagrams. 

Highlights:
* switch from multiple inputs to one multiblock dataset input (otherwise the filter cannot be applied after CinemaProductReader)
* export a distance matrix between the persistence diagrams. By default, the distance matrix is not computed until the checkbox `Output a Distance Matrix`
* transfer FieldData from PersistenceDiagram inputs
* remove the template functions in the VTK layer, since the template parameter was coming from the `Persistence` vtkDoubleArray which has a fixed type (10x faster build time).
* better interfacing with the (old) debug system
* improve code readability with `const`
* avoid segfaults and dangling pointers with references/additional checks
* fix Clang warnings in this filter
* remove some dynamic allocations

A corresponding PR on ttk-data should make the relevant states compatible with the change in input types.

Note that the new input system will break the standalone application since I have not tried to make ttkProgramBase support multiblock datasets alongside standard vtkDataSets.

Enjoy,
Pierre


julien-tierny:
awesome pierre!
thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

Issue #316, opened 2020-01-23
Incorrect PYTHON_SITE_PACKAGES_SUFFIX in core/vtk/CMakeLists.txt


julien-tierny:
on my system (gentoo linux), the cmake variable PYTHON_SITE_PACKAGES_SUFFIX is set to "lib64/python/site-packages" while paraview (and the vtk is ships with) uses "/usr/local/lib/python3.6/site-packages/".

this forces users to include both directories in the env variable PYTHONPATH, which is not necessary (confusing) and incompatible with previous releases of ttk.

@CharlesGueunet 
thanks for fixing that :)

@pierre-guillou may want to have a look as well

cheers,

julien-tierny:
fixed with PR #317.
thanks!

PR #317, opened 2020-01-23
fix(Python): detect real python path used by VTK

CharlesGueunet:
Dear Julien,

This PR address #316, the cmake now detect the real python path used by VTK instead of the one returned by `vtk_module_python_default_destination`.

Charles


julien-tierny:
awesome, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #319, opened 2020-01-23
Overhauled EigenField and HarmonicField filters

pierre-guillou:
This PR modifies the ttkEigenField and the ttkHarmonicField filters, using the new architecture and APIs developed by Jonas
* the VTK layer inherits from the base layer instead of composition,
* both layers are using the new Debug API,
* the base layer has been transformed towards a stateless class. No more setters!
* triangulation pointers have been replaced by references in the base layer. The VTK layer should be responsible for checking them.
* the base layer has been templated around the triangulation type, but in practice only ttk::Triangulation is used: triangulation accesses are not critical in these filters and I prefer not to increase build times.
* "preprocessTriangulation" -> "preconditionTriangulation"
* more uniform coding style between the filters

No change so far in ttk-data/states/harmonicSkeleton (apart from 3 welcome logos instead of one).

Enjoy,
Pierre


julien-tierny:
looks good to me.
let's investigate the multiple welcome message though.
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #320, opened 2020-01-24
New implementation for the Discrete Gradient (ProcessLowerStars)

pierre-guillou:
This PR contains the updated Discrete Gradient module with a new algorithm called ProcessLowerStars, from the "Theory and Algorithms for Constructing Discrete Morse Complexes from Grayscale Digital Images" but applied on triangulations.

The Morse-Smale Complex filter has been adapted to the use of this new gradient (less post-processing). The dependency to ScalarFieldCriticalPoints has been severed.

The ttk-data states should run fine (and faster) with it.

Enjoy,
Pierre


julien-tierny:
looks good, thanks!
julien
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #322, opened 2020-01-27
DiscreteGradient fixes

pierre-guillou:
This PR brings two small fixes to the new Discrete Gradient implementation:

- the ttkDiscreteGradient filter now does not perform reverse the gradient
- the computation of the ManifoldSize array for the Morse-Smale Complex critical points has been moved into its own method in the .cpp, parallelized and accelerated. After a trivial parallelization of the computation loop, I got an additional 200x speedup on ctBones.vti by pre-caching some parts of the computation. The resulting execution times of this loop is now negligible compared with the whole Morse-Smale Complex computation (before: 250s-2/3 of the whole computation with 16 threads, after: 1.2s). Only drawback:  with small data sets, this loop's computation time tends to be 2x slower: from 0.01 to 0.02s for morseMolecule, with is still negligible.

Enjoy,
Pierre


julien-tierny:
looks good, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #323, opened 2020-01-29
Sanity checks for "ttkTriangulation from vtkUnstructuredGrid"

kappe-c:
Hello,
sorry this is going to be a bit lengthy - see the TLDR at the end.
I have run into this issue a few times before so now I'd like to contribute this small "patch".
As is more or less well known, TTK does only work on triangulations (which is a reasonable constraint I agree). However, as far as I can tell this is never enforced: there is no code to triangulate non-triangular meshes or at least trigger some warning. I have tracked down the problem to 
ttkTriangulation.cpp ttkTriangulation::setInputData line 214
where vtkUnstructuredGrid objects are handled. The documentation for this function does in fact say "Specify the input VTK object representing a triangulation or a regular grid" but this is a function which no (module) developper actually calls. At least not explicitly - only via the TTK_SETUP macro in ttkWrapper.h line 186 (I think).
So in practice (for me with Paraview) it can easily happen that a TTK algorithm is started with an unsupported input mesh, with the result that the application crashes or produces arbitrary results. And the user doesn't get any helpful message, not even a programmer with a debug build. (It took me quite some time to identify the problem and where it occurs.)
My code is just a minimal solution that makes the necessary checks in debug mode. Because of what I said above it might actually be good to have the code run also in release mode; I only hesitate to remove the "debug guards" because the traversal of all cells as in `grid->IsHomogeneous()` may be a little costly.
I also leave relatively verbose comments for the case that anyone ever wants to tackle the issue more thoroughly. I think only being able to deal with already triangulated meshes does reduce the applicability of TTK quite a bit, which is especially sad since things like a set of rectangles (or actually any convex 2D cells) could be triangulated quite easily.

I attach one data and two minimal state files to "proof my point". The dataset is just an artificial scalar field on a rectilinear grid. If output type Unstructured is selected in Paraview (for whatever reason), a TTK algorithm (e.g. PersistenceDiagram) will treat the mesh as tetrahedra. In the example, file the result (I set "Embed in domain") happens to look quite reasonable but their may be crashes or undefined behavior later on.
In the second state file, I apply a "Clip" filter (literally "Common" in the menu). This seems to always output an unstructured grid, however not a triangulation. So currently a pipeline that may work well with TTK, suddenly fails after introducing a simple clip filter.
(The example happens to highlight a "bug" in Paraview: the corners of the clipped box are cut away leading to a mix of quads and four triangles.)

[unstructured problems.zip](https://github.com/topology-tool-kit/ttk/files/4127285/unstructured.problems.zip)

Concerning the concrete code: I also introduce the variable `grid` in the scope to make the existing code a little nicer. This explains the diffs outside of the "#ifndef NDEBUG" block.

TLDR: I would like to add some code to check if a vtkUnstructuredGrid is actually a triangulation that can be used by TTK. This can at least make debugging easier.


kappe-c:
Hm ... I did not expect any trouble with CI for this one. Seems to be a problem of the old Ubuntu 14.04 (trusty). I'll see what I can do about this.

kappe-c:
I'm sorry, I have no idea where that error comes from. The C++ compiles fine from what I see. It seems to be some problem on binary (assembly) level, maybe in relation with the old binutils shipped with Ubuntu 14.04. Because I cannot really see any suspicious line of code, I can only debug this by randomly editing anything, then push and see what Travis says. After a few trials I have no more motivation to do this. Since not even the publishers of the OS support it anymore, I don't really see, why the TTK devs should (support for Trusty ended in April 2019). But of course that's not for me to decide; I know that many (research) institutions are quite conservative when it comes to operating systems.
For completeness' sake I am going to add a commit here with an Ubuntu 16.04 Travis config (still rather old imho ...) that passes the tests, even though I know that such a substantial change in the build/test environment will not be undertaken lightly.

julien-tierny:
hi christopher,

thanks a lot for your contribution to ttk!

I have several remarks here:

- we changed a bit the structure of the repo lately. PRs should now be prepared for the **dev** branch. I have updated the CONTRIBUTING.md accordingly earlier today.

- you can ignore travis-ci issues as we switched the ci to azure on the dev branch.

- regarding your remark about the consistency check, you are absolutely right. if explicit meshes are given on the input, we expect them to be simplicial meshes. in practice, what we recommend is that users always call the tetrahedralizer filter first on their explicit meshes prior to using ttk (to make sure they are simplicial). there's no need to call tetrahedralizer if the input is a vtkImageData (these will be implicitly triangulated by ttk). but I agree with you, there's no explicit message for the user and this could be improved. from there, I'd have several suggestions:

  - the test IsHomogenous() seems too restrictive (determine if all cells are of the same type). in principle, simplicial complexes can contain simplices of arbitrary dimensions (heterogeneous cell types in the VTK terminology). we want to authorize VTK_LINE, VTK_TRIANGLE, VTK_TETRA. you may want to implement your own function that tests for that.
  - you want to throw an explicit error message with recommendations. something like "Input explicit mesh is not simplicial. Please call the tetrahedralizer module first."
  - such a test also needs to happen for the vtkPolyData processing further down ttkTriangulation.cpp
  - you don't need to restrict that test to the debug mode, as the function setInputData() should be called only the first time a vtkPolyData or vtkUnstructuredGrid passes through a ttk module (see ttkTriangulation::deepCopy() and ttkTriangulation::shallowCopy(), which protect the call to setInputData() with the isEmpty() check). so in principle, setInputData() should not be called at the execution of each ttk filter, but only once, and only for the top one of a pipeline.
  - travis.yml fixes are no longer necessary.

I hope that helps. 
that'd be awesome if you could update your pr accordingly :)

thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

kappe-c:
hi julien,
thanks for the reply.

- ok, i'm in sync with the dev branch now
- of course that solves my travis issues
- ok, i already imagined that sth. like the tetrahedralizer can be used as work-around;
  glad my code to output a warning in problematic cases is generally welcome

I was aware that, in principle, simplicial complexes may include cells of various dimensions. Nevertheless i wrote the code the way i did because during debugging i found that in several places it is assumed that the cell array of a ttkTriangulation is homogeneous.
The most fundamental place is probably ExplicitTriangulation.h `getDimensionality` which returns `cellArray_[0] - 1`. To be precise, this doesn't require the simplexes to be homogeneous but only to be sorted descendingly by num. vertices per cell but i don't think this is actually done (and maintained through the pipeline).
Another prominent place is the `buildCellEdges` method in e.g. the class `ThreeSkeleton`. There, a cell array is iterated over using `int vertexPerCell = cellArray[0];`.
This is maybe stuff for a separate discussion but interestingly the doc for the parameter `cellNumber` says "Number of maximum-dimensional cells in the triangulation". This sounds to me like the `cellArray` parameter should be already filtered to contain only simplexes of maximum dimension (so that is has size `cellNumber`), which would make it homogeneous. Again, I don't think such filtering is done; only the cell array from the ttkTriangulation is passed over. This cell iteration in `buildCellEdges` is one concrete example where basically random memory is read if `vertexPerCell` is not constant in the given data, which leads to the observed crashes.
Someone please correct me if i'm overlooking something here. I don't mean to be complaining just out of spite.

- I will incorporate the other remarks as soon as i find time for that, probably until monday evening

julien-tierny:
hi christopher,

thanks a lot for your commit and feedback.

you're absolutely right, in its current form the code assumes homogeneous simplicial complexes but, you're right as well, this is something we plan on fixing soon (hopefully this year).

your commits look good to me (thanks for putting the specialized checker together). I will probably adjust the content of the messages a little bit with less negative terms.

I'll wait for the azure ci to be fixed prior to merging your PR.

thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
hi christopher,

can you please sync your dev-chris branch with the dev branch of the master ttk repo? (this will fix the continuous integration)

thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

kappe-c:
Alright, I have done the merge and now the continuous integration tests are passed.
Of course feel free to edit as you see fit.

julien-tierny:
hi christopher,

fyi, @CharlesGueunet is currently working on another PR (#328) that will conflict with yours (ttkTriangulation.cpp).

if you don't mind, I'd like to give him priority as his PR covers aspects that are fundamental to the interface between VTK and TTK, and which need to be addressed asap (and before the release of paraview 5.8 for sure).

charles needs to check one last bit before he switches his PR from the "WIP" status to the "ready" status. once it is merged, you'll have to synchronize your branch again. this will involve some conflict solving but hopefully, that should be fairly modest.

I hope you will understand.

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
hi christopher,

the PR #328 has now been merged with the dev branch.

could you please synchronize again your dev-chris branch with the ttk:dev branch and work out the resulting conflicts?

thanks a lot for your help and contribution.

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

kappe-c:
Strangely I haven't got this conflict locally. Come to think of it, maybe I forgot a `git fetch upstream` ...  I am going to look into this tomorrow morning.

kappe-c:
Good morning,
I hope the merge is fine now. (I like the changes being made.)

I am aware that there are still two NOTEs of mine regarding "special" cases of a vtkPolyData.
Because they have no practical relevance for me (and my time is also limited ;-)) I leave this for the future.

Having reviewed the code for the merge, I have two remarks:

1) In the following place @CharlesGueunet seems to have made a little mistake.
Lines 2 and 4 are exactly the same; in line 4 i suppose it should be `GetLines` instead of `GetPolys`.
```
      auto polysData = static_cast<LongSimplexId *>(
        ttkUtils::GetVoidPointer(vtpDataSet->GetPolys()->GetData(), 0));
      auto linesData = static_cast<LongSimplexId *>(
        ttkUtils::GetVoidPointer(vtpDataSet->GetPolys()->GetData(), 0));
```

2) Concerning `if(vtpDataSet->GetPolys()) {`.
In my vtk-8.2 documentation it says about
`vtkCellArray* vtkPolyData::GetPolys ( )`
"Get the cell array defining polygons.
If there are no polygons, an empty array will be returned (convenience to simplify traversal)."

So I think, `GetPolys` never returns a null pointer.
That way they don't have to -- but we cannot -- check for null pointers. One can check the number of elements instead.

If I may, I could address the two issues in small follow-up commits.

julien-tierny:
hi christopher,

> I hope the merge is fine now. (I like the changes being made.)
looks like it.

> I leave this for the future.
looks good.

> Lines 2 and 4 are exactly the same; in line 4 i suppose it should be `GetLines` instead of `GetPolys`.
good catch.

> 2) Concerning `if(vtpDataSet->GetPolys()) {`.
you may be right. you may want to double check with the nighly version of vtk though. in anyways it does not hurt as it is right now (this is fairly high level code that is not run intensively, so no performance impact).

> If I may, I could address the two issues in small follow-up commits.
please do.

@CharlesGueunet
can you confirm you're ok with these two suggestions?

thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

CharlesGueunet:
>  Lines 2 and 4 are exactly the same; in line 4 i suppose it should be `GetLines` instead of `GetPolys`.

That is right, my mistake.

>  2) Concerning `if(vtpDataSet->GetPolys()) {`.

Old versions of TTK used `vtpDataSet->GetPolys()->GetPointer()` that we want to avoid. Using something like the number of cell would be more appropriate indeed :)

kappe-c:
Regarding GetLines()/GetPolys I found the same documentation for the nightly vtk, so i did in fact edit as propsed.

julien-tierny:
alright, looks good.
thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

kappe-c:
My pleasure :-)
Thank you for the all the feedback and guidance!

PR #329, opened 2020-02-07
Fix infinite loop in DiscreteGradient when returning saddle connectors

pierre-guillou:
The current PR reverts is a new take on the missing separatrices issue in  the morseMolecule state file. The previous solution is causing infinite loops when returning saddle connectors for some data sets.

It's the removal of  the `nconnections` early returns in `getAscendingPathThroughWall` and `getDescendingPathThroughWall` to fix the missing dependencies in morseMolecule that has been causing those infinite loops.

The current solution is to keep the `nconnections` early returns and to introduce a new boolean parameter, by default set to false, to activate those early returns, especially when computing saddle connectors.

No change to the morseMolecule state file so far.

As a bonus, I made the `getDescendingPathThroughWall` method return bool instead of int, to make it more similar to the other `getAscendingPathThroughWall` method.

Pierre

### DIFF
--- a/core/base/discreteGradient/DiscreteGradient.cpp
+++ b/core/base/discreteGradient/DiscreteGradient.cpp
@@ -465,6 +465,7 @@ int DiscreteGradient::getDescendingPathThroughWall(
   const Cell &saddle1,
   const vector<wallId_t> &isVisited,
   vector<Cell> *const vpath,
+  const bool stopIfMultiConnected,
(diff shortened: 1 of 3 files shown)


julien-tierny:
alright, looks good.
thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #332, opened 2020-02-09
debug: added a space before line filler; ttkAlgorithm: automatic memo…

JonasLukasczyk:
debug: added a space before line filler

ttkAlgorithm: automatic memory managment by removing 'new' and observing owner destruction

### DIFF
--- a/core/base/common/Debug.h
+++ b/core/base/common/Debug.h
@@ -418,6 +418,9 @@ namespace ttk {
       std::string combinedMsg = msg;
 
       if(filler.length() > 0) {
+        if(msg.length() > 0)
+          combinedMsg += " ";
+
         int gapWidth = debug::LINEWIDTH - this->debugMsgPrefix_.length()
                        - combinedMsg.length() - right.length();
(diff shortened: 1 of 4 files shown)


JonasLukasczyk:
I incorporated your edits Julien, so now the registry uses the ScalarPointer as a unique key to identify a vtkImageData object (vtkCellArray is used for vtkUnstructuredGrid and vtkPolyData objects). Before it was always necessary that the owner of a triangulation was a vtkObject, so I'm not super happy with the ScalarPointer, but I know that there is currently no other alternative instead of just recreating the implicit triangulation every time.

Best
Jonas

julien-tierny:
alright, thanks for the quick fix!

the memory handling in ttkAlgorithm looks a little over-engineered but that will do.

in the future: can you make separate PRs in such configurations? (1 per topic) thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #333, opened 2020-02-10
Refactor Triangulation method signatures

pierre-guillou:
This PR refactors all uses of the following pattern:
```cpp
#ifdef TTK_ENABLE_KAMIKAZE
int function(int arg) {
#else
int functionInternal(int arg) {
#endif
```
into 
```cpp
int TTK_TRIANGULATION_INTERNAL(function)(int arg) {
```

through all triangulation variants. Here, `TTK_TRIANGULATION_INTERNAL` is a macro that concatenates the function name with "Internal" if the KAMIKAZE flag is not defined.

In the end, there should be only one `#ifdef TTK_ENABLE_KAMIKAZE` throughout the entire TTK codebase.

In addition, I took the liberty to fix some Clang warnings about missing qualifiers and types.

Enjoy!
Pierre

### DIFF
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -19,6 +19,12 @@
 #include <array>
 #include <ostream>
 
+#ifdef TTK_ENABLE_KAMIKAZE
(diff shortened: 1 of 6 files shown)


julien-tierny:
ok.

PR #335, opened 2020-02-27
Use NewInstance to replace several New in Blank

CharlesGueunet:
Dear Julien,

This is a small PR to improve a bit the code in the Blank module.
The switch case seems to be there to filter some vtkType not supported by VTK:
* VTK_BIT
* VTK_SIGNED_CHAR
* VTK_UNSIGNED_CHAR
* VTK_UNSIGNED_INT
* VTK_LONG
* VTK_UNSIGNED_LONG

Is this list still up to date with the current TTK ?

Charles

### DIFF
--- a/core/vtk/ttkBlank/ttkBlank.cpp
+++ b/core/vtk/ttkBlank/ttkBlank.cpp
@@ -46,39 +46,16 @@ vtkStandardNewMacro(ttkBlank)
   // allocate the memory for the output scalar field
   if(!outputScalarField_) {
     switch(inputScalarField->GetDataType()) {
-
       case VTK_CHAR:
-        outputScalarField_ = vtkCharArray::New();
-        break;
-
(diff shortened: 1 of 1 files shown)


julien-tierny:
hi charles,

thanks a lot.

> This is a small PR to improve a bit the code in the Blank module.
thanks, but as a matter of fact, since the Blank module served as a base example for a long time, many modules suffer from the same issue.
(this will need to be addressed at the spring hackathon)

> Is this list still up to date with the current TTK ?
well, from now on, the base example will be the HelloWorld module, which does not suffer from this issue (see core/vtk/ttkHelloWorld/ttkHelloWorld.cpp:144)

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #336, opened 2020-03-03
DiscreteGradient : Sort vertices prior to computation

pierre-guillou:
This PR improves the performance of the discrete gradient filter by sorting the vertices prior to the main computation. The observed gain varies between 5% and 15%.

This transformation also allows to un-template DiscreteGradient methods that were relying on the scalars and offsets fields to compare vertices. Those methods have been moved into the DiscreteGradient.cpp file. As a consequence, build times should also be reduced.

Enjoy,
Pierre


julien-tierny:
looks good, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #339, opened 2020-03-10
Fix C3848 compile error with VS2019 (Version 16.4.0)

PetiteViking:
Compiling TTK on Visual Studio 2019 (v 16.4.0), I got
image

Const-qualifying the `operator()` for `filtrationCtCmp` fixes the issue.


PetiteViking:
Resubmitted in #340

PR #341, opened 2020-03-11
Conflict with FTMAtomicVector.h and FTRAtomicVector.h

dbanesh:
Both core/base/ftmTree/FTMAtomicVector.h and core/base/ftrGraph/FTRAtomicVector.h use the following definition:

#ifndef ATOMICVECTOR_H
#define ATOMICVECTOR_H

Therefore, it seems that when both files are included, it sometimes causes the compiler to use the incorrect version. This was made apparent when compiling the examples/c++ sample code with the addition of the "#include <FTRGraph.h>" line. In this particular example, it was looking into the FTMAtomicVector.h file for a const definition of end() while it should be looking into the FTRAtomicVector.h file. This causes the compilation to break.

Updating the FTMAtomicVector.h file to include the member definition allows the example code to compile and run properly, however this is a partial fix. The ifndef's should probably be updated to include unique identifiers and tested to make sure it doesn't break anything else in the hierarchy.

PR #343, opened 2020-03-23
rename AtomicVector to the persepctive instance for ftmTree or ftmGraph

ollielo:
This is an amendment and replacement of #341 and #342. It renames AtomicVector to either FTMAtomicVector or FTRAtomicVector depend on which instance it is. Also update the include guard.

The deeper question for the core TTK developers is why do we have two essentially the same version of AtomicVector and AtomicUnionFind? Isn't **Do Not Repeat Yourself (DRY)** the first principle of good software engineering?


julien-tierny:
hi,

sorry for the late reply but the situation is a little hectic here.

thanks a lot for your contribution to TTK!

> This is an amendment and replacement of #341 and #342.
@dbanesh 
I'm going to go ahead a close #342 then. 
thanks for confirming you're ok with this.

> It renames AtomicVector to either FTMAtomicVector or FTRAtomicVector depend on which instance it is.
thanks, looks good to me.
@CharlesGueunet: can you double-check this is OK for you too?

> why do we have two essentially the same version of AtomicVector and AtomicUnionFind? 
looks like some mutualization could happen here indeed.
@CharlesGueunet: any thought on that?

regarding commmit be34c3411b1151a494a24ea6b7f6e1ab9d41df51:
the purpose of the "examples" directory is to show how to put together a simple ttk pipeline (always the same one, taken from the companion paper) for the different APIs supported by ttk. 
so it short, it shows how to integrate ttk in an external program.

its purpose, as it currently stands, is not to exemplify all the features of ttk. ttk-data/states/ (https://github.com/topology-tool-kit/ttk-data/tree/dev/states) serves that purpose.
so I would recommend to remove that commit from the PR.

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

CharlesGueunet:
> can you double-check this is OK for you too?

This PR is all right for me too.

> looks like some mutualization could happen here indeed.
any thought on that?

Agreed, this would be helpful especially when bug are fixed (like the one on the const iterator) and would make this container more available for every developer.

Not related to this MR, the save and load function used here to read and write OFF could also be mutualized for every TTK module to use. This way, even the TTK OFF Reader / Writer could use them.

ollielo:
@julien-tierny 

Thanks a lot for the feedback.

I agree that the added example is not necessary for the MR. I put it in as a way to communicate with Divya on a working example/test.

ollielo:
@julien-tierny 

We also found some mysterious seg faults when FTRAtomicVector is destroyed, especially when T = SuperArc (which contains its own std::vector via SuperArc). When the Vector is destroyed, the destructor of std::vector is called on invalid instances o std::vector and causes seg fault. I am stilling trying to figure out why there are those invalid instances but Is suspect the problem comes from reserve() v.s. resize() (and how AtomicVector override, confuses them).

dbanesh:
@julien-tierny confirming to please close out #341 and #342 given this PR.

julien-tierny:
> I agree that the added example is not necessary for the MR. I put it in as a way to communicate with Divya on a working example/test.
@ollielo
can you please remove commit be34c34 from the PR?
I'll go ahead and merge after that.

thanks
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

ollielo:
@julien-tierny 

Changes to examples were backed out. Please review.

julien-tierny:
alright, looks good to me.

thanks a lot for your contribution to TTK.
may that be the first of many :)

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #344, opened 2020-04-06
ParaView 5.8 & VTK 9

CharlesGueunet:
Dear Julien,

With this PR, TTK becomes VTK 9 and ParaView 5.8 compliant !

* [x] States tested with PV 5.8
* [x] States tested with PV 5.7

This PR implements the CellArray version for VTK9/ParaView 5.8.
The new CellArray is used in replacement of the old one when ParaView 5.8 is detected or when VTK 9 is detected or when neither of them is used and the user set the CMake variable `TTK_CELL_ARRAY_LAYOUT` to "OffsetAndConnectivity".

On this PR, modules relying on CellArray have been updated to fill two array as in the new layout. Additionally, a translator method has been added to transform them in a single array for elder version of VTK (<9) and ParaView (<5.8).

Bonus:
 * Some CMake improvements to avoid warning with cmake 3.17
 * Some optional libraries are now only linked for modules that requires them (eigen, spectra and boost)
 * Base code can now declare `OPTIONAL_DEPENDENCY`
 * Update ParaView patches for 5.8
 * Use VTK return code in wrappers to interrupt pipeline in case of error

Note, icoSphereFromPoint may still be improved in a futur PR.

Best,
Charles


julien-tierny:
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

CMakeLists.txt
```diff
@@ -31,22 +31,17 @@ endif()
...
 
 # Base code
 # ---------
 
 include(CMake/CompilerFlags.cmake)
 include(CMake/BaseCode.cmake)
-add_subdirectory(core)
-
 add_library(baseAll INTERFACE)
-add_library(ttkAll INTERFACE)
```
julien-tierny, line 49: wasn't that necessary for the vtk bundle?

vollia: I tried the examples that use TTK as third party and they works well. The TTKVTKConfig.cmake continue to define the same target as before so this ttkAll should not be required

julien-tierny: you mean the examples in examples/c++/ and examples/vtk-c++?

core/base/auction/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 ttk_add_base_library(auction
 
-  SOURCES Auction.cpp 
+  SOURCES Auction.cpp
   HEADERS Auction.h AuctionImpl.h AuctionActor.h
-	LINK triangulation persistenceDiagram kdTree)
```
julien-tierny, line 5: have you changed all LINK keywords by DEPENDS keywords systematically?

core/base/eigenField/CMakeLists.txt
```diff
@@ -3,23 +3,16 @@ ttk_add_base_library(eigenField
...
-  find_path(SPECTRA_INCLUDE_DIR Spectra/SymEigsSolver.h)
-  if(SPECTRA_INCLUDE_DIR STREQUAL "SPECTRA_INCLUDE_DIR-NOTFOUND")
-    option(TTK_ENABLE_SPECTRA "Enable Spectra support" OFF)
-    message(STATUS "Spectra not found, disabling Spectra support in TTK.")
-  else()
-    option(TTK_ENABLE_SPECTRA "Enable Spectra support" ON)
-  endif()
-else()
-  option(TTK_ENABLE_SPECTRA "Enable Spectra support" OFF)
-  message(STATUS "Spectra not found, disabling Spectra support in TTK.")
```
julien-tierny, line 22: I'm confused here. where did all that code block go?

core/base/explicitTriangulation/ExplicitTriangulation.h
```diff
@@ -1004,20 +1000,47 @@ namespace ttk {
       return 0;
     }
 
+#ifdef CELL_ARRAY_NEW
```
julien-tierny: can you point me to where this variable is defined?
shouldn't we give it a more explicit name (like TTK_CELL_ARRAY_NEW)?

julien-tierny: ok, I found where it was defined (core/base/skeleton/CMakeLists.txt).
however, the explicit name would be better (thanks for pushing that!)

core/base/auction/CMakeLists.txt
vollia: Yes, in order to stay consistent with the wrappers so we do not have to use LINK here and DEPENDS on the vtk wrapper for the same purpose.

core/base/eigenField/CMakeLists.txt
vollia: In config.cmake as we only look for SPECTRA if NUMPY is enabled.
The idea is to link SPECTRA only for module, but to `find` it (so enable/disable it) only once.

julien-tierny: I'm afraid you made a mistake here.
there's no relation between spectra and numpy.
numpy is used in dimensionReduction.
spectra is used in eigenField. spectra should only be activated if eigen3 has been activated.
so it looks like that code block should come back here (since eigenField is the only module using spectra).

vollia: Sorry, my comment is indeed wrong. In `config.cmake` we only look for SPECTRA if EIGEN is enabled (not numpy, sorry).
As you can see in `config.cmake`, SPECTRA is still searched only if EIGEN is found and disbled otherwise. I just wanted to avoid any `find_package` in a base cmake file. I think all dependencies should be resolved once in the `config.cmake`.

CMakeLists.txt
vollia: Exactly

julien-tierny: alright, thanks!

core/base/eigenField/CMakeLists.txt
julien-tierny: gotcha. thanks!

core/base/skeleton/CellArray_legacy.h
```diff
@@ -7,12 +7,20 @@
 ///
 /// %CellArray is a generic container that allows to deal with various cell
 /// layouts in memory \sa Triangulation \sa ttkTriangulation
+/// This version assume the cell data to be given a unique array.
```
julien-tierny: assume*s* the cell data to be given *as* a unique array.

core/base/skeleton/CellArray_legacy.h
```diff
@@ -7,12 +7,20 @@
...
+/// This array contain for each cell the number of points in the cell then the
```
julien-tierny: contain*s*

core/base/skeleton/CellArray_new.h
```diff
@@ -0,0 +1,97 @@
...
+/// %CellArray is a generic container that allows to deal with various cell
+/// layouts in memory \sa Triangulation \sa ttkTriangulation
+/// This version assume the cell data to be given by two
+/// arrays:
+/// * Connectivity, containing the list of point ids of each cells
+/// * Offsets, containing the beginning and end position of each cell in the
+/// Connectivity array See VTK 9 vtkCellArray documentation for more details
+/// about this layout.
+/// This class is more a view on data than a complete container. It does
+/// not support memory reloated operations.
```
julien-tierny: reloated?

core/base/skeleton/ThreeSkeleton.cpp
```diff
@@ -243,15 +243,16 @@ int ThreeSkeleton::buildCellNeighborsFromVertices(
   vector<vector<SimplexId>> &cellNeighbors,
   vector<vector<SimplexId>> *vertexStars) const {
 
-  if(cellArray.getDimension() == 2) {
+  // TODO: ASSUME uniform mesh here!
```
julien-tierny, line 246: yes indeed. thanks for the headsup :)
(useful for future plans)

paraview/patch/patch-paraview-5.8.0.sh
```diff
@@ -0,0 +1,92 @@
...
+if [ -z "$1" ] ||[ ! -d "$1" ]; then
+  echo "Error: Missing ParaView source tree path."
+  echo "Usage:"
+  echo "   ./patch.sh <ParaView source tree path>"
+  exit 2
+fi
+
+PATCH_DIR=$(pwd)
+
+if [ ! -f ${PATCH_DIR}/paraview-5.7.0-CMakeLists.txt.patch ]; then
```
julien-tierny: 5.8.0

julien-tierny:
alright charles,

I've finished reviewing the code for this PR.
for me it's all good. 

could you take care of the following adjustments real quick?
[ x ] explicit variable names (TTK_CELL_ARRAY_NEW, TTK_CELL_ARRAY_LEGACY)
[ x ] typos in core/base/skeleton/CellArray_new.h
[ x ] typos in core/base/skeleton/CellArray_legacy.h
[ x ] typos in paraview/patch/patch-paraview-5.8.0.sh 

(in the meantime I'll go ahead and test the states files over here).

cheers, 
julien
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
I forgot I had permissions on your fork.
I'm on these tiny edits.
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
alright, looks ready to go.
thanks charles & thanks kitware!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #347, opened 2020-05-06
CinemaProductReader: Overwrite FILE column in products FieldData

pierre-guillou:
I came upon the following behavior while using the TTK Cinema filters to extract a subset of a first Cinema Database and writing it in a second one.

Since CinemaProductReader adds only non-existing Field Data Arrays to database products, extracts from the first database will contain a FILE Field Data Array relative to this first database. This Field Data will be written in the second database along the product, preventing another CinemaProductReader on the second database to replace it with a updated and correct path.

This PR aims at fixing this behavior by always overwriting the FILE array.

@JonasLukasczyk Do you think this fix is compatible with the Cinema Database spec?

Pierre

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -238,21 +238,7 @@ jobs:
   steps:
   - script: |
       mkdir build
-      SET "FILENAME=%~dp0\boost.zip"
-      bitsadmin /transfer myDownloadJob /download /priority normal "https://dl.bintray.com/boostorg/release/1.67.0/source/boost_1_67_0.zip" "%FILENAME%"
-      move  "%FILENAME%" boost.zip
-      unzip boost.zip
-    displayName: 'Download and extract Boost'
-
-  - script: |
-      echo "compiler initialization"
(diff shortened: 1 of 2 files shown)


JonasLukasczyk:
Sounds good!

We should also consider always deleting the FILE array in the CinemaWriter. It is just important that the input is just passed through as is, so internally the writer has to make initially a shallow copy, and then removes the FILE field data array. At this point we can also delete other temporary ttk field data arrays such as ttk_IterationInfo. If you are in favor of this I would work on that.

Best
Jonas

pierre-guillou:
Thanks for your answer!

I agree with your suggestion about removing the Field Data generated by TTK Cinema filters before writing products.

Thanks,
Pierre

julien-tierny:
thanks a lot pierre and thanks for the azure fix!
cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #349, opened 2020-05-12
Accelerate integer power computations

pierre-guillou:
std::pow is often used to compute powers throughout the TTK codebase. However, std::pow is optimized for floating-point exponents and can be quite slow with an integer exponent. As an example, std::pow is a bottleneck in the PersistenceDiagramClustering module even though this module computes only squares and square root (with the default wasserstein distance).

This PR introduces two new templated power functions dedicated to integer exponents (in the Geometry module) and replaces the calls to std::pow when applicable. In several modules using similar powers of ten to compute a precision value, those values have been pre-computed and stored into global variables.

As a consequence, time spend in the PersistencDiagramClustering for the two relevant state files in ttk-data has been halved. Other state files using modified functions should perform either the same or a little bit better than before if std::pow was not the main bottleneck of the used filters.

Additionally, the template specialization code in the Geometry module has been refactored using a macro to reduce the number of lines of code (and one function declaration without an implementation has been removed).

Also, I run some profiling tool yesterday evening on the quadrangulation state file and found some minor performance bottlenecks that have been fixed and attached to this PR.

Enjoy,
Pierre

### DIFF
--- a/core/base/bottleneckDistance/BottleneckDistanceMainImpl.h
+++ b/core/base/bottleneckDistance/BottleneckDistanceMainImpl.h
@@ -102,7 +102,7 @@ int BottleneckDistance::computeBottleneck(const std::vector<diagramTuple> &d1,
         const diagramTuple a, const diagramTuple b) -> dataType {
     BNodeType ta1 = std::get<1>(a);
     BNodeType ta2 = std::get<3>(a);
-    double w = wasserstein > 1 ? wasserstein : 1; // L_inf not managed.
+    const int w = wasserstein > 1 ? wasserstein : 1; // L_inf not managed.
 
     // We don't match critical points of different index.
     // This must be ensured before calling the distance function.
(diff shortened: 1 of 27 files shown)


julien-tierny:
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

core/base/bottleneckDistance/BottleneckDistanceMainImpl.h
```diff
@@ -102,7 +102,7 @@ int BottleneckDistance::computeBottleneck(const std::vector<diagramTuple> &d1,
         const diagramTuple a, const diagramTuple b) -> dataType {
     BNodeType ta1 = std::get<1>(a);
     BNodeType ta2 = std::get<3>(a);
-    double w = wasserstein > 1 ? wasserstein : 1; // L_inf not managed.
+    const int w = wasserstein > 1 ? wasserstein : 1; // L_inf not managed.
```
julien-tierny, line 105: is w the exponent used in the expression of the distance?
if so, in principle, this can be a real value (not necessarily an integer).

would that be possible that powInt behaves differently based on the type it is given as an argument?
for instance, if it detects the argument is indeed an integer, use the current powInt implementation.
if it detects it is not, use the regular pow implementation.

pierre-guillou: Here, `w` depends on the value of the `wasserstein` variable, which is an integer defined at line 41 of the same file. In this case, using a double type should be less efficient.

I tried to keep using `std::pow` every time the exponent was clearly a double variable.

In theory, it should be possible to unite the two behaviors using `std::is_integral` and `std::is_floating_point`. I will check that.

julien-tierny:
alright, looks good.
thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #350, opened 2020-05-12
Preconditions for implicit triangulations

pierre-guillou:
This PR adds preconditions to the implicit triangulation module in order to accelerate the triangulation access methods.

Several new methods have been added to the ImplicitTriangulation class:
* preconditionVerticesInternal,
* preconditionTetrahedronInternal,
* preconditionEdgesInternal and
* preconditionTrianglesInternal.
The last two override the AbstractTriangulation corresponding methods and are called automatically by the user modules. The two first however don't have parent functions in AbstractTriangulation. They are called at the end of the setInputGrid function to avoid calling them in the user modules.

These four new methods compute and store in member vectors a "position" (enum variant) and 3D integer coordinates on the grid. These member vectors serve as cache for subsequent triangulation accesses.

Observed speedups range from +10% (DiscreteGradient) to +40% (ScalarFieldSmoother, ScalarFieldCriticalPoints).

No change has been noticed when running the validation state files.

Enjoy,
Pierre


julien-tierny:
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
alright, that looks good to me.
could you port that preconditioning to the periodic grid as well though?
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

pierre-guillou:
I can try to replicate these optimizations to the periodic grid.
Is there a test case in ttk-data that use it ? I will need one to check for performance improvements and that I'm not introducing new regressions.

julien-tierny:
thanks pierre for the quick feedback.

> Is there a test case in ttk-data that use it ?
unfortunately, not.

I'd recommend to write a small program that tests all the involved adjacency queries on a small example and write the result to disk (csv file for instance).

once this is done, you can run that program with the current implementation of periodic grids.

then, you can include this preconditioning and check that the new outputs of the small test program are identical to the first ones.

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
alright, looks good.
let's port the preconditioning to the periodic grids in June.
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

Issue #351, opened 2020-05-12
New ttkAlgorithm API breaks module management scripts


julien-tierny:
The module management scripts (scripts/createTTKmodule.sh) recently switched to the new ttkAlgorithm API for the generation of a new module (based on HelloWorld, commit fdffe9ec7856048b277fdc8f33cafadb7e8ab93c).

However, when building now ttk with the new module (MyNewTestModule), a linking error systematically occurs:
../../lib64/libttkMyNewTestModuleCS.a(ttkMyNewTestModuleClientServer.cxx.o): In function `ttkUtils::GetVoidPointer(vtkDataArray*, long long)':
ttkMyNewTestModuleClientServer.cxx:(.text+0x10): multiple definition of `ttkUtils::GetVoidPointer(vtkDataArray*, long long)'
../../lib64/libttkHelloWorldCS.a(ttkHelloWorldClientServer.cxx.o):ttkHelloWorldClientServer.cxx:(.text+0x10): first defined here

I guess this one is for @JonasLukasczyk  and @CharlesGueunet as both of you guys recently touched the involved parts of ttkUtils.

pierre-guillou:
I can confirm that this bug also happens to me.

I think the problem is caused by the functions in ttkUtils being defined in the header file, instead of being defined in a separate .cpp file.

One quick workaround is to move the ```#include <ttkUtils.h>``` statement from ttkHelloWorld.h to ttkHelloWorld.cpp. This makes the ttkUtils functions only visible to the ttkHelloWorld module and avoids collisions.

JonasLukasczyk:
PR #361 fixes this issue. I guess we can close this issue.

julien-tierny:
alright, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #353, opened 2020-05-19
Separate declaration and implementation in ttkUtils

CharlesGueunet:
Dear Julien,

This PR address #351 by splitting the **ttkUtils.h** files into the usual declaration and implementation scheme (cpp + h).
The linking problem does not occur anymore this way.

Additionally, I have moved the `ttkUtils.h` include in the cpp file in the module `ttkHelloWorld` as it is better to limit the number of include in header files when possible, as suggested by @pierre-guillou .

Best,
Charles


julien-tierny:
hi charles,

thanks a lot for the quick fix :)

> Additionally, I have moved the `ttkUtils.h` include in the cpp file in the module `ttkHelloWorld` as it is better to limit the number of include in header files when possible, as suggested by @pierre-guillou .
quick question: it seems to me that it's preferable to minimize the number of #include call in a cpp source file and instead to move them to headers as much as possible. what's the recommended good practice at kitware on that matter?

also, can you please update your PR with pierre's azure fix (PR #347)

thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

CharlesGueunet:
I have rebased the PR on the dev branch containing the azure fix. Let's see if tests are passing :)

For the include, it is usually considered a good practice to limit the number of include on header files to only the required ones. When it is possible, it is better to only include in the cpp file: [stackoverflow](https://stackoverflow.com/questions/16821763/where-should-i-include-header-files-in-the-cpp-or-in-the-h-file/16821899)

At Kitware, we enforce this rule, we even use forward declaration when possible to avoid some includes in header files. For example, if you class only manipulate `vtkImageData*`, it does not need to know how to construct this a vtkImageData so the include may be replaced by a forward declaration:
```
class vtkImageData;
```
Then, the include is done in the corresponding cpp files.

JonasLukasczyk:
I opened a similar PR that now also includes the features of this PR and more. So I guess we can close this one.

Best
Jonas

PR #356, opened 2020-05-20
TTKCinemaWriter: Support avoiding to forward input to output

pierre-guillou:
The TTKCinemaWriter filter shallow copies its input to its output. In an Catalyst environment where filters are triggered by downstream Catalyst writers, this could lead to writing twice the same data: the first in the Cinema Database through TTKCinemaWriter and the second time with the downstream Catalyst writer. 

Even though the Catalyst writer can overwrite previously-written data (and, this way, reducing disk occupancy), the second disk write is still happening.

I propose with this PR to add a new property to TTKCinemaWriter that controls the shallow copying of the input. When activated, it prevents this shallow copy and acts like a sink node for input data. The downstream Catalyst writer will only receive empty data and the disk write will be reduced to minimum.

Enjoy,
Pierre


julien-tierny:
alright, looks good to me.
thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #357, opened 2020-05-22
PersistenceDiagramClustering: Fix switching the distance approach in ttk-data/persistenceDiagramDistance.pvsm

pierre-guillou:
This PR clean some member variables of the `ttkPersistenceDiagramClustering` class before computation.
This fixes the bug in the ttk-data state `persistenceDiagramDistance.pvsm` when switching from the auction to the progressive approach led to data corruption.

Enjoy,
Pierre


julien-tierny:
ok, looks good.
thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #359, opened 2020-05-26
Share compiler warnings between base, VTK and standalone layers

pierre-guillou:
Since the ParaView 5.7 migration, compiler flags that were previously applied on both base and VTK layer have been reduced to only the base layer. Such flags include:
* -Wall
* -Wshadow
* -march=native
* /W4 for MSVC

This PR reworks the corresponding CMake code to gather all compiler flags into a CMake variable, TTK_COMPILER_FLAGS, used in base, VTK and standalone. I chose this approach instead of turning the target_compile_options from PRIVATE to PUBLIC in CMake/BaseCode.cmake to avoid warnings from VTK-generated code (for instance the Python wrapper).

The `ttk_add_base_template_library` has been left untouched by this modification (apart from one small fix). In the majority of the TTK setups, base template libraries are compiled through the VTK layer.

Subsequent compiler warnings were fixed in the TTK codebase. Most of them were variable renaming to accomodate -Wshadow. TTK should now build without warnings with GCC and LLVM.

Other small CMake quirks were also fixed during the process.

@CharlesGueunet Do you think this is the right way to do this?

Enjoy,
Pierre

### DIFF
--- a/CMake/BaseCode.cmake
+++ b/CMake/BaseCode.cmake
@@ -145,7 +145,7 @@ endfunction()
 function(ttk_set_compile_options library)
 
   # compilation flags
-  target_compile_options(${library} PRIVATE ${COMPILER_FLAGS})
+  target_compile_options(${library} PRIVATE ${TTK_COMPILER_FLAGS})
 
   if(Boost_FOUND)
     target_link_libraries(${library} PUBLIC Boost::boost)
--- a/CMake/BaseCode.cmake
+++ b/CMake/BaseCode.cmake
@@ -247,7 +247,7 @@ function(ttk_find_python)
   find_package(PythonLibs QUIET)
 
   if(PYTHON_INCLUDE_DIRS)
-    include_directories(${PYTHON_INCLUDE_DIRS})
+    include_directories(SYSTEM ${PYTHON_INCLUDE_DIRS})
 
     #     if (${CMAKE_VERSION} VERSION_GREATER "3.12"
     #       OR ${CMAKE_VERSION} VERSION_EQUAL "3.12")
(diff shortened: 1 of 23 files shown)


CMake/BaseCode.cmake
```diff
@@ -64,7 +64,7 @@ function(ttk_add_base_template_library library)
 
   string(TOUPPER "${CMAKE_BUILD_TYPE}" uppercase_CMAKE_BUILD_TYPE)
   if(uppercase_CMAKE_BUILD_TYPE MATCHES RELEASE)
```
vollia, line 66: I think this whole if can be removed if this flag is already set by the new `CompilerFlags.cmake`

vollia: (Having this flag INTERFACE was not good anyway)

standalone/CMakeLists.txt
```diff
@@ -9,6 +9,8 @@ if(NOT TARGET ttkUserInterfaceBase)
   set(gui_dirs "")
 endif()
 
+add_compile_options(${TTK_COMPILER_FLAGS})
```
vollia, line 14: Using the non targeted version is deprecated, but honestly for the standalone if its avoid modifying each files I can accept this solution.

CharlesGueunet (commented):
Nice

CMake/BaseCode.cmake
pierre-guillou: Isn't this block still useful when building an application against the base layer only?

standalone/CMakeLists.txt
pierre-guillou: Yeah, this was the only way I found to do it in a minimum number of lines (too many files to modify otherwise). I agree that this is not really following best practices.

julien-tierny: that'd be nice to add a comment here in the cmake file, like:
```
# warning deprecated
```

julien-tierny:
apart for the warning comment on the deprecated call, that looks good to me.
@CharlesGueunet: is merging ok for you?

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

standalone/CMakeLists.txt
pierre-guillou: I've added a warning on the `add_compile_option` call.

CharlesGueunet:
Yes, it is ok for me.
@julien-tierny would you want TTK developers to use a standard review mechanism ?

julien-tierny:
alright, thanks.

> @julien-tierny would you want TTK developers to use a standard review mechanism ?
does not look necessary for now, maybe in the future.
(for the moment I just wanted to make sure you were on board with this).

cheers
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #361, opened 2020-05-27
CreateModule fix

JonasLukasczyk:
This commit just moves the ttkUtils.h inclusion in .../core/vtk/ttkHelloWorld.h to .../core/vtk/ttkHelloWorld.cpp. This also removes the dependency to ttkTriangulationAlgorithm, but needed some small additions to the new ttkMacros.h files.

Note, the inclusion of ttkUtils that just defines a namespace with static functions causes for some reason linking errors if included in other header files! So we need to always include ttkUtils.h in the .cpp files (this is also the actual place where it should be).

### DIFF
--- a/core/vtk/ttkAlgorithm/ttkMacros.h
+++ b/core/vtk/ttkAlgorithm/ttkMacros.h
@@ -1,3 +1,24 @@
+#pragma once
+
+#define TTK_COMMA ,
+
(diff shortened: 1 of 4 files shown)


julien-tierny:
alright, thanks.
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

Issue #362, opened 2020-05-28
IcoSphereFromPoint reports invalid xyz coordinates


julien-tierny:
**Describe the bug**
Computing IcoSphereFromPoint from an input point clouds fails, reporting invalid xyz coordinates (+e38)

**To Reproduce**
Load the following state file (which uses ttk-data/clustering0.csv)
https://www-apr.lip6.fr/~tierny/private/icoSphereFromPoint.pvsm

**Expected behavior**
Sphere centered at each point (see the correct result with SphereFromPoint on the left view)

**Screenshots**
![screenshot](https://user-images.githubusercontent.com/3018628/83101581-ef70ee80-a0b2-11ea-84e4-139613b47bf5.png)

**System (please complete the following information):**
linux, gcc

**Additional context**
This one is for @JonasLukasczyk 
Looks like something's not right in IcoSphereFromPoint (float/double issue?)

JonasLukasczyk:
Hi, I can reproduce the bug which is related to two things:
1) TabelToPoints does not produce a valid vtkPolyData object (ParaView just does not crash). The generated object just has 1 cell and 1500 points. This filter should actually create a cell for each point.
2) IcoSphereFromPoint currently only supports float coordinates, but TabelToPoints seems to produce double coordinates.

I can work on both these issues during the migration.

julien-tierny:
hi jonas,

thanks for your quick reply.

> 1) TabelToPoints does not produce a valid vtkPolyData object 
I don't know what you mean here.
TableToPoints is doing what it should.
(fyi, it's a vanilla filter)

> 2) IcoSphereFromPoint currently only supports float coordinates, but 
yes, you need to template that for float and doubles.

> I can work on both these issues during the migration.
as you want, but you will run into this issue a lot in the state files of ttk-data.

cheers
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

JonasLukasczyk:
1) In my PV the information of the object produced by TableToPoints states that the object consists of 1 cell and 1500 points. This is incorrect, it should state 1500 cells and 1500 points, where the cells are all of type VTK_VERTEX. Even vanilla filters are not perfect ;-)

2) You are absolutely right about the precision. This feature is already waiting in the tukl branch ;-)

julien-tierny:
> 2) You are absolutely right about the precision. This feature is already waiting in the tukl branch ;-) 
[...]

PR #364, opened 2020-05-29
icosphere upgrades: +double precision support; +optional normal compu…

JonasLukasczyk:
Hi Julien,
this PR provides the promised upgrades and are working on my machine on all ttk-data statefiles (as expected since the Paraview API did not change). However, I noticed that I called the modules IcoSphereX instead of IcosphereX which would be the correct spelling. I will open a new PR that renames those modules, which also means I need to open a new PR on ttk-data.

Best
Jonas

### DIFF
--- a/core/base/icoSphere/IcoSphere.h
+++ b/core/base/icoSphere/IcoSphere.h
@@ -30,9 +30,9 @@ namespace ttk {
      * Efficiently computes for a given subdivision level the number of
      * resulting vertices and triangles.
      */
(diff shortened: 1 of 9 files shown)


JonasLukasczyk:
All builds succeed and this PR will solve issue #362

julien-tierny:
awesome, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #366, opened 2020-06-02
Fixes for CinemaWriter/TopologicalCompression, PersistenceDiagramDistanceMatrix

pierre-guillou:
This PR fixes bugs related to VESTEC pipelines and some were arguably introduced by me in previous PRs. Those include:

- a race condition in a (not so) parallel loop in PersistenceDiagramDistanceMatrix (no time penalty by removing the parallelism)
- one buffer overflow in TopologicalCompression leading to a segfault in the VESTEC Space Weather  prototype after 80 to 120 cycles. This segfault was occuring in the zlib compression routine but was ultimately linked to the state of the TopologyCompressionWriter member variable inside the CinemaWriter class. This member variable was not reset after each use. In the end, the member variable was demoted to a local variable to get a clean state before each use.
- a second buffer overflow in the TopologicalCompression. This one cause the read of garbage values out of the bounds of a vector but does not provoke side-effects (and the overflow seems to be managed by the code). To avoid further complaints of the Address Sanitizer, the vector has been padded.
- a buffer overflow in CinemaProductReader.
- some ImplicitTriangulation copies were replaced by references (for std::array with fundamental types, move is a copy).

Enjoy,
Pierre

### DIFF
--- a/core/base/implicitTriangulation/ImplicitTriangulation.cpp
+++ b/core/base/implicitTriangulation/ImplicitTriangulation.cpp
@@ -2684,7 +2684,7 @@ int ImplicitTriangulation::preconditionVerticesInternal() {
 #pragma omp parallel for num_threads(threadNumber_)
 #endif // TTK_ENABLE_OPENMP
     for(SimplexId i = 0; i < vertexNumber_; ++i) {
-      std::array<SimplexId, 3> p{};
+      auto &p = vertexCoords_[i];
       vertexToPosition2d(i, p.data());
 
       if(0 < p[0] and p[0] < nbvoxels_[Di_]) {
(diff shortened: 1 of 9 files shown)


julien-tierny:
hi pierre,

thanks a lot for this PR.

can you please re-sync it with the dev branch?
(charles' PR on vtk9/pv58 migration just got merged)

thanks!
julien
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
looks good, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #368, opened 2020-06-08
Fix for ZFP and new layout

CharlesGueunet:
Dear Julien,

This PR fix the C++ example for new layout version.
Also add a small fix for ZFP to be exported correctly while installing TTK.

Charles


config.cmake
```diff
@@ -193,12 +193,9 @@ else()
   message(STATUS "SQLITE3 not found, disabling SQLITE3 support in TTK.")
 endif()
 
-find_package(ZFP QUIET)
```
vollia, line 196: This change is meant to allow to disable zfp even if `ZFP_DIR` is set (for TTK based applications)
Otherwise, if `ZFP_DIR` is set, even with `TTK_ENABLE_ZFP=OFF` the target `zfp::zfp` is created and TTK based application link against zfp.

julien-tierny:
ok, looks good thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #370, opened 2020-06-08
Fix ZFP target and port c++ example to new layout

CharlesGueunet:
Dear Julien,

This PR is he new version of #368 that does not change the ZFP default behavior while still fixing the target bug.

Charles


julien-tierny:
alright, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #371, opened 2020-06-09
Fix Python wrapping for ttkTriangulationAlgorithm

lhofmann:
Python wrapping `ttkTriangulationAlgorithm` has produced undefined symbols (e.g., `PyvtkPolyData_ClassNew`), because the TTK classes `ttkImageData`, `ttkPolyData` and `ttkUnstructuredGrid` have a non-VTK base class.

This PR fixes wrapping by simply hiding the non-VTK base classes from the VTK wrapper. Similar workarounds can be found in the VTK code base itself, for example in [VTK/Common/Core/vtkDoubleArray.h](https://gitlab.kitware.com/vtk/vtk/-/blob/master/Common/Core/vtkDoubleArray.h#L31).

### DIFF
--- a/core/vtk/CMakeLists.txt
+++ b/core/vtk/CMakeLists.txt
@@ -81,9 +81,7 @@ if(VTK_WRAP_PYTHON)
   # Generate __init__.py
   set(InitContent "from __future__ import absolute_import\n")
   foreach(MODULE ${TTK_ENABLED_MODULES})
-    if (NOT "${MODULE}" STREQUAL "ttkTriangulationAlgorithm")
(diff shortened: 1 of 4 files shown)


julien-tierny:
hi lutz,

thanks a lot for your PR.

can you elaborate a little bit on the underlying issue? 
the example in examples/python/ runs fine here.
was that a problem for anaconda?

have you tested this commit against the example in examples/python/ by any chance?

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

Issue #372, opened 2020-06-10
ZFP support in topological compression is broken


julien-tierny:
**Describe the bug**
ZFP support in topological compression is broken.

**To Reproduce**
- go to the ttk-data repository
- run 
```
paraview states/persistenceDrivenCompression.pvsm
```
The 2 bottom views are black.

**Expected behavior**
See this screenshot from the gallery page of TTK's website
https://topology-tool-kit.github.io/img/gallery/persistenceDrivenCompression.jpg

**Screenshots**
![topologicalCompressionBroken](https://user-images.githubusercontent.com/3018628/84252826-670d3780-ab0f-11ea-924d-f842a74d8bb7.png)

**System (please complete the following information):**
 - OS: gentoo
 - Compiler gcc
 - Installed TTK's optional dependencies: all

**Additional context**
This has broken very recently, I suspect after merging either PR #370 or #366.
So it looks like this one is for @CharlesGueunet and/or @pierre-guillou 

@pierre-guillou
I suspect it is a consequence of fixes on zlib's usage, #366. 
one quick work around would be to re-compress these data sets for the state file to work again.
but in the end, what it means is that the current version of ttk is not able to load anything that's been written to disk prior to #366, which is really a bummer for retro-compatibility.
would it be possible to support both behaviors?

in any case, it looks like we need to write in the header of *ttk files the version of ttk that's been used to save the file (if we want to have multiple implementations down the road, for retro-compatibility).

julien-tierny:
so actually, I have just tried compressing and loading back and the code actually crashes.
this is even worse.

pierre-guillou:
On my computer, the same problem occurs after the #370 merge. However, I can't reproduce it between the #366 merge and the #370 merge.

I will try to search for the root cause.

pierre-guillou:
Here are a verbose version of the TopologicalCompression module logs:

> [TopologicalCompression] Could not find magic bytes in input file!
> [TopologicalCompression] File may be corrupted!
> [TopologicalCompression] Successfully read metadata.
> [TopologicalCompression] Successfully read compressed data.
> [TopologicalCompression] Successfully uncompressed data.
> [TopologicalCompression] Successfully read topology.
> [TopologicalCompression] Successfully read geomap.
> [TopologicalCompression] Attempted to read a ZFP block but ZFP is not installed.
> [TopologicalCompression] Failed to write (possibly ZFP)!
> [TopologicalCompression] File may be corrupted!

[...]

PR #374, opened 2020-06-10
Add missing compile option

CharlesGueunet:
Dear Julien,

This PR address #372 . The `TTK_ENABLE_ZFP` preprocessor marco was removed instead of moved in MR #370 .

Charles

### DIFF
--- a/core/base/topologicalCompression/CMakeLists.txt
+++ b/core/base/topologicalCompression/CMakeLists.txt
@@ -11,10 +11,14 @@ ttk_add_base_library(topologicalCompression
     topologicalSimplification
   OPTIONAL_DEPENDS
     zfp::zfp
-    ${ZLIB_LIBRARY}
+    ZLIB::ZLIB
   )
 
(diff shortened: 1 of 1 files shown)


core/base/topologicalCompression/CMakeLists.txt
```diff
@@ -11,10 +11,14 @@ ttk_add_base_library(topologicalCompression
     topologicalSimplification
   OPTIONAL_DEPENDS
     zfp::zfp
-    ${ZLIB_LIBRARY}
+    ZLIB::ZLIB
```
julien-tierny, line 14: what if zlib has been disabled?

julien-tierny: (topological compression can work in principle even if zlib is not available)

julien-tierny:
alright, thanks for the quick fix.
cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #375, opened 2020-06-14
hello world standalone

JonasLukasczyk:
Hi Julien,
this PR adds a hello world standalone program and modifies the standalone CMakeLists.txt file to also build programs that are not inside a cmg/gui folder.

Some things to consider:
* The ttkCommandLineParser (ttkCLP) does not correctly print integer arguments. For example, after setting the thread level to 4, the ttkCLP prints this parameter as an empty string: "-t: ".
* ttkCLP does not support bool parameters.
* The "-l" option controls if the program should list all available arrays, but the current ttkCLP requires that the value of "l" must be set explicitly. So instead of "ttkHelloWorld -i SomePath -l" one has to write "ttkHelloWorld -i SomePath -l 1", which is a bit unintuitive.
* Since the standalone builds already require vtk, I champion the vtksys/CommandLineArguments (vtkCLP) class to replace the ttkCLP. If I remember correctly, you pointed out that the advantage of the ttkCLP is that it already sets some default values (debug level/thread number), but this could also be achieved by just inheriting from vtkCLP. I think this would reduce the amount of code we have to maintain, and it solves the issues mentioned above. Here is the corresponding header file:
https://github.com/Kitware/VTK/blob/master/Utilities/KWSys/vtksys/CommandLineArguments.hxx.in

Best
Jonas

### DIFF
--- a/core/base/common/CommandLineParser.h
+++ b/core/base/common/CommandLineParser.h
@@ -75,7 +75,7 @@ namespace ttk {
 
         if(isAnOption_) {
           s += " (default: ";
-          s += *boolValue_;
+          s += std::to_string(*boolValue_);
           s += ")";
         }
 
@@ -109,9 +109,12 @@ namespace ttk {
     };
 
     CommandLineParser() {
+      // running from the command line, add a default value.
+      ttk::globalDebugLevel_ = 3;
       setArgument("d", &(ttk::globalDebugLevel_), "Global debug level", true);
       setArgument("t", &ttk::globalThreadNumber_, "Global thread number", true);
       debugLevel_ = (int)(debug::Priority::INFO);
+      
       setDebugMsgPrefix("CMD");
     };
 
@@ -183,6 +186,8 @@ namespace ttk {
           }
         }
       }
+      
+      setDebugLevel(ttk::globalDebugLevel_);
 
       return 0;
     };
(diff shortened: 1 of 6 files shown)


JonasLukasczyk:
The azure pipeline failed without an actual build error. Looks like an azure bug.

julien-tierny:
hi jonas,

thanks for the PR.

> * The ttkCommandLineParser (ttkCLP) does not correctly print integer arguments. For example, after setting the thread level to 4, the ttkCLP prints this parameter as an empty string: &quot;-t: &quot;.
fixed (regression due to the shift in debug api).

> * ttkCLP does not support bool parameters.
yes it does (see setOption, fixed).

> * Since the standalone builds already require vtk, I champion the vtksys/CommandLineArguments (vtkCLP) class to replace the ttkCLP.
I'd rather not for the moment, for various reasons.

also included in my commits:
- the vtk wrapper now indicates which array it's using (that should be a standard behavior)
- removed unnecessary dep on ttkProgramBase
- default output prefix
- TODO instructions (for filling up a new module based on HelloWorld)
- default strategy (used to crash otherwise)
- logging based on new debug api
- updated module scripts

cheers,
julien
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #381, opened 2020-06-17
Improve azure config file

CharlesGueunet:
Dear Julien,

This PR fix the detection of the new layout to use the unofficial 8.90 version of VTK. It is important to compile TTK against ParaView 5.8 when the ParaView plugin is disabled.

I have also removed a `vtkStandardNewMacro` on a non `vtkObject`.

Charles

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,10 +12,12 @@ jobs:
 # Linux build
 # ------------------------------
 
-# VTK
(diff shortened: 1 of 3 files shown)


julien-tierny:
alright, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #383, opened 2020-06-23
PersistenceDiagramDistanceMatrix: Support more ways to filter pairs

pierre-guillou:
This PR adds the support of three new ways of filtering out pairs in the PersistenceDiagramDistanceMatrix module:

- filtering out pairs under a global, absolute persistence level threshold,
- filtering out pairs under a per-diagram persistence threshold relative to the diagram's highest persistence pair,
- filtering out pairs under a persistence threshold relative to the highest persistence pair of all diagrams.

The two other methods, using all pairs and only using the N most persistence pairs, were left unaltered. Although the code for using the N most persistence pairs is complex and might be buggy (according to Jules), it is quicker and seems to produce better results than my alternative and simpler re-implementation. This is still the default filter method, however, I raised the default number of pairs to keep from 10 to 20 (after some tests, 10 seems to not be enough).

Enjoy,
Pierre

### DIFF
--- a/core/base/persistenceDiagramDistanceMatrix/PersistenceDiagramDistanceMatrix.cpp
+++ b/core/base/persistenceDiagramDistanceMatrix/PersistenceDiagramDistanceMatrix.cpp
@@ -269,25 +269,64 @@ void PersistenceDiagramDistanceMatrix::enrichCurrentBidderDiagrams(
         }
       }
     }
+    return;
+  }
 
-  } else if(this->Constraint == ConstraintType::NUMBER_PAIRS) {
-    for(size_t i = 0; i < nInputs; ++i) {
-      // index of pair in bidder_diags + corresponding persistence value
-      using PairIV = std::pair<size_t, double>;
-      std::vector<PairIV> pairs(bidder_diags[i].size());
-      for(int j = 0; j < bidder_diags[i].size(); ++j) {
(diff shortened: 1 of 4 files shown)


julien-tierny:
hi pierre, 

thanks a lot!

> This is still the default filter method, 
could you switch the default method to global absolute persistence (with a threshold at 0.1 for instance)?
that seems like the more generic strategy if you have no a priori knowledge about your data.

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

pierre-guillou:
Do you mean absolute or relative? Relative seems more generic.

julien-tierny:
> Do you mean absolute or relative? Relative seems more generic.
sorry, good catch :)
I meant relative indeed.
cheers
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

pierre-guillou:
Done!

julien-tierny:
thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
looks good,
thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

Issue #382, opened 2020-06-23
MatrixToHeatMap Paraview UI won't show RegExp edit line


hugo-manet:
When "Select Fields with a Regexp" is checked, the corresponding regexp doesn't show.

**To Reproduce**
1. Load a TTKMatrixToHeatMap in paraview
2. Check "Select Fields with a Regexp" in "Input Options"

**Expected behavior**
The "Regexp" line should appear, as suggested in the .XML file

**System (please complete the following information):**
 - OS: Ubuntu
 - Paraview Version 5.8.0
- TTK Version 0.9.9, dev branch

**Fix**
Remove `panel_visibility="advanced"` of the corresponding widget in the .XML file, line 76

julien-tierny:
good catch.
can you PR the fix and mention that issue in the PR?

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #384, opened 2020-06-23
Fix issue #382

hugo-manet:
Fix issue #382 by removing an option to the corresponding Paraview UI widget


julien-tierny:
hi hugo,

thanks a lot for the PR.

actually I'd argue that regexp are probably for the most advanced users and should probably be left in "advanced" mode. 
I'd suggest to move the checkbox "SelectFieldsWithRegexp" to "advanced" as well instead (the problem was that one was advanced and not the other).

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
cool, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #386, opened 2020-06-23
Remove some manual memory management

pierre-guillou:
This PR replaces manual memory management in several filters by better alternatives:

- in KDTree and PersistenceDiagramClustering, raw pointers were wrapped into std::unique_ptr,
- in ContourAroundPoint, std::vectors are used to replace heap arrays (and switching to LongSimplexId avoids a vtkIdType -> SimplexId vector data conversion)
- in ZeroSkeleton, heap-allocated vectors are placed on the stack. The parallel algorithm performance is still inefficient though and remains disabled.
- in ProgramBase, the heap-allocated member is placed on the stack.

No change has been observed in the ContourAroundPoint and the PersistenceDiagramClustering states in the corresponding ttk-data validation states.

Enjoy,
Pierre


core/base/skeleton/ZeroSkeleton.cpp
```diff
@@ -531,21 +531,18 @@ int ZeroSkeleton::buildVertexStars(
   printMsg("Building vertex stars", 0, 0, threadNumber_,
            ttk::debug::LineMode::REPLACE);
 
-  vertexStars.resize(vertexNumber);
```
julien-tierny, line 534: where did that go for threadNumber_ != 1?

pierre-guillou: For threadNumber_ > 1, it is done at line 543.
For threadNumber_ == 1, in the std::move at the end of the computation (line 594).

julien-tierny: gotcha

julien-tierny:
cool, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #387, opened 2020-06-23
Parallelize some Morse-Smale Complex routines

pierre-guillou:
The current PR parallelize some routines in the MorseSmaleComplex module(s):

- getAscendingSeparatrices2, getDescendingSeparatrices2 and getSaddleConnectors were parallelized using the same "trick": walls are stored per thread in a boolean array (for fast access and light memory footprint) but also in a SimplexId vector (for fast cleanup of the previous boolean array while avoiding de-allocation).
- setSaddleConnectors has been merged into setSeparatrices1 (the two functions were very similar).
- the 3 setSeparatrices1 calls (for ascending, descending 1-seps and saddle connectors) were merged into one to reduce memory allocations.
- the omp_setAscendingSeparatrices2 and omp_setDescendingSeparatrices2 methods were removed (and with them, the PrioritizeSpeedOverMemory parameter).
- the setAscendingSeparatrices2 and setDescendingSeparatrices2 methods were parallelized (more efficiently than in omp_set{A|De}scendingSeparatrices2). However, the order of the points in the points coordinates array might have changed.
- the setFinalSegmentation has been parallelized. Even though the resulting manifold indices should be different than in the ones in the old sequential version, the indices are computed to be deterministic and use a compact range (from 0 to number of segmentations - 1).
- among other small fixes, a conversion bug for saddle-saddle persistence pairs has been fixed.

All these changes should lead to a 4x speedup for computing the Morse-Smale complex of ctBones.vti (ttkMorseSmaleComplexCmd: 130s -> 45s). The memory footprint should be similar. No change has been observed in the ttk-data states using the Morse-Smale complex (especially morseMolecule).

There is still a lot of time spent in filling (sequentially) the VTK data structures. I'll try to use class members + vtkDataArray::setArray during the migration to the new API next week. And there is still on-going work on returning saddle connectors that is left to a future PR.

Enjoy,
Pierre


julien-tierny:
cool, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #389, opened 2020-06-23
Add non-square matrices support, with a warning

### LINKED ISSUE #385: MatrixToHeatMap should accept non-square matrices
**Is your feature request related to a problem? Please describe.**
Distance matrices are generally used with corresponding lines and columns (ie self-distance matrices), but not always. We should accept all distance matrices.
We keep a warning whenever the matrix isn't square, which indicates that it isn't a self-distance matrix, for example when there are metadata fields in the data matrix.

**Describe the solution you'd like**
Accept any matrix that can be converted to double.

**Describe alternatives you've considered**
Maybe add some more sanity checks to decide between : 
- the matrix should be a square self-similarity matrix, but some metadata is polluting it
- the matrix isn't meant to be square at all, we can display it as a whole
To help in this, maybe look for a 0 diagonal ?

hugo-manet:

MatrixToHeatMap works with non-square matrices :
- Replace the error message + return 0 by a warning
- Adapt code for two different dimensions

Fix #385

### DIFF
--- a/core/vtk/ttkMatrixToHeatMap/ttkMatrixToHeatMap.cpp
+++ b/core/vtk/ttkMatrixToHeatMap/ttkMatrixToHeatMap.cpp
@@ -57,15 +57,15 @@ int ttkMatrixToHeatMap::RequestData(vtkInformation * /*request*/,
   }
 
   const auto nInputs = ScalarFields.size();
-  if(nInputs != static_cast<size_t>(input->GetNumberOfRows())) {
-    this->printErr("Distance matrix is not square");
-    return 0;
+  const auto nRows = static_cast<size_t>(input->GetNumberOfRows());
+  if(nInputs != nRows) {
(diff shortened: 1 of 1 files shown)


julien-tierny:
cool, thanks!

@pierre-guillou: can you double-check everything's ok for you?
(actually, we want to propagate that back up to other matrix based modules. to discuss)
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

pierre-guillou:
Looks fine to me.
I'll check on how to support that in LDistanceMatrix and PersistenceDiagramDistanceMatrix.

julien-tierny:
awesome, thanks
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #390, opened 2020-06-23
ImplicitTriangulation: Cache vertex neighbor accesses

pierre-guillou:
This PR replaces some computations with array accesses in the ImplicitTriangulation module. The arrays are populated during the precondition. Only the getVertexNeighbor is concerned (for now, hopefully).

This should speedup filters that rely heavily on the triangulation. No changes observed in the ttk-data states.

Enjoy,
Pierre


julien-tierny:
awesome, thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #391, opened 2020-06-23
Change ParaView / VTK detection strategy:

CharlesGueunet:
Dear Julien,

With this PR, TTK always try to find ParaView, then VTK.
This approach allows to compile TTK with the VTK inside ParaView instead
of looking for a VTK when the `TTK_BUILD_PARAVIEW_PLUGINS` is set to **OFF**.

On the other side, it is now up to the developer to check the VTK used by TTK is
the one he wants when `TTK_BUILD_PARAVIEW_PLUGINS` is **OFF** if he do not wish
to use the one inside ParaView.

Charles


julien-tierny:
alright, looks good.
thanks!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #392, opened 2020-06-24
Random fixes

pierre-guillou:
This PR fixes some compiler warnings, mostly from MSVC, some others with advanced GCC warnings such as       -Wduplicated-cond, -Wduplicated-branches or -Wuseless-cast (c.f. https://github.com/lefticus/cppbestpractices/blob/master/02-Use_the_Tools_Available.md#compilers). Incidentally, TTK should now build without warnings with -Wextra -Wno-unused-parameter enabled. 

This PR includes some small code modernization: `typedef` were replaced with `using` when applicable, deprecated std::random_shuffle has been replaced with std::shuffle, etc.

Custom filters have been removed from the Example1.pvsm.

A fix for the implicit triangulation caching is also included.

No changes observed in the ttk-data states.

Enjoy,
Pierre


julien-tierny:
ok.
cheers
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #401, opened 2020-06-29
[api-migration-hackathon] ttkScalarFieldSmoother

julien-tierny:
This is work-in-progress. Still needs:

- [x] work on order in templating

- [x] work on logic for mask field (vtk layer)

- [x] work on ttk-data state files (ScalarFieldSmoother & GeometrySmoother)


julien-tierny:
companion ttk-data PR: https://github.com/topology-tool-kit/ttk-data/pull/31

PR #405, opened 2020-06-30
[api-migration-hackathon] ttkAlgorithm and ttk::ExplicitTriangulation now validate input cells …

JonasLukasczyk:
This PR adds validity checks to test if cells can actually be triangulated during the ttkAlgorithm::GetTriangulation call. These checks test if:
1) Input cells are homogeneous;
2) Input cells are of dimension 1-4;
3) and if cells are actually simplices (vertices, lines, triangles, and tetrahedra).


julien-tierny:
hi jonas,

thanks a lot for the PR.

> 2) Input cells are of dimension 1-4;
you mean 1-3 right?

the generated messages mention ttkExtract to isolate simplices of a fixed diemnsion. did you modify that too?

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

JonasLukasczyk:
Sorry, you are correct. Actually I meant 0-3 dimensional.

I'm working on the ttkExtract upgrade (that will be another PR)

julien-tierny:
@CharlesGueunet
can you have a quick look at this PR to confirm that the accesses to the cell arrays are compliant with your new interface?

thanks
julien 
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -165,18 +150,20 @@ ttk::Triangulation *ttkAlgorithm::InitTriangulation(void *key,
...
         = (vtkIdType *)ttkUtils::GetVoidPointer(cells->GetConnectivityArray());
       auto offsets
         = (vtkIdType *)ttkUtils::GetVoidPointer(cells->GetOffsetsArray());
-      triangulation->setInputCells(
-        cells->GetNumberOfCells(), connectivity, offsets);
-#else
-      triangulation->setInputCells(
-        cells->GetNumberOfCells(), cells->GetData()->GetPointer(0));
-#endif
+      int status = triangulation->setInputCells(nCells, connectivity, offsets);
```
vollia, line 159: Is this ParaView 5.7 compliant ? I am afraid the `setInputCells` method with two parameters may not work for older versions

CharlesGueunet (approved):

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -165,18 +150,20 @@ ttk::Triangulation *ttkAlgorithm::InitTriangulation(void *key,
     }
 
     // Cells
-    if(cells->GetNumberOfCells() > 0) {
-#ifdef TTK_CELL_ARRAY_NEW
```
vollia, line 169: I think this `ifdef` should not be removed for the elder cell array api to be supported (PV 5.7)
There is two `setInputCell` methods and the one used depends on the version of ParaView / VTK we are using.
Elder version represent cells with a single flat array. You may add the validation code only in the newer layout using both `connectivity` and `offset` but you need to still use the elder one when `TTK_CELL_ARRAY_NEW` is not define for ParaView 5.7.

CharlesGueunet (changes requested):

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
JonasLukasczyk: Hi, I can put these guards back in, but is it worth keeping that old code around? We are moving to the next version anyway. Also, as Peter found out, building TTK only with a VTK repo (no paraview) doesn't work right now anyway.

vollia: you need to be on a VTK commit close to the one used by ParaView 5.7 to use an "old" VTK.
Whether we want to still keep the compatibility with PV 5.7 and corresponding VTK is not up to me.

julien-tierny:
> Whether we want to still keep the compatibility with PV 5.7 and corresponding VTK is not up to me.
to my understanding, the support for pv57 is not compatible with pv56 anyways, so it looks like we already committed to move forward without backward compatibility.

I guess initially, we wanted to keep a bit of pv57 support in order not to break everybody's repo upon charles PR's merge.

so maybe, dropping support for pv57 is ok.

the two of you agree?

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

CharlesGueunet:
> so maybe, dropping support for pv57 is ok.
 the two of you agree?

I agree, in this case this PR seems right to me

julien-tierny:
hi jonas,

this PR is currently queueing in the continuous integration.

can you pull ttk:dev and commit on that branch?
(I disabled a few vms on azure for faster builds).

that will reset and start fewer jobs.

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

JonasLukasczyk:
Will do

julien-tierny:
@JonasLukasczyk 
in the future, for PRs which impact all TTK, please remember to test them against ttk-data prior to opening the PR.

PR #410, opened 2020-06-30
[api-migration-hackathon] PersistenceCurve Migration

tbmasood:
This is partial migration of the PersistenceCurve module to the new API spec. I have migrated as much as I could, but this module has a dependency on FTMTree and MSComplex modules which have not been migrated yet. Only a few things are remaining which I have marked as TODOs.

I tested this on a few datasets and seems to be working okay. If these changes are fine, you can merge now. Or should we wait for migration of FTMTree and MSComplex first?


julien-tierny:
hi talha,

thanks a lot for the PR.

> I tested this on a few datasets and seems to be working okay. If these changes are fine, you can merge now. Or should we wait for migration of FTMTree and MSComplex first?
that's a good question.

in principle, we could merge now.
the problem is, when FTM and the MSC will be migrated, the current PR will trigger a lot of errors for the new PR, forcing someone else (@CharlesGueunet, @pierre-guillou) to propagate their changes through dependent modules.

@CharlesGueunet
@pierre-guillou
what's your take?

I'd probably in favor of putting the current PR on hold until FTM and MSC are migrated.

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

CharlesGueunet:
I will propagate my changes on this module once it is merged, do not worry about this.

pierre-guillou:
I'll do the same.

julien-tierny:
btw @tbmasood, have you run the ttk-data examples which use PersistenceCurve?

julien-tierny:
it looks like this PR needs another PR at the ttk-data level in order not to break the examples.
(see section 9 of the migration spreadsheet)

tbmasood:
@julien-tierny No, I have not tested this with ttk-data. I will do a PR for that repo soon.

tbmasood:
Companion data PR: topology-tool-kit/ttk-data#37

julien-tierny:
alright, thanks a lot talha!
I tested with your ttk-data PR and everything works as expected :)
awesome, thanks.

PR #411, opened 2020-06-30
[newModule] ttkDilateErode filter

JonasLukasczyk:
This PR adds a new module called ttkDilateErode which either
a) dilates a specified label by assigning the label of a corresponding vertex to all its neighbors, or
b) erodes a specified label by assigning to a corresponding vertex the largest label among its neighbors.

This filter is useful for overlap based tracking, where one for example wants to cast a wider net by dilating feature labels, or improve reliability by dilating a background label (i.e., eroding feature labels).


julien-tierny:
New modules will be considered after the migration is complete.

julien-tierny:
would that new module deserve a pvsm file for ttk-data?

JonasLukasczyk:
Hi, I will close this PR and open a new one for Version 2.0

PR #412, opened 2020-06-30
[api-migration-hackathon] ttkBlockAggregator Update

JonasLukasczyk:
This PR updates the ttkBlockAggregator module to now accept an unlimited number of input data objects by making the first port repeatable. In the past this filter had 5 hardcoded  optional inputs.

This update effects the nestedTrackingGraph state file in ttk-data (see related PR in ttk-data).


julien-tierny:
have you run it against ttk-data?

PR #418, opened 2020-07-01
[api-migration-hackathon] deleted cell data selector and converter

JonasLukasczyk:
This PR deletes the vtk-only filters ttkCellDataConverter and ttkCellDataSelector. These filters are not used in ttk-data and their functionality is already provided by ttkArrayEditor and vtkPassArrays (minus the regular expression).


julien-tierny:
this one needs no ci

PR #420, opened 2020-07-01
[api-migration-hackathon] New data array management for PointSetToCurve, StringArrayConverter & LDistanteMatrix

pierre-guillou:
In this PR, PointSetToCurve, StringArrayConverter and LDistanceMatrix were converted to use the `GetInputArrayToProcess` and `SetInputArrayToProcess` methods.

Since no ttk-data state file use those filters, no PR for ttk-data.

Enjoy,
Pierre


julien-tierny:
cool, thanks.

PR #423, opened 2020-07-01
[api-migration-hackathon] Tracking from overlap migration

floWetzels:
The TrackingFromOverlap module was migrated to the new API. Changes in style (e.g. member variable definition and use of SetInputArrayToProcess) were omitted, because the module will get a large revision/update soon after the official release. Since no changes were made to how the input array is set, the example state files stayed the same.

### DIFF
--- a/core/vtk/ttkTrackingFromOverlap/ttkTrackingFromOverlap.h
+++ b/core/vtk/ttkTrackingFromOverlap/ttkTrackingFromOverlap.h
@@ -63,7 +63,8 @@ public:
   int FillInputPortInformation(int port, vtkInformation *info) override {
     switch(port) {
       case 0:
-        info->Set(vtkAlgorithm::INPUT_REQUIRED_DATA_TYPE(), "vtkMultiBlockDataSet");
+        info->Set(
+          vtkAlgorithm::INPUT_REQUIRED_DATA_TYPE(), "vtkMultiBlockDataSet");
         break;
       default:
         return 0;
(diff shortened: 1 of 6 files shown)


core/vtk/ttkTrackingFromOverlap/ttkTrackingFromOverlap.h
```diff
@@ -149,10 +130,10 @@ class TTKTRACKINGFROMOVERLAP_EXPORT ttkTrackingFromOverlap
   vector<vector<Edges>> levelTimeEdgesTMap; // E_T
   vector<vector<Edges>> timeLevelEdgesNMap; // E_N
 
-  bool needsToAbort() override {
+  bool needsToAbort() /*override*/ {
```
julien-tierny, line 133: in principle this can go away now (but that's ok for this PR, since it's meant to be temporary).

core/vtk/ttkTrackingFromOverlap/ttkTrackingFromOverlap.h
```diff
@@ -149,10 +130,10 @@ class TTKTRACKINGFROMOVERLAP_EXPORT ttkTrackingFromOverlap
...
     return GetAbortExecute();
   };
-  int updateProgress(const float &progress) override {
+  int updateProgress(const float &progress) /*override*/ {
```
julien-tierny, line 136: in principle this can go away now (but that's ok for this PR, since it's meant to be temporary).

PR #424, opened 2020-07-01
Renamed input port of ttkIcoSpheresFromPoints (NO impact on ttk-data / NO need for CI)

JonasLukasczyk:
Hi Julien,
this tiny PR just changes name of the input port name of ttkIcoSpheresFromPoints in the paraview.xml file. It has no impact on ttk-data and does not require CI. In the future we can profit from this PR by simply running
`sed -i 's/SphereFromPoint/IcospheresFromPoints/g' ./states/*.pvsm`
and manually adjusting the refinement level and enable normals. Then we could delete the SphereFromPoint module.

Best
Jonas


julien-tierny:
in the future, thanks for testing your PRs prior to opening them, especially when you write "NO impact on ttk-data".

PR #427, opened 2020-07-01
[api-migration-hackathon] Import embedding from table migration

floWetzels:
ImportEmbeddingFromTable module was migrated to new API. The module still uses string inputs instead of SetArrayToProcess, because we were not able to use this function for columns of a vtkTable in the paraview layer.
This module has no example state files in ttk-data.

### DIFF
--- a/core/vtk/ttkImportEmbeddingFromTable/ttkImportEmbeddingFromTable.h
+++ b/core/vtk/ttkImportEmbeddingFromTable/ttkImportEmbeddingFromTable.h
@@ -16,60 +16,18 @@
 /// VTK pipeline.
 #pragma once
 
-// VTK includes
-#include <vtkCharArray.h>
-#include <vtkDataArray.h>
-#include <vtkDoubleArray.h>
-#include <vtkFiltersCoreModule.h>
-#include <vtkFloatArray.h>
-#include <vtkInformation.h>
-#include <vtkIntArray.h>
-#include <vtkObjectFactory.h>
-#include <vtkPointData.h>
-#include <vtkPointSet.h>
-#include <vtkPointSetAlgorithm.h>
-#include <vtkSmartPointer.h>
-#include <vtkTable.h>
-
 // VTK Module
 #include <ttkImportEmbeddingFromTableModule.h>
 
 // ttk code includes
-#include <Wrapper.h>
+#include <ttkAlgorithm.h>
 
 class TTKIMPORTEMBEDDINGFROMTABLE_EXPORT ttkImportEmbeddingFromTable
-  : public vtkPointSetAlgorithm,
-    protected ttk::Wrapper {
+  : public ttkAlgorithm{
 
 public:
   static ttkImportEmbeddingFromTable *New();
-  vtkTypeMacro(ttkImportEmbeddingFromTable, vtkPointSetAlgorithm)
-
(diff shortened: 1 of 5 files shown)


core/vtk/ttkImportEmbeddingFromTable/ttkImportEmbeddingFromTable.h
```diff
@@ -105,18 +54,15 @@ class TTKIMPORTEMBEDDINGFROMTABLE_EXPORT ttkImportEmbeddingFromTable
...
 
-  bool UseAllCores;
-  int ThreadNumber;
-
-  int doIt(vtkPointSet *inputDataSet,
-           vtkTable *inputTable,
-           vtkPointSet *output);
-  bool needsToAbort() override;
-  int updateProgress(const float &progress) override;
+  bool needsToAbort();
```
julien-tierny, line 66: again, these can go away. but it's ok for now I guess.

core/vtk/ttkImportEmbeddingFromTable/ttkImportEmbeddingFromTable.h
```diff
@@ -83,20 +41,11 @@ class TTKIMPORTEMBEDDINGFROMTABLE_EXPORT ttkImportEmbeddingFromTable
...
-      info->Set(vtkDataObject::DATA_TYPE_NAME(), "vtkTable");
-
-    return 1;
-  }
-
 protected:
   ttkImportEmbeddingFromTable() {
-    UseAllCores = true;
-
+    this->setDebugMsgPrefix("ttkImportEmbeddingFromTable");
```
julien-tierny, line 46: note: the other modules omit the ```ttk``` prefix in their name. I'll fix that later.

julien-tierny:
alright, looks good.
thanks!

core/vtk/ttkImportEmbeddingFromTable/ttkImportEmbeddingFromTable.h
floWetzels: I thought the other modules omit the ttk prefix in the base layer and it is added automatically in the vtk layer. Is that wrong?

julien-tierny:
actually, with the new inheritance between vtk wrappers and base classes, this distinction is no longer possible unfortunately (@JonasLukasczyk).
cheers
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

JonasLukasczyk:
> actually, with the new inheritance between vtk wrappers and base classes, this distinction is no longer possible unfortunately (@JonasLukasczyk). cheers -- Dr Julien Tierny CNRS Researcher Sorbonne Universite http://lip6.fr/Julien.Tierny
> […](#)
> On Thursday, 2 July 2020 10:49:31 CEST floWetzels wrote: @floWetzels commented on this pull request. > protected: ttkImportEmbeddingFromTable() { - UseAllCores = true; - + this->setDebugMsgPrefix("ttkImportEmbeddingFromTable"); I thought the other modules omit the ttk prefix in the base layer and it is added automatically in the vtk layer. Is that wrong?

It is actually possible to do this by extending the inheritance diamond structure even more. However, my personal preference is to keep the same style and prefix for the vtk and base layer.

PR #428, opened 2020-07-01
[api-migration-hackathon] PersistenceDiagram migration

tbmasood:
This PR is intended for migration of PersistenceDiagram to the new API spec.

@julien-tierny I am currently facing following issues:

- This module is used by TrackingFromFields module. So, any change I do here affects that module. For now, I have commented out the calls to PersistenceDiagram module in TrackingFromFields just to compile TTK and test the current code.
- Same as PR #410 , PersistenceDiagram module has dependence on FTMModule  (@CharlesGueunet) and MorseSmaleComplex (@pierre-guillou ). So, some parts of this module have to be changed appropriately once those modules are ready to complete the migration.


tbmasood:
@julien-tierny Similar to the #410, this PR is ready now, except for the dependence of TrackingFromFields on this module here:
https://github.com/topology-tool-kit/ttk/blob/4ed077a987a1a66dab0e3f778cf120f8df1264b8/core/base/trackingFromFields/TrackingFromFields.h#L106-L121

For now, I have commented out this part of the code in TrackingFromFields to compile and test the PersistenceDiagram module. But, I guess we should not merge this PR as it will temporarily break the TrackingFromFields code until its migration. 

In my opinion, I should wait for TrackingFromFields to be migrated first, then I can do necessary modifications in TrackingFromFields code shown above to comply with the PersistenceDiagram migration. Then we will always have perfectly working code after each PR.

julien-tierny:
thanks talha for the heads up.
I'll let @julesvidal jump in here as he claimed TrackingFromFields.
cheers

julien-tierny:
hi talha, there seems to be some linking issues under macos.
you may want to double check everything is right at this level (check for instance ttk.module and vtk.module)

julien-tierny:
> In my opinion, I should wait for TrackingFromFields to be migrated first, then I can do necessary modifications in TrackingFromFields code shown above to comply with the PersistenceDiagram migration. Then we will always have perfectly working code after each PR. 
actually, I'd recommend to edit TrackingFromFields (just the way it calls PersistenceDiagram) to make sure everything works once your PR is merged (even if TrackingFromFields is not fully migrated).

the rule of thumb is that all ttk-data state files examples should work after the merge of each PR individually.

I think this is the way dependant modules should be handled (this is how @pierre-guillou and @CharlesGueunet will handle your migration of PersistenceCurve and PersistenceDiagram). 

so instead of commenting out TrackingFromFields from field, I'd suggest to make the few necessary edits to make it run.

also, you'll need to PR ttk-data to update all state file examples which use PersistenceDiagram.

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julesvidal:
Hi, sorry for the late reply. 
Either way works for me, but isn'it it easier to migrate persistenceDiagram first ?

I put trackingFromFields on hold for the moment.

tbmasood:
Companion data PR: topology-tool-kit/ttk-data#41

julien-tierny:
this PR is still having issues on two state files from ttk-data:

- [ ] states/timeTracking.pvsm

- [ ] states/tribute.pvsm

talha2
for tribute, the separatrices of the Morse-Smale complex are not elevated as they should (see the bottom right view)
I suspect this might also be a PVSM bug, but I'm not sure.

talha1
for timeTracking, I suspect it's a code issue. 
I put @JonasLukasczyk  in the loop as it may have some more general impact at the API level.
when clicking on the play button of the animation (see where the cursor is located), the persistence diagram does not get updated in the left view, while it should. 
it might be a simple detail to fix, or something more substantial that requires investigation at the GetInputArrayToProcess() level.

JonasLukasczyk:
Its hard to tell whats wrong here since the timeTracking.pvsm does a lot of things. Can someone explain at which point the statefile broke?

julien-tierny:
the state file is not broken but it seems that ttkPersistenceDiagram fails at realizing that the input array has changed since the last call. on the left view, the diagram is static while it should be dynamic when the animation plays (see the mismatch between the field and the critical points).
(you need to clone talha's ttk and ttk-data to further look into this).

JonasLukasczyk:
I'm building the branch right now, but my first guess is that the animation controls a property that has been renamed during the migration (e.g., ScalarField -> ScalarFieldNew).

julien-tierny:
I don't believe that's the issue (the scalar field does get updated in the background, the name remains the same).
as I wrote above, I think it's a failure from ttkPersistenceDiagram to realize the array has been updated.

core/vtk/ttkPersistenceDiagram/ttkPersistenceDiagram.cpp
```diff
@@ -48,139 +35,339 @@ int ttkPersistenceDiagram::FillOutputPortInformation(int port,
...
-#ifndef TTK_ENABLE_KAMIKAZE
-  if(!inputScalars_) {
-    cerr
-      << "[ttkPersistenceDiagram] Error : input scalar field pointer is null."
-      << endl;
-    return -3;
-  }
-#endif
-
-  if(this->GetMTime() < inputScalars_->GetMTime())
```
julien-tierny, line 82: I think that was the bit checking for updates.

JonasLukasczyk: Mh, I don't think it is sensible to compare the modification time of a filter to the modification time of an array. The filter should only update its output if either its input or parameters change. Here it looks like someone (TTK) is overriding in the background an array and then checks if this occurred. Is this the case?

julien-tierny: sensible or not, the example is broken.
@tbmasood could you add this check back in somehow to see if that's the issue?

tbmasood: Yes, this fixes the issue in timeTracking state file. I have just pushed the change.

The tribute state file still has the issue you pointed out.

tbmasood:
@julien-tierny I investigated the issue with the tribute state file a bit more today. It looks like this issue may not be in the persistenceDiagram code but it may be somewhere else. Because, even with current dev branch, although MS complex separatrices are initially correct but as soon as you change the "Scale Factor" in "WrapByScalar" filter to some other value, the separatrices don't change while the MS segmentation changes.

I suspect this may be an issue with MS Complex migration, or may be even more fundamental issue in ttkAlgorithm and triangulation. Unfortunately, I can't figure out what may be the problem on my own.

julien-tierny:
alright, I found a work around for the tribute state file (pushed on your PR).
let's go!

PR #431, opened 2020-07-02
[api-migration-hackathon] ttkDataSetToTable

JonasLukasczyk:
Not used in ttk-data, but I tested it locally.


paraview/DataSetToTable/DataSetToTable.xml
```diff
@@ -45,54 +44,14 @@
...
-      <PropertyGroup panel_widget="Line" label="Testing">
-        <Property name="UseAllCores" />
-        <Property name="ThreadNumber" />
-        <Property name="DebugLevel" />
-      </PropertyGroup>
+      ${DEBUG_WIDGETS}
 
       <Hints>
-        <ShowInMenu category="TTK - High Dimension / Point Cloud Data" />
+        <ShowInMenu category="TTK - Pipeline" />
```
julien-tierny, line 54: not sure about this, but why not.

core/vtk/ttkDataSetToTable/ttkDataSetToTable.cpp
```diff
@@ -1,124 +1,72 @@
...
-  doIt(input, output);
+  this->printMsg("Converting " + targetAttributeName + "Data to Table", 0, 0,
+                 ttk::debug::LineMode::REPLACE);
 
-  {
-    stringstream msg;
-    msg << "[ttkDataSetToTable] Memory usage: " << m.getElapsedUsage() << " MB."
-        << endl;
-    dMsg(cout, msg.str(), memoryMsg);
+  auto input = vtkDataSet::GetData(inputVector[0]);
```
julien-tierny, line 53: you may want to check input is not nullptr here, don't you?

julien-tierny: alright, forget about it. we'll do a global pass on those at a later stage.

paraview/DataSetToTable/DataSetToTable.xml
JonasLukasczyk: The filter seemed to be very generic so I thought it will feel more at home over there.

core/vtk/ttkDataSetToTable/ttkDataSetToTable.cpp
JonasLukasczyk: Yes, I will add a sanity check.

PR #434, opened 2020-07-02
Cmake

kappe-c:

Very simple, see commits / messages

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 # cmake_minimum_required(VERSION 3.12)
-cmake_minimum_required(VERSION 3.10.2) # confirmed by Chris <kappe@cs.uni-kl.de>
+cmake_minimum_required(VERSION 3.15)
 # The C is required to build with VTK when it has MPI
 project(ttk VERSION 0.9.9 LANGUAGES CXX C)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,4 @@
-# cmake_minimum_required(VERSION 3.12)
-cmake_minimum_required(VERSION 3.15)
+cmake_minimum_required(VERSION 3.12.2) # confirmed by Chris <kappe@cs.uni-kl.de>
 # The C is required to build with VTK when it has MPI
 project(ttk VERSION 0.9.9 LANGUAGES CXX C)
 
--- a/CMake/ParaViewFilter.cmake
+++ b/CMake/ParaViewFilter.cmake
@@ -1,8 +1,8 @@
-# fot the ttk_parse_module_file
+# for the ttk_parse_module_file
 include(CMake/VTKModule.cmake)
 
 # register a new filter to build in the TTK plugin
-# deduce the location of the corresonding vtk.module file
+# deduce the location of the corresponding vtk.module file
 # also register the xml file if given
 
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 # cmake_minimum_required(VERSION 3.12)
-cmake_minimum_required(VERSION 3.10.2) # confirmed by Chris <kappe@cs.uni-kl.de>
+cmake_minimum_required(VERSION 3.15)
```
julien-tierny: I'm not so sure about this.
one of my systems has cmake 3.14.6 and built ttk fine (at least before the hackathon).

julien-tierny: @CharlesGueunet  
@JonasLukasczyk 
any thoughts on that?

vollia: Is there a specific instruction in the current state that requires 3.15 ? VTK still requires only 3.12 so I think TTK should too.

JonasLukasczyk: I didn't check the minimum required version. On Arch I use cmake v3.17.

JonasLukasczyk: Paraview only requires 3.10: 
cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
I guess we do not need anything more fancy, do we?

JonasLukasczyk: As Charles pointed out, VTK requires 3.12:
cmake_minimum_required(VERSION 3.8...3.12 FATAL_ERROR)

kappe-c: To quote my commit message (partially hidden by the web gui):
needed for the cmake command *add_compile_definitions*

I had v3.10.2 before and it did not compile! Googling the error I found that 3.15 is needed which I did in fact had to install on my system to do the Hackathon.

kappe-c: I did not mention it ealier because I supposed that everybody just had a newer version than I.  And at least for Jonas (3.17) that is apparently true.

kappe-c: So, looking closer, maybe 3.12 is enough. I just have no possibility to check this because for my Linux distrubtion it is either 3.10 or at least 3.15. 
Maybe set it to Julien's 3.14.6 if this is the oldest that anybody can confirm to work?

pierre-guillou: According to CMake's documentation, `add_compile_definitions` has been introduced in v3.12 (c.f. https://cmake.org/cmake/help/v3.12/command/add_compile_definitions.html).

julien-tierny: How about, to be on the safe side, moving it to the least recent version which has been observed working as christopher suggests? (i.e. 3.14.6)
I'm fine with 3.12 (as the documentation suggests) but I'd prefer to put that after someone tested it to be on the safe side.

kappe-c: Good (late) morning. Since I started the issue, I now installed cmake **3.12**.2 (conda environments can be quite useful after all) and it runs without error. Sorry, next time I'll investigate more closely to begin with.
So I guess for maxium applicability we set this low version requirement. You guys decide how precise it should be (third number (.2) or not? I think this is just the "patch" ID. I got major.minor.2 installed by default).

julien-tierny: I'd say 3.12.2 to be on the safe side (although 3.12 looks sufficient based on the documentation).

PR #435, opened 2020-07-02
[api-migration-hackathon] DimensionReduction

pierre-guillou:
This PR migrates the DimensionReduction module to the new architecture.

For now, the current way of selecting input table arrays has been preserved. As a consequence, the ttk-data state files using DimensionReduction work without modifications.

Enjoy,
Pierre


julien-tierny:
awesome pierre, thanks!

PR #436, opened 2020-07-03
Bug fix: `make install` for ParaView plugins

dsakurai:
- [X] Review our [Contributor Guidelines](https://github.com/topology-tool-kit/ttk/blob/dev/CONTRIBUTING.md), in particular regarding code formatting (with clang-format) and continuous integration.

- [X] Please provide a quick description of your contributions below:

When you issue a `make install` on macOS to install the TTK plugin for ParaView,
`TopologyToolKit.so` has failed to find shared libs in the subdirectory `TopologyToolKit`.
This is because CMake does not set rpaths properly for libraries in the
same project. Following CMake's documentation, this commit sets the
rpaths.

Perhaps an equivalent fix is needed on Linux (and on Windows.)

### DIFF
--- a/paraview/singlePlugin/CMakeLists.txt
+++ b/paraview/singlePlugin/CMakeLists.txt
@@ -17,6 +17,8 @@ paraview_add_plugin(TopologyToolKit
   )
 
 if(NOT "${TTK_INSTALL_PLUGIN_DIR}" STREQUAL "")
+  ttk_set_paraview_rpath(TopologyToolKit)
+
   install(
     TARGETS
       TopologyToolKit
--- a/CMake/ParaViewFilter.cmake
+++ b/CMake/ParaViewFilter.cmake
@@ -33,3 +33,22 @@ macro(ttk_register_pv_filter vtkModuleDir xmlFile)
     endif()
   endif()
 endmacro()
+
+function(ttk_set_paraview_rpath TARGET_NAME)
+  if(APPLE)
+    # On macOS,
+    # look into the subdirectory "TopologyToolKit"
+    # to find the actual plugins.
+    get_target_property(TEMP
+        ${TARGET_NAME} INSTALL_RPATH
+        )
+    set_target_properties(${TARGET_NAME}
+      PROPERTIES
+        INSTALL_RPATH "@loader_path/${TTK_PLUGIN_SUBDIR};${TEMP}"
+    )
+    get_target_property(TEMPTWO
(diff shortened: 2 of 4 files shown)


julien-tierny:
thanks daisuke for the PR!
@CharlesGueunet can you double-check really quick if everything's OK for you?
cheers

CharlesGueunet:
It seems right to me. @dsakurai would it be possible to move the method declaration in `CMake/VTKModule.cmake` and `CMake/ParaViewFilter.cmake` ? I would prefer the `config.cmake` to stay free of method declaration.

dsakurai:
Done.
[See the change summary](https://github.com/topology-tool-kit/ttk/pull/436/files#diff-5c802d742761ffc52e8fe073a351cab4).

@CharlesGueunet you meant like this?

paraview/singlePlugin/CMakeLists.txt
```diff
@@ -17,6 +17,8 @@ paraview_add_plugin(TopologyToolKit
   )
 
 if(NOT "${TTK_INSTALL_PLUGIN_DIR}" STREQUAL "")
+  ttk_set_paraview_rpath(TopologyToolKit)
```
vollia, line 20: Isn't this the `${TTK_PLUGIN_SUBDIR}` ?

CharlesGueunet:
Yes, thanks ! One last minor remarks and it is all right for me :)

julien-tierny:
@dsakurai0 @CharlesGueunet 
are charles' latest concerns addressed?
cheers,

paraview/singlePlugin/CMakeLists.txt
dsakurai: Actually it's not. That's the CMake target for `TopologyToolKit.so`, which happens to be `TopologyToolKit`.

`${TTK_PLUGIN_SUBDIR}` is a subdirectory under `TTK_INSTALL_PLUGIN_DIR` which is also named `TopologyToolKit`. This subdirectory contains the libttk*.so files that have the actual computation code.

dsakurai:
@julien-tierny 
> are charles' latest concerns addressed?

Yes, from my point of view. (Reply added to Charles' question.)
Waiting for @CharlesGueunet's confirmation.

CharlesGueunet:
Right, this is good for me !

julien-tierny:
alright, thanks a lot!

PR #437, opened 2020-07-03
[api-migration-hackathon] TopologicalCompression

pierre-guillou:
This PR migrates the TopologicalCompression module to the new architecture:

- the new logging API replaces the old one (and the `std::cerr <<` and the `fprintf`),
- the triangulation-using methods have been templated,
- the new `GetInputArrayToProcess`, `SetInputArrayToProcess` method to access arrays has been used,
- TopologicalCompressionReader, TopologicalCompressionWriter and TopologicalSimplification have been updated to fix compatibility issues.

Since no ttk-data filter uses TopologicalCompression (excluding persistenceDrivenCompression that uses TopologicalCompressionReader), no need to update the state files.

Enjoy,
Pierre

### DIFF
--- a/core/vtk/ttkTopologicalCompressionReader/ttkTopologicalCompressionReader.cpp
+++ b/core/vtk/ttkTopologicalCompressionReader/ttkTopologicalCompressionReader.cpp
@@ -134,7 +134,7 @@ int ttkTopologicalCompressionReader::RequestData(
   // decompressed->SetVoidArray(, vertexNumber, 0);
   mesh->GetPointData()->AddArray(decompressed);
 
-  if(SQMethod != 1 && SQMethod != 2 && !ZFPOnly) {
+  if(SQMethodInt != 1 && SQMethodInt != 2 && !ZFPOnly) {
     vtkNew<vtkIntArray> vertexOffset{};
     vertexOffset->SetNumberOfTuples(vertexNumber);
     vertexOffset->SetName(ttk::OffsetScalarFieldName);
--- a/core/vtk/ttkTopologicalCompressionReader/ttkTopologicalCompressionReader.cpp
+++ b/core/vtk/ttkTopologicalCompressionReader/ttkTopologicalCompressionReader.cpp
@@ -115,7 +115,8 @@ int ttkTopologicalCompressionReader::RequestData(
   topologicalCompression.preconditionTriangulation(
     triangulation.getTriangulation());
 
-  const auto status = topologicalCompression.ReadFromFile<double>(fp);
+  const auto status = topologicalCompression.ReadFromFile<double>(
+    fp, *(triangulation.getTriangulation()));
   if(status != 0) {
     vtkWarningMacro("Failure when reading compressed TTK file");
   }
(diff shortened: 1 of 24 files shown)


core/vtk/ttkTopologicalCompressionWriter/ttkTopologicalCompressionWriter.cpp
```diff
@@ -75,19 +77,19 @@ int ttkTopologicalCompressionWriter::AllocateOutput(
...
     UseTopologicalSimplification);
   topologicalCompression.setZFPOnly(ZFPOnly);
   topologicalCompression.setCompressionType(CompressionType);
-  topologicalCompression.setOutputDataPointer(
-    outputScalarField->GetVoidPointer(0));
   topologicalCompression.setMaximumError(MaximumError);
+  topologicalCompression.setTolerance(Tolerance);
   switch(inputScalarField->GetDataType()) {
-    vtkTemplateMacro(topologicalCompression.execute<VTK_TT>(Tolerance));
+    vtkTemplateMacro(topologicalCompression.execute(
```
julien-tierny: why not templating on the triangulation type here?
(and then call ```triangulation.getTriangulation().getData()``` instead)?

core/vtk/ttkTopologicalCompressionReader/ttkTopologicalCompressionReader.cpp
```diff
@@ -112,9 +112,11 @@ int ttkTopologicalCompressionReader::RequestData(
   BuildMesh();
 
   triangulation.setInputData(mesh);
-  topologicalCompression.setupTriangulation(triangulation.getTriangulation());
+  topologicalCompression.preconditionTriangulation(
+    triangulation.getTriangulation());
 
-  const auto status = topologicalCompression.ReadFromFile<double>(fp);
+  const auto status = topologicalCompression.ReadFromFile<double>(
+    fp, *(triangulation.getTriangulation()));
```
julien-tierny: this should be templated by triangulation type.

julien-tierny:
hi pierre,

thanks a lot for your PR!
I added a couple of remarks regarding the reader and the writer. that would be wonderful if you could take care of those.
also, have you run this PR against ttk-data? (persistenceDrivenCompression)

cheers,

pierre-guillou:
Sorry, I thought only the TopologicalCompression module was in the scope of the migration. I'll modify TopologicalCompressionReader and TopologicalCompressionWriter to better fit the new APIs.

pierre-guillou:
Now TopologicalCompressionReader and TopologicalCompressionWriter should:

- use concrete triangulation types when calling the base layer,
- depend on ttkAlgorithm instead of ttkTriangulationAlgorithm,
- use the new log API,
- inherit from the base layer and use its class member directly,
- use the Get|SetInputArrayToProcess API to fetch the scalar array to process.

In particular, TopologicalCompressionWriter no longer inherits from vtkWriter but redefines some of its methods. This should have no impact on the actual use of this writer, it is still accessible in ParaView's Save Data menu.

In a similar manner, TopologicalCompressionReader no longer inherits from vtkImageAlgorithm and similar techniques (duck-typing) were used to enforce compatibility. The state file persistenceDrivenCompression.pvsm remains unchanged by these modifications (maybe more log appears).

julien-tierny:
alright, looks good to me!

PR #440, opened 2020-07-03
Precondition caching for periodic triangulation

pierre-guillou:
This PR applies the changes made in #350 to the PeriodicImplicitTriangulation. Vertices, edges, triangles and tetrahedron "position" is cached at the precondition phase to fasten latter queries.

Enjoy,
Pierre


julien-tierny:
looks good. thanks!

PR #442, opened 2020-07-07
DimensionReduction: Add warning when MDS is used with multiple threads

pierre-guillou:
This PR adds a warning in DimensionReduction when MDS is used with multiple threads.

Enjoy,
Pierre


julien-tierny:
awesome, thanks!

PR #443, opened 2020-07-08
ImportEmbeddingFromTable Migration

floWetzels:
Removed "ttk" from console message prefix and removed needsToAbort and updateProgress functions.


julien-tierny:
cool, thanks!

PR #445, opened 2020-07-09
PersistenceDiagramDistanceMatrix: Support rectangular matrices

pierre-guillou:
This PR adds the support of rectangular matrices alongside square matrices in the PersistenceDiagramDistanceMatrix module. Square matrices are still computed in a triangular fashion.

Enjoy,
Pierre


julien-tierny:
looks good to me, thanks!

PR #446, opened 2020-07-09
[api-migration-hackathon] OBJWriter, OFFWriter, OFFReader

pierre-guillou:
This PR migrates the OBJWriter, the OFFWriter and the OFFReader modules to the new log API (with a new dependency to the common library).

The modules were also cleaned: some class members have been demoted to local variables, some useless includes have been removed.

No change has been observed to the written OBJ and OFF files.

Enjoy,
Pierre


julien-tierny:
looks good, thanks!

PR #448, opened 2020-07-09
ImportEmbeddingFromTable : Remove InputArray requirement

### LINKED ISSUE #447: ttkImportEmbeddingFromTable should accept dataless points
**Describe the bug**
the TTK ImportEmbeddingFromTable filter doesn't accept unstructured grids that doesn't have a point-defined data array.

**To Reproduce**
1 Open Paraview
2 Generate your favorite Unstructured Grid data, without any data  defined on points
3 See that you can't apply the TTK ImportEmbeddingFromTable filter
4 Add any data to points (with Elevation for exemple)
5 See that you can apply the filter (if you have an available table)

**Expected behavior**
Adding data to points shouldn't be required to apply the filter.

**Fix**
Remove the following lines from the XML descriptor file for paraview : 
```
        <InputArrayDomain name="input_scalars"
                          attribute_type="point">
          <Property name="DataSet" function="FieldDataSelection" />
        </InputArrayDomain>
```

hugo-manet:
Fix #447 
Remove useless requirement in paraview XML file

### DIFF
--- a/paraview/ImportEmbeddingFromTable/ImportEmbeddingFromTable.xml
+++ b/paraview/ImportEmbeddingFromTable/ImportEmbeddingFromTable.xml
@@ -26,10 +26,6 @@
         <DataTypeDomain name="input_type">
           <DataType value="vtkPointSet"/>
         </DataTypeDomain>
-        <InputArrayDomain name="input_scalars"
-                          attribute_type="point">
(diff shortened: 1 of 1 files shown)


julien-tierny:
cool, thanks!

PR #450, opened 2020-07-10
Add missing depends and fix RPATH

mcmigliore:
This PR adds two missing depends in modules.  
Note that if the modules are not present, TTK will not configure. This is something that should be fixed but not related to this PR.

It also adds tests to avoid overwriting user set variables related to RPATH.


mcmigliore:
@julien-tierny @CharlesGueunet please review

config.cmake
```diff
@@ -317,9 +317,18 @@ endif()
 set(TTK_PLUGIN_SUBDIR "TopologyToolKit")
 
 # Install rpath
-
-set(CMAKE_MACOSX_RPATH TRUE)
-set(CMAKE_INSTALL_RPATH TRUE)
-set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
-set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib/ttk/")
-set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+if(NOT DEFINED CMAKE_MACOSX_RPATH)
```
vollia, line 320: why not using `option` for boolean entries ?

CharlesGueunet (approved):

julien-tierny:
looks good for me. just waiting for the ci to do its job.
thanks a lot michael!

PR #451, opened 2020-07-13
CMake: automatically disable module when dependencies are not met

CharlesGueunet:
Dear Julien,

This PR improve the build of TTK by disabling modules (with a warning) when ther VTK / ParaView dependencies are not met. This allows to install TTK using specific ParaView / VTK more convenient.
For example, we can now compile TTK agains a Catalyst edition of ParaView without having to specify a long list of plugin to disable on the command line.

@mcmigliore FYI

Charles


julien-tierny:
looks good. thanks!

PR #452, opened 2020-07-13
[api-migration-hackathon] MandatoryCriticalPoints migration

pierre-guillou:
This PR migrates the MandatoryCriticalPoints module:

- the new log API is used,
- the new architecture is in place,
- arrays are now set with `Set|GetInputArrayToProcess`,
- the standalone has also been migrated.

Due to compatibility issues, two statements in the ContourTree module have been temporary commented (AbstractTriangulation is always empty…).

The LowestCommonAncestor and RangeMinimumQuery module have also been updated.

Corresponding ttk-data PR: https://github.com/topology-tool-kit/ttk-data/pull/48

Enjoy,
Pierre


core/vtk/ttkMandatoryCriticalPoints/vtk.module
```diff
@@ -1,6 +1,4 @@
 NAME
  ttkMandatoryCriticalPoints
 DEPENDS
-  VTK::FiltersCore
-PRIVATE_DEPENDS
-  VTK::CommonCore
+ ttkAlgorithm
```
vollia, line 4: @julien-tierny this change is what I am mentioning in #451 
@pierre-guillou do not worry, if this PR is merged I can rebase mine and fix this on my side

vollia: Never-mind, we can keep the mechanism with `ttkAlgorithm`.

julien-tierny:
awesome, thanks!

julien-tierny:
@pierre-guillou 
could you also migrate the logging for core/base/contourTree as well? 
(the legacy one is still polluting the console output).

julien-tierny:
plus there seems to be some double-logging in core/base/lowestCommonAncestor

pierre-guillou:
> @pierre-guillou
> could you also migrate the logging for core/base/contourTree as well?
> (the legacy one is still polluting the console output).

Will do. I thought it was a dependency of FTM and/or ContourForests and that Charles or Jonas would deal with this module.

> plus there seems to be some double-logging in core/base/lowestCommonAncestor

There is one LowestCommonAncestor method that is called twice by MandatoryCriticalPoints (preprocess). I reduced the warning level in this method so it won't appear by default.

PR #453, opened 2020-07-15
Remove persistenceDiagram dependency when unused

pierre-guillou:
Several TTK base modules declare persistenceDiagram as a dependency although they don't use it in practice.

This PR removes these unused dependencies. This should lead to faster build time, given that for the corresponding compilation units, the compiler doesn't have to parse the PersistenceDiagram header (and the included MorseSmaleComplex/DiscreteGradient headers).

No change has been observed in the ttk-data states that could have been impacted.

Enjoy,
Pierre


julien-tierny:
looks good to me. thanks!

PR #456, opened 2020-07-17
[api-migration-hackathon] LDistance

pierre-guillou:
This PR migrates the LDistance module to the new API/Architecture.

No ttk-data states are impacted by this modification.

Enjoy,
Pierre


julien-tierny:
looks good, thanks!

PR #460, opened 2020-07-17
[api-migration-hackathon] PointDataSelector

pierre-guillou:
This PR migrates the PointDataSelector module to the new API/Architecture.

As the scalar field selection system has been left untouched, no ttk-data states are impacted by this modification.

Enjoy,
Pierre


julien-tierny:
awesome, thanks!

PR #461, opened 2020-07-17
[api-migration-hackathon] UncertainDataEstimator

pierre-guillou:
This PR migrates the UncertainDataEstimator module to the new API/Architecture.

No ttk-data states are impacted by this modification.

Enjoy,
Pierre


julien-tierny:
ok, thanks!

PR #463, opened 2020-07-22
[api-migration-hackathon] WRLExporter

pierre-guillou:
This PR migrates the WRLExporter module to the new log API.

No state in ttk-data should be impacted.

Enjoy,
Pierre


julien-tierny:
looks good, thanks

PR #464, opened 2020-07-22
[api-migration-hackathon] ContourTree

pierre-guillou:
This PR migrates the ContourTree module to the new log API.

I also converted the line endings from DOS (\r\n) to Unix (\n) to be consistent with the rest of the codebase.

No ttk-data state should be impacted.

Enjoy,
Pierre


julien-tierny:
git's diff seems to be broken on this one (it reports complete file changes), but that looks ok overall.

pierre-guillou:
> git's diff seems to be broken on this one (it reports complete file changes)

It's related to the line endings changes that I performed with `dos2unix`.

PR #465, opened 2020-07-22
[api-migration-hackathon] ReebSpace, FiberSurface, JacobiSet

pierre-guillou:
This PR migrates the ReebSpace, the FiberSurface and the JacobiSet modules to the new architecture/log API.

The JacobiSet class has been de-templated.

To reduce build times, a new CMake option has been introduced: `TTK_AVOID_DOUBLE_TEMPLATING` (ON by default...). This option enforce a single templating of these modules, returning an error if the two input scalar fields don't have the same type.

Corresponding ttk-data PR: https://github.com/topology-tool-kit/ttk-data/pull/51

Enjoy,
Pierre


core/vtk/ttkJacobiSet/ttkJacobiSet.h
```diff
@@ -19,6 +19,24 @@
 /// This filter can be used as any other VTK filter (for instance, by using the
 /// sequence of calls SetInputData(), Update(), GetOutput()).
 ///
+/// The input data arrays needs to be specified via the standard VTK call
```
julien-tierny, line 22: I'll go ahead and merge that PR, but would it be possible to have exactly the same documentation for core/vtk/ttkReebSpace/ttkReebSpace.h?

julien-tierny:
thank you very much pierre, this is great.
if you could add that doxygent bit for the Reeb space, that'd be awesome.

core/vtk/ttkJacobiSet/ttkJacobiSet.h
pierre-guillou: I'll do that

PR #466, opened 2020-07-22
Harden linker flags

pierre-guillou:
Since I began to play with explicit template instantiations (in particular in MorseSmaleComplex) to factor compile times and in order to eliminate `symbol lookup` run-time errors, this PR adds these two linker flags:

- `no-undefined` will return error at link time when symbols are missing, typically when the VTK layer calls a missing explicit template instantiation in the base layer,
- `as-needed` should reduce the linked library size by only linking to what is currently used.

At this moment, these flags are only applicable for GCC (ld.bfd and ld.gold) and Clang (ld.lld). I am not (yet) aware of similar flags for the MSVC linker, thus the `if(not MSVC)` everywhere.

Enjoy,
Pierre

### DIFF
--- a/CMake/BaseCode.cmake
+++ b/CMake/BaseCode.cmake
@@ -36,7 +36,7 @@ function(ttk_add_base_library library)
       endif()
     endforeach()
   endif()
- 
+
   ttk_set_compile_options(${library})
 
   install(TARGETS ${library}
@@ -130,6 +130,11 @@ function(ttk_set_compile_options library)
 
   # compilation flags
   target_compile_options(${library} PRIVATE ${TTK_COMPILER_FLAGS})
+
+  # linker flags
+  if(NOT MSVC)
+    target_link_options(${library} PRIVATE ${TTK_LINKER_FLAGS})
+  endif()
 
   if (TTK_ENABLE_KAMIKAZE)
     target_compile_definitions(${library} PUBLIC TTK_ENABLE_KAMIKAZE)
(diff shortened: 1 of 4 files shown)


lhofmann:
FYI, in certain configurations VTK will intentionally produce undefined symbols when linking against Python, so linking the VTK modules with `--no-undefined` would fail.
See the initial discussion here (this feature is included in VTK 9): https://discourse.vtk.org/t/segfault-for-python-importing-vtk-but-not-for-vtkpython/126

pierre-guillou:
Since the macOS linker doesn't seem to support `--as-needed` (they don't use lld?), I put this PR into draft mode. I plan to only enable those flags on Linux.

@lhofmann I came across the `VTK_PYTHON_OPTIONAL_LINK` variable when I got some undefined symbols from Python. Since there was only one module (dimensionReduction) that needed those symbols, I forced the link to `libpython`. Do you think this could cause further problems?

lhofmann:
@pierre-guillou I only know for certain, that the conda build on OSX would be affected. However, in that case I would simply patch out these linker flags and skip linking against libpython during the package build (as I am currently doing for the VTK 8 version).

pierre-guillou:
For now, I restricted these flags to Linux only, the conda builds on macOS shouldn't be impacted. I think it's good to have these flags on at least one platform, as a safeguard against run-time lookup errors.

julien-tierny:
looks good, thanks!

PR #467, opened 2020-07-22
Add ttk prefix to paraview proxy names

### LINKED ISSUE #454: ParaView proxy names should have a ttk prefix
Currently, all XML files contain declarations like this:
```xml
<SourceProxy name="ArrayEditor" class="ttkArrayEditor" label="TTK ArrayEditor">
```
Because proxy names are unique within ParaView, I have run into duplicate proxy names before, since I had similar naming schemes in my own code base. Especially the more generic ones like ArrayEditor, Extract, etc. could easily cause conflicts elsewhere.
These names are as far as I know only used internally in ParaView and in the state files. Since the pvpython interface uses the labels as Python class names, all pvpython scripts will be unaffected.

I propose to change all declarations such that the name is equal to the class name wherever possible:
```xml
<SourceProxy name="ttkArrayEditor" class="ttkArrayEditor" label="TTK ArrayEditor">
```
Compatibility with state files (especially those existing outside of ttk-data) could be maintained by adding aliases like this:
```xml
<SourceProxy base_proxygroup="filters"
             base_proxyname="ttkArrayEditor"
             class="ttkArrayEditor"
             name="ArrayEditor"
             label="TTK ArrayEditor (deprecated)">
    <Deprecated deprecated_in="5.8" to_remove_in="5.9">
        Please update your state file to use ttkArrayEditor instead of ArrayEditor.
    </Deprecated>
</SourceProxy> 
```
These will not appear in the filters menu, and a deprecation message is printed when used (e.g. loaded from an old state file).

lhofmann:
This PR appends a `ttk` prefix to all ParaView proxy names (closes #454).

* The renaming has been done by the script `scripts/rename_plugins.py`.
This script can also convert arbitrary state files if called with a list of .pvsm files as command line parameter.
* There is a compatibility plugin (`paraview/Compatibility/Compatibility.xml`) generated, that allows old state files to still be loaded while displaying a warning message.

* I'm not sure if the compatibility plugin should be included in TTK by default, since for each "TTK Xyz" filter, an entry "TTK Xyz (deprecated)" will appear in the Ctrl+Space menu (but not in the main menus). Alternatively, the XML file `Compatibility.xml` could also be loaded as a plugin by the user directly without building a plugin.

### DIFF
--- a/paraview/ArrayEditor/ArrayEditor.xml
+++ b/paraview/ArrayEditor/ArrayEditor.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <ServerManagerConfiguration>
     <ProxyGroup name="filters">
-        <SourceProxy name="ArrayEditor" class="ttkArrayEditor" label="TTK ArrayEditor">
+        <SourceProxy name="ttkArrayEditor" class="ttkArrayEditor" label="TTK ArrayEditor">
             <Documentation long_help="TTK ArrayEditor" short_help="TTK ArrayEditor">This filter adds data arrays to a 'vtkDataObject' (called target) based on a string or point/cell/field data of an optional second 'vtkDataObject' (called source). This filter can also be used to directly edit an array (including renaming, type conversion, and reindexing).</Documentation>
             <!-- Inputs -->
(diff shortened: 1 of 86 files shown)


julien-tierny:
looks pretty neat lutz, thanks a lot!

PR #469, opened 2020-07-24
Move numberOfInput_ to inside PersDiagClustering::execute(), remove setter

hugo-manet:
Hello !
This numberOfInput_ class member is:
- obsolete: the code implies that it should necessarily be equal to intermediateDiagram.size()
- dangerous: the code segfaults if it is smaller
- preventing the execute() function from being reentrant: other class members are most likely set only once, or at least won't make the code explode if modified at any point

I suggest we remove it.

Other cleanings could be done (points_added_ and points_deleted_ are never used) but this isn't critical.


julien-tierny:
thanks hugo.
@pierre-guillou @julesvidal can you double check this is ok for you?

julesvidal:
Hey, it looks fine !
Edit: The points_added_  and points_deleted_ members were probably used for logging purpose in the first place, they can be safely removed

julien-tierny:
looks good, thanks!

PR #470, opened 2020-07-24
[api-migration-hackathon] ContourForests

pierre-guillou:
This PR migrates the ContourForests module to the new log API (partially, only for the default debugLevel_), to the new architecture and the new Get|SetInputArrayToProcess API.

No ttk-data state file is using this filter.

Enjoy,
Pierre


julien-tierny:
thanks!

PR #471, opened 2020-07-28
PersistenceDiagram: Fix diagrams not recomputing

pierre-guillou:
The current PR fixes a bug in the PersistenceDiagram module in which changing the input scalar field does not trigger a recomputation. Incidentally, modifying the number of threads and the log level should also re-trigger the computation, contrary to the ShowInsideDomain option.

Enjoy,
Pierre


julien-tierny:
awesome, thanks!

PR #472, opened 2020-07-28
Fix diagram size in input of diagrams diagonal is present or not

hugo-manet:
Fix a bug in persistence diagram inputs:
* Checks whether the diagonal (assumed to be the last cell, with `PairID == -1`) exists or not, and thus remove it or not from the input
* Fix the diagram size. Before the fix, the diagram pair of index `pairingsSize - 1` was never initialized, only created through the `vector::resize()`


hugo-manet:
Also, both functions duplicate what they believe is the (min,max) pair (the one with `pairID == 0`). `ttkPersistenceDiagramClustering` does this if it is doing a clustering (ie if `NumberOfCluster > 1`), and `ttkPersistenceDiagramDistanceMatrix` always does it:
* this seems like a hotfix for distance problems, that might no longer be needed ; at least it should be an option that can be disabled
* clustering results don't output the (min,max) pair in first position, so this code can duplicate the wrong pair and thus yield absurd distances

I suggest we do something about it too.

julien-tierny:
ok.

PR #473, opened 2020-07-28
[api-migration-hackathon] TopologicalSimplification

pierre-guillou:
This PR migrates the TopologicalSimplification module to the new API/architecture.

Note that the work on sorting offset field should arrive later, since it depends on the (pending) migration of the FTMTree and FTRGraph modules (ttk-data states break without a full support of this new feature).

Corresponding ttk-data PR: https://github.com/topology-tool-kit/ttk-data/pull/53

Enjoy,
Pierre


julien-tierny:
that's awesome. thanks a lot pierre.

PR #409, opened 2020-06-30
[api-migration-hackathon]  WIP: FTM Tree module migration

CharlesGueunet:
Dear Julien,

This PR update the FTM Tree module to the new TTK API.
It is still a work in progress as I need to test that everything works more extensively.

Charles

PR #474, opened 2020-07-30
[api-migration-hackathon] FTMTree

pierre-guillou:
This PR migrates the FTMTree module to the new architecture. This PR is based on and should supersede #409 .

PersistenceDiagram and PersistenceCurve were also modified (some migration bits were depending on this module).

Since no ttk-data state uses FTM directly (only through PersistenceDiagram), no need to modify them.

Enjoy,
Pierre


pierre-guillou:
In fact, there were some ttk-data states using the FTMTree module (under the lowercase name `ttkftmtree`). The corresponding PR is https://github.com/topology-tool-kit/ttk-data/pull/55.

julien-tierny:
alright, let's keep our fingers crossed.

PR #475, opened 2020-07-30
[api-migration-hackathon] ftrGraph

pierre-guillou:
This PR migrates the ftrGraph module to the new architecture.

Due to the complex nature of the code (and my lack of prior knowledge about it), base classes were left templated (by the scalar type and the triangulation type). The de-templatization (and the passing of the input scalar array and the triangulation object to every methods using them) is left as an exercise.

Corresponding ttk-data PR: https://github.com/topology-tool-kit/ttk-data/pull/54

Enjoy,
Pierre


julien-tierny:
looks good, thanks!

PR #478, opened 2020-08-31
MatrixToHeatMap: Parallelize filling the grid

pierre-guillou:
This PR parallelizes filling the output grid in the MatrixToHeatMap filter, leading to an at least x5 speedup with large input distance matrices.

Enjoy,
Pierre


julien-tierny:
awesome, thanks!

Issue #477, opened 2020-08-29
TTK + Python: Calls to ttkBottleneckDistance Getresult() return -1


josiahblaisdell:
**Fix**
I have fixed this in my own code. I just passed the result by reference into the computeDistance() function and set the value to the distance. 

**Describe the bug**
I created 2 persistence diagrams and passed both of these diagrams and used the ttkBottleneckDistance() to compute the distance between the persistence diagrams. There is a member variable "result" and a vtk macro to get the result in ttkBottleneckDistance.h. After computing the bottleneck distance between two persistence diagrams this "result" is still -1. Not the result of the bottleneck distance algorithm 

**To Reproduce**
```
    #1. Get the names of the variables in the MRI data
    name1 = vtk1.GetScalarsNameInFile(0)
    name2 = vtk2.GetScalarsNameInFile(0)

    #2. threshold fata greater than 0
    e1_threshold = vtkThreshold()
    e1_threshold.SetInputConnection(vtk1.GetOutputPort())
    e1_threshold.SetInputArrayToProcess(0,0,0,0,name1)
    e1_threshold.ThresholdBetween(0.1, 999999)
    e1_threshold.Update()
    pdb.set_trace()

    #3. Tetrahedralize the data in order to create persistence diagram
    tet1 = vtkDataSetTriangleFilter()
    tet1.SetInputConnection(e1_threshold.GetOutputPort())
    tet1.SetInputArrayToProcess(0,0,0,0,name1)
    tet1.Update()

    #4. Create the persistence diagram
    persistence_dgm1 = ttkPersistenceDiagram()
    persistence_dgm1.SetInputConnection(tet1.GetOutputPort())
    persistence_dgm1.SetInputArrayToProcess(0,0,0,0,name1)
    persistence_dgm1.Update()

    #5. Threshold out the persistence diagram main diagonal
    threshold1 = vtkThreshold()
    threshold1.SetInputConnection(persistence_dgm1.GetOutputPort())
    threshold1.SetInputArrayToProcess(0,0,0,vtkDataObject.FIELD_ASSOCIATION_CELLS,'Persistence')
    threshold1.ThresholdBetween(.01,9999999)
    threshold1.Update()

    #6 Threshold the data greater than 0
    e2_threshold = vtkThreshold()
    e2_threshold.SetInputConnection(vtk2.GetOutputPort())
    e2_threshold.SetInputArrayToProcess(0,0,0,0,name2)
[...]

PR #479, opened 2020-08-31
After migration fixes

pierre-guillou:
This PR fixes a lot of small bugs and compiler warnings, mostly introduced during the migration hackathon, including:

- using CMake `find_package` for the Spectra dependency,
- MSVC compiler fixes (un-initialized variables in Auction, default template parameters in DiscreteGradient),
- GCC -Wnull-dereference warning fixes in the VTK layer with new null-pointers tests,
- a lot of typos corrections,
- further cleaning in the DistanceField, the FTRGraph and the LowestCommonAncestor modules,
- a fix for #477 about the BottleneckDistance return value,
- using the correct dependencies and the `ttkUtils::GetVoidPointer` method, replacing `vtkSmartPointer` with `vtkNew`.

No change has been observed while running all the ttk-data states.

Enjoy,
Pierre


julien-tierny:
ok, looks good. thanks!

PR #481, opened 2020-09-01
Support sorted offset fields

pierre-guillou:
This PR add the support of sorted offset fields. Sorted offset fields are now associated with scalar fields and stored in vtkDataSets as vtkDataArrays with the name $ScalarFieldName + "_Order". These offset fields are created when needed or fetched if already existing.

The creating step sorts the vertices using the scalar field values and the implicit vertex order or a pre-existing offset field.

The new vtkDataArrays are attached to the input of the current filter (see core/vtk/ttkAlgorithm/ttkAlgorithm.cpp). The sorting process is located in core/base/common/OrderDisambiguation.h.

Several modules have been ported to take fully advantage of sorted offsets: DiscreteGradient, MorseSmaleComplex, ScalarFieldCriticalPoints, TopologicalSimplification. The other should be ready in a few days.

Enjoy,
Pierre

### DIFF
--- a/core/base/abstractMorseSmaleComplex/AbstractMorseSmaleComplex.h
+++ b/core/base/abstractMorseSmaleComplex/AbstractMorseSmaleComplex.h
@@ -20,7 +20,19 @@
 
 #include <queue>
 
+#if defined(__GNUC__) && !defined(__clang__)
+#include <parallel/algorithm>
+#endif
+
 namespace ttk {
+
+#if defined(_GLIBCXX_PARALLEL_FEATURES_H) && defined(TTK_ENABLE_OPENMP)
+#define PSORT                               \
+  omp_set_num_threads(this->threadNumber_); \
+  __gnu_parallel::sort
+#else
+#define PSORT std::sort
+#endif // _GLIBCXX_PARALLEL_FEATURES_H && TTK_ENABLE_OPENMP
 
   /**
    * Utility class representing Ridge lines, Valley lines
--- a/core/base/common/CMakeLists.txt
+++ b/core/base/common/CMakeLists.txt
@@ -8,6 +8,7 @@ ttk_add_base_library(common
         CommandLineParser.h
         Debug.h
         DataTypes.h
+        OrderDisambiguation.h
         Os.h
         ProgramBase.h
         Wrapper.h
--- a/core/base/common/OrderDisambiguation.h
+++ b/core/base/common/OrderDisambiguation.h
@@ -1,0 +1,69 @@
+#include <DataTypes.h>
+
+#include <algorithm>
+#include <vector>
+
+#if defined(__GNUC__) && !defined(__clang__)
+#include <parallel/algorithm>
+#endif
+
+namespace ttk {
+
+  /**
+   * @brief Sort vertices according to scalars disambiguated by offsets
+   *
+   * @param[in] nVerts number of vertices
+   * @param[in] scalars array of size nVerts, main vertex comparator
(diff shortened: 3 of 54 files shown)


julien-tierny:
@JonasLukasczyk thanks for letting us know when you're done with your pass.
cheers,

JonasLukasczyk:
I'm debugging my changes right now, but I encountered something odd:
I'm able to create (if necessary) an order field and attach it to the input. The algorithm can execute and everything is fine. But if I reexecute the filter the order field is gone and ttkAlgorithm will create a new one.

@pierre-guillou: can you verify that your commit permanently adds an order field (so the next time a filter is executed it just fetches the array instead of creating a new one?).

It seems to me that VTK might have put in some mechanism to prevent us from permanently changing the input (since we should not do that in the first place).

pierre-guillou:
I just checked and I also got what you described: if I modify the options of a filter that generates an order field, the order field is regenerated a second time. The order array doesn't appear in the previous filter list of arrays. Although we modify the filter input, the previous filter's output stays unmodified.

JonasLukasczyk:
I did various tests and I ended up with the same result. I was aware that the PV UI will not show the added array in the previous filter (because the UI is not aware that something got added to the input filter), but "at least" in the past the input object was actually changed under the hood.

So now that this doesn't work anymore I guess we should have another meeting and discuss alternatives. I'm still free today if you have time.

julien-tierny:
alternatives include:
* creating a simulation of simplicity module (in charge of the pre-sorting and attaching the order field)
  * downside: several pvsm files in ttk-data will require important updates
* adding the order field to the output when possible (that covers only a fraction of the use cases)
* implement an internal indexing mechanism similar to the current triangulation management (but saving/loading order fields may be a pain).

I'm available at 11am if you guys want to discuss this.
cheers

julien-tierny:
hi @JonasLukasczyk 
it seems that this PR is currently stuck.
do you need any input from us at this point?
cheers,

JonasLukasczyk:
I'm just busy with VIS but I will push something today.

julien-tierny:
makes sense :) 
thanks!

JonasLukasczyk:
Hi @pierre-guillou, I don't have permission to push on this branch. Can you please update my privileges?

pierre-guillou:
Hi Jonas, I gave you access to my repository but I can see that the invitation is still pending. It seems that you have to accept it in order to have full access.

JonasLukasczyk:
Hi Pierre,
can you resolve the conflict and also run a quick test?

For now GetOffsetArray returns an existing order array, or it creates one temporarily just based on the corresponding scalars (ignoring "old" offsets that might have been stored in a file) with the hint that you should run a preconditioning filter. I guess topological compression writer needs some custom logic to incorporate old offsets.

pierre-guillou:
Hi Jonas,
I just rebased the current branch on top of dev.

I am not totally convinced by your handling of the `enforceOrderArrayIdx` parameter. In my mind, it was to select a specific, 'old' offset array that served to disambiguate flat plateaux (the third parameter of `ttk::sortVertices` that you replaced with a nullptr). In the current state, some ttk-data examples that set the `enforceOrderArrayIdx` boolean to true fail with a segfault (at least `BuildInExample3.pvsm`).

Three solutions:
- either remove this option and sort everything according to the implicit vertex order on plateaux,
- go back to the previous behavior,
- fix the ttk-data states.

julien-tierny:
ok, thanks a lot!

PR #480, opened 2020-08-31
Fix c++ example (and PersistenceCurve API)

pierre-guillou:
This PR fixes the c++ example by using the new API and the correct after-migration calls to the base layer modules.

The PersistenceCurve base layer API has been slightly rolled-back to allow the passing of optional arguments (JTPlot, STPlot and CTPlot). The new API is now more in line with how it is used in the c++ example.

Enjoy,
Pierre

PR #482, opened 2020-09-01
Fix c++ example (and PersistenceCurve API) 2

pierre-guillou:
This PR should fix the link issue in #480 by removing the explicit template instantiations in morseSmaleComplex2D and morseSmaleComplex3D: the example code was calling a ttk::Triangulation implementation whereas only the implementations with the specific triangulation types were instantiated.

Now, without explicit instantiations, the ttk::Triangulation variant will be generated during the build of the example code.

Also, the PersistenceCurve API has been modified again after review: the wrong output would had been computed in the example.

Enjoy,
Pierre


julien-tierny:
alright, thanks!

PR #485, opened 2020-09-07
ContourTreeAlignment Migration from TUKL repository

floWetzels:
Added new module ContourTreeAlignment, which is already migrated to new API.
The module computes an alignment of contour trees for fuzzy contour tree computation (https://doi.org/10.1111/cgf.13985).
ttk-data pull request will follow soon.


julien-tierny:
hi florian,

thanks a lot for the PR.
the CI fails for MacOS though (core/base/contourTreeAlignment/contourtree.cpp:42). 
could you fix that really quick?

cheers,

julien-tierny:
hi florian, 
thanks for your fixes. the ci is now successful.
could you please make a PR to ttk-data to provide one (nice-looking) test example?
also, if you want people to play with this feature, I'd highly recommend to prepare a tutorial video based on the above example (like these: https://topology-tool-kit.github.io/tutorials.html)
cheers,

floWetzels:
Hi Julien,
PR #57 for ttk-data adds a statefile and an example data set the contour tree alignment.

julien-tierny:
hi florian,

sorry for the latency, a lot happened here lately.
could you please provide a screenshot of the result of the state file, to make sure I can reproduce it correctly here?

thanks!
julien

floWetzels:
Hi Julien,
a screenshot should be attached to this comment. What I just noticed: Should I give the filters better names and do another commit for the ttk-data PR?

Screenshot from 2020-09-17 10-05-56

julien-tierny:
hi florian, 

thanks for the great piece of code.

a couple of things:
- [ ] can you clarify what the class contourtree is doing? does it build a tree out of FTM's output?
- [ ] there's a bunch of deletes in the code and we're trying to stay away from new and delete (some even consider 
their usage as bad practice). in principle, I suspect you can replace all of your new/delete with an appropriate stl container (such as vectors for instance).

thanks for your feedback.
julien

core/base/contourTreeAlignment/ContourTreeAlignment.cpp
```diff
@@ -0,0 +1,1726 @@
...
+  if(!ct->isBinary())
+    return false;
+
+  // compute initial Alignment from initial contourtree
+
+  BinaryTree *t = ct->rootAtMax();
+
+  std::queue<std::pair<BinaryTree *, AlignmentNode *>> q;
+
+  AlignmentNode *currNode = new AlignmentNode();
```
julien-tierny, line 186: please avoid new/delete

core/base/contourTreeAlignment/ContourTreeAlignment.cpp
```diff
@@ -0,0 +1,1726 @@
...
+
+  alignmentRoot = nodes[0];
+
+  return true;
+}
+
+void ttk::ContourTreeAlignment::computeNewAlignmenttree(AlignmentTree *res) {
+
+  for(AlignmentNode *node : nodes) {
+    delete node;
```
julien-tierny, line 503: please avoid new/delete

core/base/contourTreeAlignment/ContourTreeAlignment.cpp
```diff
@@ -0,0 +1,1726 @@
...
+/// aligning two trees
+///=====================================================================================================================
+
+std::pair<float, AlignmentTree *>
+  ttk::ContourTreeAlignment::getAlignmentBinary(BinaryTree *t1,
+                                                BinaryTree *t2) {
+
+  // initialize memoization tables
+
+  float **memT = new float *[t1->size + 1];
```
julien-tierny, line 1016: please avoid new/delete

floWetzels:
Hi Julien,
 
> a couple of things:
> 
>     * [ ]  can you clarify what the class contourtree is doing? does it build a tree out of FTM's output?
Basically, yes. It is just my base layer contour tree class with of important helper functions (tranformation to a rooted tree, branch decomposiion). They are constructed from the arrays passed from the vtk layer, so usually an FTM tree.

>     * [ ]  there's a bunch of deletes in the code and we're trying to stay away from new and delete (some even consider
>       their usage as bad practice). in principle, I suspect you can replace all of your new/delete with an appropriate stl container (such as vectors for instance).
I can remove the new/delete commands if it's necessary but this will take a while. The code will need some restructuring. I have tested my my module for memory leaks (by building a standalone module with which I could run ttk with fsanitize=address), they should all be gone (at least in the code that is used by the vtk layer).

Please let me know if I have to remove them all.

Best regards
Florian

julien-tierny:
hi florian,

thanks for the quick feedback.

> Basically, yes. It is just my base layer contour tree class with of important helper functions (tranformation to a rooted tree, 
branch decomposiion). They are constructed from the arrays passed from the vtk layer, so usually an FTM tree.

alright, sounds good.

> I can remove the new/delete commands if it's necessary but this will take a while

understood. we can go ahead and merge now but that'd be really appreciated if you can come back to that later. in the long run, these aren't good at all. at some point, we'll throw a hackathon to get rid of all the remaining ones in ttk, so that'd be good not to further introduce new ones on top of that :)
we're not in a rush but if that could be taken care of in the next few weeks, that'd be awesome.

cheers

floWetzels:
> understood. we can go ahead and merge now but that'd be really appreciated if you can come back to that later. in the long run, these aren't good at all. at some point, we'll throw a hackathon to get rid of all the remaining ones in ttk, so that'd be good not to further introduce new ones on top of that :)
> we're not in a rush but if that could be taken care of in the next few weeks, that'd be awesome.

 Alright, I can do that. I can do some more cleanup and optimization at the same time. Do I have to do another PR then?

How about the state file? Should I rename the filters?

julien-tierny:
> Alright, I can do that. I can do some more cleanup and optimization at the same time. Do I have to do another PR then?

I think so. I think I'm going to go ahead and merge that one and later you can open a new PR just to remove the new/delete.
would that work?

> How about the state file? Should I rename the filters?

you mean the names in the pipeline browser? if so, I think it's fine as it is (that way, people can readily understand what filters you put together)

floWetzels:
> I think so. I think I'm going to go ahead and merge that one and later you can open a new PR just to remove the new/delete.
> would that work?
Sure, great!

> you mean the names in the pipeline browser? if so, I think it's fine as it is (that way, people can readily understand what filters you put together)
Ok, that makes sense.

Thanks for all the feedback!
Florian

PR #486, opened 2020-09-10
[Order-based SoS] Module migrations

pierre-guillou:
This PR replaces the use of offset fields with order fields in the following filter:
* ContourForests
* DiscreteGradient & MorseSmaleComplex
* FTMTree
* FTRGraph
* IntegralLines
* PersistenceDiagram
* PersistenceCurve
* ScalarFieldCriticalPoints
* TopologicalCompression
* TopologicalSimplification
* TrackingFromFields

In particular, the offset sorting routines in ContourForests, FTMTree and FTRGraph have been remove.

A bug in the order field generated by TopologicalSimplification has also been fixed (wrong initial value).

Due to the fixed type of the order arrays (a buffer of SimplexId), some now unecessary copies/conversions have been remove, particularly in JacobiSet and ReebSpace. In the VTK layer, tests of the offset field type around `vtkTemplateMacro` calls have been removed, leading to a reduction of generated code (and improved build times). Base code has been de-templatized over the offset type, particularly in DiscreteGradient and MorseSmaleComplex.

The base layer C++ example has been updated to the new API calls.

No change has been observed in the ttk-data states.

Enjoy,
Pierre

### DIFF
--- a/core/base/abstractMorseSmaleComplex/AbstractMorseSmaleComplex.h
+++ b/core/base/abstractMorseSmaleComplex/AbstractMorseSmaleComplex.h
@@ -228,7 +228,7 @@ namespace ttk {
      * Set the input offset field associated on the points of the data set
      * (if none, identifiers are used instead).
      */
-    inline int setInputOffsets(const void *const data) {
+    inline int setInputOffsets(const SimplexId *const data) {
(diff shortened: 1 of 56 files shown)


julien-tierny:
ok, looks good, thanks!

PR #487, opened 2020-09-11
Order arrays preconditioning for users of the base layer API

pierre-guillou:
This PR introduces a new function, `preconditionOrderArray`, in `core/base/common/OrderBaseDisambiguation.h` that wraps the `sortedVertices` function. This new functions is for users of the base layer API that don't use the automatic mechanisms available in the VTK layer (`GetOrderArray` and `ArrayPreconditioning`).

Doxygen comments that warn to use this function to precondition order arrays before executing base layer modules have been introduced above the relevant methods. 

The C++ example has been fixed to use this new method.

Enjoy,
Pierre


julien-tierny:
awesome, thanks!

PR #491, opened 2020-09-15
ParaView 5.8.0 patches for generating binary packages

pierre-guillou:
This PR adds patches to the `paraview/patch` folder that enable the creation of binary packages for the following targets:
- Ubuntu 18.04 (bionic)
- Ubuntu 20.04 (focal)
- Windows Server 2019
- macOS 10.15 (Catalina).

c.f. https://docs.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners for a description of the available virtual environments.

From the ParaView 5.8.0 sources, the TTK patching process should set the relevant build options, the packaging meta-data and set the Github Action CI file.

Packages should be automatically created when the patched ParaView sources are pushed to Github with tags that match the patterns `v*` or `dev*`. A Github release should be automatically created with links to the generated packages.

Enjoy,
Pierre

### DIFF
--- a/paraview/patch/main.yml
+++ b/paraview/patch/main.yml
@@ -1,0 +1,186 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branch-ignore:
+      - '*'
+    tags:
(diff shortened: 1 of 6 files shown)


julien-tierny:
looks good, thanks!
can you now PR the yml file with the ttk binary packaging?

PR #492, opened 2020-09-17
HelloWorld Fix

JonasLukasczyk:
With the order array commit we accidentally added some debug msgs in the HelloWorld module, which are removed with this PR. This PR also polishes the comments in the HelloWorld module.


julien-tierny:
awesome, thanks!

PR #493, opened 2020-09-17
Create Ubuntu 18.04 & 20.04 packages with GitHub Actions

pierre-guillou:
This PR adds the support of GitHub Actions to the TTK repository to build .deb packages targeting the last two Ubuntu LTS versions (Bionic 18.04 and Focal 20.04).

To do so, a `main.yml` file has been added to the `.github/workflows` directory and CPack metadata have been added to the main `CMakeLists.txt` file. The two generated .deb packages are tested on the four examples in this repository in a clean VM.

The creation of those packages depends on pushing tags matching the `v*` and `dev*` patterns. A GitHub release will be created with links to the generated packages. This time, the releases are marked as pre-releases (c.f. `actions/create-release@v1` in the YAML file).

To help with the package generation, the handling of the optional dependencies zfp and zlib (and eigen in a smaller degree) has been modified: their headers are now included in the `TopologicalCompression.cpp` source file instead of `TopologicalCompression.h`  As a consequence, they have become run-time dependencies instead of build-time.

To further circumvent the absence of packaging for zfp, the CI builds zfp as a static library and the base TTK modules are build as shared libraries. This ensures that the zfp code used in `topologicalCompression` ends in the created shared library. Thus, there is no need to package zfp ourselves. 

To enable the build of the base modules as shared libraries, a new CMake option, `TTK_ENABLE_SHARED_BASE_LIBRARIES` has been introduced. It is set to false by default for keeping the old behavior of build static libraries. Note that for TTK to behave correctly, the install folder of those base module shared libraries must be referenced by the `LD_LIBRARY_PATH` environment variable.

Environment variables such as `PV_PLUGIN_PATH` and `LD_LIBRARY_PATH` are set in the new `ttk-env.sh`, generated and installed in `/etc/profile.d` on Linux targets. This file should be sourced automatically when login into the graphical session but has to be sourced manually after installing the packages if the user does not want to reload their session.

Enjoy,
Pierre


julien-tierny:
the ci is timing out for windows but I'm gonna go ahead and merge that one.

this is some awesome work and an important milestone for ttk.

@lhofmann and @cgarth 
we'll be triggering the build of *deb packages for each new release (we can do intermediate snapshots if you want).
this should make anaconda and docker packaging much simpler.
thanks for letting me know if you want me to trigger the build of a snapshot now, in order to try things out on your ends.

cheers,

chgarth:
@julien-tierny I'm slightly cautious about using debs (or the conda binary package) to set up the Docker images. That would seem to loose a lot of flexibility in customizing the images easily (e.g. enable Catalyst, OSPRay / OptiX rendering, etc.).

As it is, I have TTK running on a HPC cluster using Singularity, from the Docker image customized with the cluster's MPI library. I don't think that would be possible without major packaging surgery, and I think it's an important use case.

Thoughts?

lhofmann:
@julien-tierny As for the conda package, there is a github bot in place that watches for new releases and automatically creates PRs at https://github.com/conda-forge/topologytoolkit-feedstock so no need to do anything here

julien-tierny:
hi christoph,

> I'm slightly cautious about using debs 
I was just mentioning that in case you would find it convenient :)
I understand the desire to fine-tune dependencies (we tried to pull them all, but ospray is not in there as far as I remember).

cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
awesome!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #494, opened 2020-09-18
Missing vtkVersion includes

lhofmann:
This PR adds missing vtkVersion.h includes in places that use `VTK_MAJOR_VERSION` macros.
This has caused build errors for me using ParaView 5.8.0.
I'm not sure if this will not break elsewhere, since in one place this include was commented out, and there is a comment about this in config.cmake: https://github.com/topology-tool-kit/ttk/blob/dev/config.cmake#L38

Also removes the use of the deprecated `vtkValuePass::SetRenderingMode` in ttkCinemaImaging, which fails to build when VTK was configured with `VTK_LEGACY_REMOVE=ON`.


julien-tierny:
looks good, thanks a lot lutz!

PR #495, opened 2020-09-18
Remove linking against all VTK_LIBRARIES

lhofmann:
This PR removes linking against ${VTK_LIBRARIES} in the `ttk_add_vtk_module` macro and adds all missing VTK module dependencies.

This way, the VTK modules only link the VTK libraries that are actually needed. Linking all VTK libraries broke building TTK in the ParaView superbuild for me (the TTK modules were linked against vtkWrappingTools, which is not contained in the binary distribution).

### DIFF
--- a/CMake/VTKModule.cmake
+++ b/CMake/VTKModule.cmake
@@ -41,7 +41,6 @@ macro(ttk_add_vtk_module)
 
     vtk_module_link(${TTK_NAME}
       PUBLIC
-        ${VTK_LIBRARIES}
         ${TTK_DEPENDS}
       )
 
--- a/core/vtk/ttkCinemaImaging/vtk.module
+++ b/core/vtk/ttkCinemaImaging/vtk.module
@@ -3,3 +3,5 @@ NAME
 DEPENDS
   ttkAlgorithm
   VTK::RenderingOpenGL2
+PRIVATE_DEPENDS
+  VTK::FiltersGeometry
--- a/core/vtk/ttkCinemaProductReader/vtk.module
+++ b/core/vtk/ttkCinemaProductReader/vtk.module
@@ -3,3 +3,7 @@ NAME
 DEPENDS
   ttkAlgorithm
   ttkTopologicalCompressionReader
+PRIVATE_DEPENDS
+  VTK::IOImage
+  VTK::IOLegacy
+  VTK::IOXML
--- a/core/vtk/ttkCinemaProductReader/vtk.module
+++ b/core/vtk/ttkCinemaProductReader/vtk.module
@@ -3,7 +3,6 @@ NAME
 DEPENDS
   ttkAlgorithm
   ttkTopologicalCompressionReader
-PRIVATE_DEPENDS
   VTK::IOImage
   VTK::IOLegacy
   VTK::IOXML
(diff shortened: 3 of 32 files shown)


julien-tierny:
hi lutz,

thanks a lot for the heads up. we're not using pv superbuilds over here so we did not see that coming.
@pierre-guillou can you have a look at this PR really (in particular to make sure it's compatible with #493)?

cheers,

julien-tierny:
a minor downside is that it will require more fine-grained vtk knowledge for new module contributors

pierre-guillou:
Hello everyone!

I've cherry-picked this commit and I'm currently running my package generator CI onto it. I'll tell you when it's finished.

pierre-guillou:
So, the GitHub Actions CI has run fine and the generated packages seems to behave correctly.
Everything looks OK for me.

julien-tierny:
awesome, thanks a lot pierre!

julien-tierny:
hi @lhofmann, it looks like this PR breaks macos' build.
could you have a quick look?
cheers

lhofmann:
I'm on it. It might take some time, since I cannot reproduce it locally and need to rely on the CI

julien-tierny:
understood, good luck :)
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
it looks like the ci made it.
could you please double-check that all the examples in the "examples/" directory are still running fine?
cheers,

lhofmann:
> a minor downside is that it will require more fine-grained vtk knowledge for new module contributors

I agree that this could be annoying for new devs.
I would add a small paragraph about it to the ttk docs. Could you point me to a suitable location in the documentation for this?

> could you please double-check that all the examples in the "examples/" directory are still running fine?

Sure, I will post a message here once I've gotten around to test this.
It would probably be convenient if the CI could just build the examples as well

julien-tierny:
hi lutz,

thanks for your feedback.

> I would add a small paragraph about it to the ttk docs. Could you point me to a suitable location in the documentation for this?
I think the best spot where to put this would be in the doxygen documentation of core/vtk/ttkHellowrld/.
there's a few TODOs already there for newcomers.
I guess you may want to add a new one either in the cpp or h file.

@JonasLukasczyk 
what do you think?
should that be the last TODO?

cheers,
julien
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
> Sure, I will post a message here once I've gotten around to test this.
awesome, thanks!

> It would probably be convenient if the CI could just build the examples as well
actually, if you could have a crack at that too, that'd be terrific :)

cheers,
julien
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

lhofmann:
Examples were successfully built and run by the CI

julien-tierny:
awesome, thanks!
do you think you would include the related documentation in core/vtk/ttkHelloWorld/ as well? 
(put an extra TODO at the appropriate place in the doxygen documentation).

julien-tierny:
alright, I guess that will do.
thanks a lot lutz!

PR #498, opened 2020-09-23
LTS Module

JonasLukasczyk:
Hi Julien,
here finally comes the LTS module. It currently consists of the base code component which is hooked up with the default topological simplification VTK component. The VTK component has a new option called UseLTS, which is off by default.

Please don't format the code until we want to finally merge it in. At least until then it is much easier for me to read and debug the code.

Best
Jonas

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -125,6 +125,7 @@ jobs:
                     -DCMAKE_BUILD_TYPE=$(BuildType)
                     -DVTK_DIR=$(Build.ArtifactStagingDirectory)/vtk-install/lib/cmake/vtk-$(VTKVPath)
                     -DTTK_BUILD_PARAVIEW_PLUGINS=OFF
+                    -DTTK_ENABLE_CPU_OPTIMIZATION=OFF
                     -DTTK_BUILD_STANDALONE_APPS=ON
                     $(TTK_MODULE_DISABLE)
                     $(TTK_MODULE_TEST)'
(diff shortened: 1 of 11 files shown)


JonasLukasczyk:
Seems like windows has an issue to find boost next:
Cannot open include file: 'boost/math/special_functions/next.hpp'

I have no idea how to fix this. Somebody here with Windows experience?

julien-tierny:
I'll have a look into this when I'll test your branch.
I believe this is as simple as including the Boost::math dep (at multiple spots though, LTS, config.cmake and binary packaging specs)
cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
hi @JonasLukasczyk 
fyi, I'm testing your branch right now.
I set UseLTS to true by default and tested the ttk-data state files.
at the moment, the following state files are failing:
- [x] imageProcessing.pvsm
- [x] oceanVortices.pvsm
- [x] tribute.pvsm

note that tribute.pvsm does not crash (so you can still play around with the gui). in particular, changing the thread number to 1 still results in an incorrect result.

could you have a quick look into this?
I have the feeling this is a small thing.

thanks!

JonasLukasczyk:
Thanks Julien,
I will have a look.

julien-tierny:
btw, it seems like the ci is working for macos now :)
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

JonasLukasczyk:
Thanks! I also noticed you removed the "experimental" warning xD

JonasLukasczyk:
Hi Julien,
I pushed a bug fix. It was fortunately a trivial thing.

Just for kicks, check out the tribute state file in the Node Editor:

Node

julien-tierny:
thanks for the quick fix and the pipeline node editor looks really cool :)

julien-tierny:
@lhofmann it seems we're getting linking runtime error in the CI for this one :(
any idea for a quick fix?
cheers,

lhofmann:
I'm seeing a `Illegal instruction     (core dumped) ./bin/ttkBlankCmd`. Not sure why this happens and it sometimes goes through after a few CI reruns.
Maybe try building with `TTK_ENABLE_CPU_OPTIMIZATION=OFF`?

julien-tierny:
Looks like that was it. Weird though it didn't show up in previous PRs...
thanks!

JonasLukasczyk:
Could also be coincidence. Is just turning the TTK_ENABLE_CPU_OPTIMIZATION off not just a workaround?

lhofmann:
The problem is probably that ccache only sees the `-march=native` flag instead of a list of the actual supported instruction sets, and thus reuses binaries that were compiled on a different CPU architectures between CI runs. That's why the error only occurs sometimes, depending on the hardware that the CI happens to be executed.
With Julien's patch `TTK_ENABLE_CPU_OPTIMIZATION=ON` is still tested on Windows, where there is no ccache.

julien-tierny:
hi @JonasLukasczyk 

it looks like the fix introduced a regression for
- [x] states/uncertainStartingVortex.pvsm

could you have another look please?
cheers,

JonasLukasczyk:
Oh boy, these errors sure get my heart pumping, but I fixed this issue. This was actually an algorithmic issue (more details at 4pm). I will push the fix soon.

julien-tierny:
alright, all flags are green now!
thanks a lot jonas. that's one of the big new features of the next release :)

PR #501, opened 2020-10-01
Reduce PersistenceDiagram overhead

pierre-guillou:
This PR reduce some overhead in the PersistenceDiagram module. For instance, the duration of the `RequestData` method on the `ctBones.vti` has dropped from 12s to 7s.

The modifications include:
* the triangulation precondition for the MorseSmaleComplex is not executed by default
* a PersistencePair struct has been introduced to replace the un-documented, templated tuple that held the pairs data.
* the persistence pair has been set to double, since in the VTK layer it is stored in a vtkDoubleArray. This helps to avoid storing templated vectors of persistence pairs on the heap (one less indirection)
* the methods generating the VTK structures for an embed and a non-embed diagram have been merged together
* filling the VTK arrays is now done in parallel
* a vector of pointers in FTMTreePP has been changed into a vector of the pointed structs to reduce indirections

No change has been observed on the ttk-data states.

Enjoy,
Pierre

### DIFF
--- a/core/base/ftmTree/FTMAtomicUF.h
+++ b/core/base/ftmTree/FTMAtomicUF.h
@@ -26,19 +26,18 @@ namespace ttk {
 
     class AtomicUF {
     private:
-      unsigned rank_;
-      AtomicUF *parent_;
+      unsigned rank_{};
+      AtomicUF *parent_{};
       SharedData data_;
 
     public:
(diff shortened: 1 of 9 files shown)


julien-tierny:
ok, looks good to me.
thanks!

PR #502, opened 2020-10-01
turned ttkUtils into class

JonasLukasczyk:
Hi everyone,
I'm working on modules that have several classes in the vtk component which all require ttkUtils functionality. However, due to a reason I forgot we had to add the content of the ttkUtils.cpp into the ttkUtils.h file. So in my scenario I get "duplicate definition" errors.

I created this PR to separate the implementation from the header again. This also required me to adjust some includes in other modules. If I remember correctly the old version caused linking errors on some platforms. So we have to make sure that everything works (not only the statefiles) before we merge this in.

Best
Jonas

### DIFF
--- a/core/vtk/ttkAlgorithm/ttk.module
+++ b/core/vtk/ttkAlgorithm/ttk.module
@@ -2,6 +2,7 @@ NAME
   ttkAlgorithm
 SOURCES
   ttkAlgorithm.cpp
+  ttkUtils.cpp
 HEADERS
   ttkAlgorithm.h
   ttkUtils.h
--- a/core/vtk/ttkAlgorithm/ttkUtils.cpp
+++ b/core/vtk/ttkAlgorithm/ttkUtils.cpp
@@ -1,0 +1,308 @@
+#include <ttkUtils.h>
+
+#include <limits>
+#include <vtkStringArray.h>
+
+#include <vtkAbstractArray.h>
+#include <vtkCellArray.h>
+#include <vtkDoubleArray.h>
+#include <vtkFieldData.h>
+#include <vtkPoints.h>
+#include <vtkSmartPointer.h>
+
+int ttkUtils::replaceVariable(const std::string &iString,
+                              vtkFieldData *fieldData,
+                              std::string &oString,
+                              std::string &errorMsg) {
+  std::string varName = iString;
+  int varIndex = -1;
+  bool varIndexDefined = false;
+
+  // Check if varIndex is specified
+  size_t indexDelimiter0 = iString.find("[");
+  size_t indexDelimiter1 = iString.find("]");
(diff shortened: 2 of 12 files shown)


JonasLukasczyk:
I don't know why Windows can not resolve the external symbols that correspond to the static functions. Can anyone who has experience with the windows build explain this? We will also run into the same issue with the TriangulationRegistry class I'm working on in a separate PR.

lhofmann:
On MSVC, symbols in dynamic libraries are hidden by default, while on GCC they are visible by default (that's why its working on Linux by accident).
The `ttkUtils` class needs the `TTKALGORITHM_EXPORT` attribute, which sets the visibility flags correctly, similar to the `ttkAlgorithm` class.

JonasLukasczyk:
Thanks Lutz! You are the man.

JonasLukasczyk:
Hi Julien, the statefiles and the helloworld standalone work on my machine. I guess someone should also run a quick test on Windows and then we are good to go. After this is merged in I will open a new PR for the TriangulationFactory.

julien-tierny:
@pierre-guillou
would you happen to have a working windows vm at hand?
cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

JonasLukasczyk:
Then I would have not said: somebody should ;-)

pierre-guillou:
Hi,
Sadly, my windows VM is not set up ATM. I tried to build ParaView 5.8.1 earlier this week and it did not ended well :(
I will retry this afternoon.

JonasLukasczyk:
Any timeline on this since this PR blocks the TriangulationFactory?

julien-tierny:
hi jonas,

thanks for the reminder, but we've just been notified a partial lock down at the universities in paris and that kept us busy.

> Any timeline on this since this PR blocks the TriangulationFactory?
two things:

1) could you update the azure pipeline to use ttkHelloWorldCmd instead of ttkBlandCmd?

2) could you make sure the tests:
- Test TTK Command line
- Test C++ Example
- Test TTKVTK Example
- Test Python Example
- Test pvpython Example
run fine under both MacOs and Windows?
(it seems that those are currently not running with sufficient debug levels)

that'd help get confidence that everything runs fine there.

cheers,
julien
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

pierre-guillou:
Also, I managed to build this branch on my Windows VM (after spending Friday afternoon rebuilding ParaView 5.8).
I'm still in the process of finding the right set of environment variables to help ParaView find the TTK plugin at run-time.
This should be done soon.
Pierre

julien-tierny:
alright, let's go!

PR #503, opened 2020-10-02
Make setThreadNumber virtual (and other fixes)

pierre-guillou:
This PR fixes, among others, a bug introduced during the migration: base modules that redefined the `setThreadNumber` method (MorseSmaleComplex, FTRGraph, ContourForests, Triangulation...) were locked at the maximum number of threads, since `ttkAlgorithm` was calling the `BaseClass` method and not the derived ones. The `BaseClass::setThreadNumber` method has been thusly made virtual (this change also triggers a compiler warning if the derived methods don't respect the base method signature).

Other fixes provided by this PR include:
* some MSVC compiler warning fixes
* disabling clang-format on a sensitive section in `Auction.h`
* reducing the usage of the `use namespace` directive in headers (to avoid side effects)
* adding missing `TTK_ENABLE_OPENMP` guards around OpenMP pragmas (especially in LocalizedTopologicalSimplification, which has been clang-formatted for the occasion)
* a global include of the `ciso646` header that allows MSVC to support the `and` and `or` keywords
* a memory leak fix in the `getcwd` call in a ununsed method of `BaseClass`
* the handling of the ZFP and zlib dependencies for the base libraries has also been resurrected in `TTKBaseConfig.cmake` but only when the base libraries are statically compiled.

No change has been observed in the ttk-data states.

Enjoy,
Pierre

### DIFF
--- a/core/base/TTKBaseConfig.cmake.in
+++ b/core/base/TTKBaseConfig.cmake.in
@@ -13,5 +13,14 @@ if (@TTK_ENABLE_OPENMP@)
   find_dependency(OpenMP REQUIRED)
 endif()
 
+if (@TTK_ENABLE_ZFP@ AND NOT @TTK_ENABLE_SHARED_BASE_LIBRARIES@)
+  set(ZFP_DIR "@ZFP_DIR@" CACHE PATH "Use TTK ZFP dir" FORCE)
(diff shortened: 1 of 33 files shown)


julien-tierny:
alright, thanks!

PR #504, opened 2020-10-02
MorseSmaleComplex: Parallelize filling the VTK arrays

pierre-guillou:
This PR reduces the overhead in the MorseSmaleComplex filter by parallelizing the last step i.e. the filling of the VTK arrays.

To do so, VTK arrays are pre-allocated with the correct size and filled in parallel using `SetTuple` instead of `InsertNextTuple`.

To manage the cells of the vtkUnstructuredGrids underlying the 1-separatrices and the 2-separatrices representations, the interleaved layout (npoints, pt0id, pt1id...) has been dropped in favor of the new offset/connectivity arrays (1-separatrices don't need to fill an offset array in the base layer since the produced cells are all VTK_LINES with two vertices).

As a result, this step has been reduced from 60s to 5s for a full MorseSmaleComplex (incl. 2-separatrices) of the ctBones.vti dataset. For reference, the base layer computation takes approximately 90s.

No change has been observed in the concerned ttk-data state files.

In a next step, I will try to avoid those now parallel copies promoting the temporary vectors filled in the base layer to class member and using `ttkUtils::SetVoidArray` to reference them in the VTK arrays. This should reduce the memory footprint of the filter (especially with the 2-separatrices).

Enjoy,
Pierre

### DIFF
--- a/core/base/abstractMorseSmaleComplex/AbstractMorseSmaleComplex.h
+++ b/core/base/abstractMorseSmaleComplex/AbstractMorseSmaleComplex.h
@@ -325,7 +325,8 @@ namespace ttk {
       SimplexId *const separatrices2_numberOfPoints,
       std::vector<float> *const separatrices2_points,
       SimplexId *const separatrices2_numberOfCells,
-      std::vector<SimplexId> *const separatrices2_cells,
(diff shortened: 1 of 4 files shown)


julien-tierny:
cool, thanks!

PR #506, opened 2020-10-05
PersistenceDiagram/Curve: Shunt the MorseSmaleComplex dependency

pierre-guillou:
This PR removes the PersistenceDiagram and PersistenceCurve dependency to the MorseSmaleComplex module for saddle-saddle pairs by moving the called MorseSmaleComplex method into the DiscreteGradient module.

Since the `MorseSmaleComplex3D::computePersistencePairs` method doesn't use anything specific to the MorseSmaleComplex and rather calls directly into the DiscreteGradient, moving it to the DiscreteGradient module is the logic action.

As a consequence, a modification of the MorseSmaleComplex headers should not trigger the recompilation of the PersistenceDiagram and other modules that depend on it (TrackingFromFields). However, no reduction of the build times of ttkPersistenceDiagram and ttkPersistenceCurve has been observed.

Also including a little bit of cleaning of the DiscreteGradient module.

No change has been observed on the ttk-data states.

Enjoy,
Pierre

### DIFF
--- a/core/base/abstractMorseSmaleComplex/AbstractMorseSmaleComplex.h
+++ b/core/base/abstractMorseSmaleComplex/AbstractMorseSmaleComplex.h
@@ -578,7 +578,6 @@ int ttk::AbstractMorseSmaleComplex::setSeparatrices1(
 #endif
 
   const auto scalars = static_cast<const dataType *>(inputScalarField_);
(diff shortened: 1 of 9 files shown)


julien-tierny:
alright, looks good, thanks.

Issue #490, opened 2020-09-14
ttkTriangulation assumes 64bit storage of topology in vtkCellArray


lhofmann:
**Describe the bug**
The VTK 9 version of vtkCellArray supports both 32bit and 64bit storage of the topology.
However, ttkTriangulation casts results of `vtkCellArray::GetConnectivityArray()` and  `vtkCellArray::GetOffsetsArray()` to LongSimplexId: https://github.com/topology-tool-kit/ttk/blob/dev/core/vtk/ttkTriangulationAlgorithm/ttkTriangulation.cpp#L322
This leads to undefined behaviour when `vtkCellArray::IsStorage64Bit()` returns false.

**To Reproduce**
```python
import vtk
import topologytoolkit as ttk

reader = vtk.vtkXMLUnstructuredGridReader()
reader.SetFileName('ttk-data/bivariateToy.vtu')
reader.Update()
# this causes the bug:
reader.GetOutput().GetCells().ConvertTo32BitStorage()

csp = ttk.ttkContinuousScatterPlot()
csp.SetInputDataObject(reader.GetOutput())
csp.Update()
```

This will result in an error similar to
```
[ExplicitTriangulation] [ERROR] Unable to create triangulation for cells of dimension 4 or higher (34359738375).
```

**Expected behavior**
It should either also work with 32bit data, or ttkTriangulation would need to check the storage type and raise an error.

julien-tierny:
hi @lhofmann 
thanks for spotting that out.
we'll have a look into this before pulling the next release.
cheers,

julien-tierny:
closing (addressed in #526)

PR #507, opened 2020-10-06
MorseSmaleComplex: Don't copy base layer output into VTK arrays

pierre-guillou:
This PR aims at eliminate data copies in the VTK layer of the MorseSmaleComplex module. To do so, local vectors that held the output of the base layer have been promoted to class members. VTK arrays now point to these class members using the `ttkUtils::SetVoidArray` method.

Some templated vectors (separatrixFunctionMaxima, separatrixFunctionMinima, separatrixFunctionDiffs) have been converted to vectors of double to allow this transformation. However, the `criticalPoints_points_cellScalars` vector has been kept templatized (critical points arrays should be smaller than separatrices arrays).

To keep compatibility with the ExplicitTriangulation, offset and connectivity cell arrays use the `long long` type (c.f. #490) instead of the more generic (and less memory-consuming) `SimplexId`.

Execution times on the relevant part of code (the second half of the `dispatch` method in `ttkMorseSmaleComplex.cpp`) of a full MorseSmaleComplex on `ctBones.vti` have been reduced to 0.5s from 5s (was 60s before #504).

No changes in the relevant ttk-data have been observed.

Enjoy,
Pierre

### DIFF
--- a/core/base/abstractMorseSmaleComplex/AbstractMorseSmaleComplex.h
+++ b/core/base/abstractMorseSmaleComplex/AbstractMorseSmaleComplex.h
@@ -282,14 +282,14 @@ namespace ttk {
       std::vector<char> *const separatrices1_points_cellDimensions,
       std::vector<SimplexId> *const separatrices1_points_cellIds,
       SimplexId *const separatrices1_numberOfCells,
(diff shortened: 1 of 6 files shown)


julien-tierny:
looks good, thanks!

Issue #500, opened 2020-09-29
Unnecessary implicit preconditioning


julien-tierny:
**Describe the bug**
When chaining filters in paraview on a regular grid, the implicit preconditioning happens at each filter. In previous versions, that preconditioning took virtually no time, now it does

**To Reproduce**
1. Load a regular grid
2. Apply a scalar field smoother 1 (max debug level)
3. Apply a scalar field smoother 2 (max debug level)
4. Apply a scalar field smoother 3 (max debug level)
In the console, the line ```Apply a scalar field smoother 2 (max debug level)``` occurs 3 times

**Expected behavior**
The preconditioning should happen only once and only 1 triangulation should be registered.

**Screenshots**
![implicitPreconditioning](https://user-images.githubusercontent.com/3018628/94549699-e788ad00-0252-11eb-958c-127a8daada6c.png)

**System (please complete the following information):**
 - OS: gentoo linux
 - Compiler gcc 9.3
 - Installed TTK's optional dependencies all

**Additional context**
@JonasLukasczyk 
I didn't have time to further look into this but that'd be great if we come up with a fix before the release.

cheers

pierre-guillou:
Seems that `ttk::AbstractTriangulation::hasPreconditionedVertexNeighbors_` should be tested in `ttk::AbstractTriangulation::preconditionVertexNeighbors()` before calling the internal method.

julien-tierny:
actually, I think the issue rather comes from the number of registered triangulations which should be 1 instead of 3 (see debug messages)

pierre-guillou:
You are right, one triangulation object is generated per filter (instead of modifying the same triangulation object in the pipeline).

julien-tierny:
alright, looks good to me!
cheers,

PR #509, opened 2020-10-06
Triangulation Factory

JonasLukasczyk:
Hi Julien,
this PR introduces a new class in core/vtk/ttkAlgorithm called ttkTriangulationFactory. This class manages the instantiation, retrieval, update, and deletion of ttk::Triangulations for vtkDataSets.

This PR solves issue #500.

I ran all statefiles in ttk-data without problems.

Best
Jonas


julien-tierny:
alright, looks good to me.
looking forward to playing with it  :) :) :)

Issue #489, opened 2020-09-12
XML file of ttkIcosphere specifies an unused input object


JonasLukasczyk:
The ParaView XML file of ttkIcoSphere still specifies an input object that was previously used as an optional input for a BoundingIcosphere computation. This is a legacy left-over since this functionality is now handled by ttkIcosphereFromObject. So this xml statement should be safe to remove.

I will do this myself and just record this here as a reminder.

JonasLukasczyk:
This issue has been solved with PR #510

PR #510, opened 2020-10-06
removed legacy input port of ttkIcoSphere

JonasLukasczyk:
This PR solves issue #489


julien-tierny:
sounds good for me.
cheers

PR #511, opened 2020-10-08
MorseSmaleComplex: Reduce ascending 2-separatrices computation memory footprint

pierre-guillou:
This PR improves the work done in #387 on parallelizing the computation of the ascending 2-separatrices. After some memory profiling using [heaptrack](https://github.com/KDE/heaptrack), memory allocations have been reworked to reduce the memory footprint of the `MorseSmaleComplex3D::setAscendingSeparatrices2` method.

As a result, the peak memory consumption of this method on `ctBones.vti` has been reduced from 35GB to 24GB. As it turns out, execution times of this particular method on the same dataset have also dropped from 33s to 22s.

No changes observed in the relevant ttk-data states.

Enjoy,
Pierre


julien-tierny:
cool, thanks!

PR #512, opened 2020-10-08
Triangulation: Avoid preconditioning twice

pierre-guillou:
This PR adds check in the `AbstractTriangulation` class to avoid running several times the same precondition function on the same triangulation object.

Special care has been taken to ensure that, when using the PeriodicGrid filter, preconditioned implicit triangulations leave place to unpreconditioned periodic ones. Switching back to the previously preconditioned implicit triangulation should be supported too, since although the `hasPreconditioned` booleans are set to false in the `ttk::Triangulation` object, they stay at their previous states in the contained `ttk::ImplicitTriangulation` and `ttk::PeriodicImplicitTriangulation` objects.

No changes observed in the ttk-data states (apart from segfaults in ctBones, interactionSites and morsePersistence, seemingly due to #509).

Enjoy,
Pierre

### DIFF
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -1831,9 +1831,12 @@ namespace ttk {
     /// \return Returns 0 upon success, negative values otherwise.
     /// \sa isEdgeOnBoundary()
     virtual inline int preconditionBoundaryEdges() {
-      preconditionEdges();
-      preconditionBoundaryEdgesInternal();
-      hasPreconditionedBoundaryEdges_ = true;
+
+      if(!hasPreconditionedBoundaryEdges_) {
+        preconditionEdges();
+        preconditionBoundaryEdgesInternal();
+        hasPreconditionedBoundaryEdges_ = true;
+      }
       return 0;
     }
 
@@ -1851,9 +1854,12 @@ namespace ttk {
     /// \return Returns 0 upon success, negative values otherwise.
     /// \sa isTriangleOnBoundary()
     virtual inline int preconditionBoundaryTriangles() {
-      hasPreconditionedBoundaryTriangles_ = true;
-      preconditionTriangles();
-      preconditionBoundaryTrianglesInternal();
+
+      if(!hasPreconditionedBoundaryTriangles_) {
+        preconditionTriangles();
+        preconditionBoundaryTrianglesInternal();
+        hasPreconditionedBoundaryTriangles_ = true;
+      }
       return 0;
     }
 
@@ -1871,8 +1877,11 @@ namespace ttk {
     /// \return Returns 0 upon success, negative values otherwise.
     /// \sa isVertexOnBoundary()
     virtual inline int preconditionBoundaryVertices() {
-      hasPreconditionedBoundaryVertices_ = true;
-      preconditionBoundaryVerticesInternal();
+
+      if(!hasPreconditionedBoundaryVertices_) {
+        preconditionBoundaryVerticesInternal();
+        hasPreconditionedBoundaryVertices_ = true;
+      }
       return 0;
     }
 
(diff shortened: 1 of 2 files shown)


julien-tierny:
@JonasLukasczyk 
I'm in sync with PR #509 and I do observe these crashes as well.
can you confirm asap that you observe these crashes too?
cheers

JonasLukasczyk:
Hi all,
I can see these crashes too. I will investigate further, but this happens when LTS is called with an invalid set of authorized extrema (for example if LTS is tasked to remove all extrema). This in turn happens if FTM tree computes an incorrect result (which it sometimes does). I will check add some sanity checks to LTS to exit gracefully.

Best
Jonas

julien-tierny:
hi jonas,

as a quick note, I tested again all states a few days ago (way after the latest lts edit) and things worked ok.

as pointed out by pierre, the problem seems to come from #509.

cheers,
julien
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

JonasLukasczyk:
Then there are two issues ;-)

JonasLukasczyk:
@pierre-guillou what makes you believe it was #509? I actually think this is related to LTS + FTM. For example, ctBones crashes sometimes in initialize LTS propagations due to FTM tree.

julien-tierny:
for ctBones, the crash occurs in a section that does not seem related to ftm or lts:

```
[GeometrySmoother] %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
[TriangulationFactory] [WARNING] Cells are not simplices.
[TriangulationFactory] [WARNING] Consider using `vtkTetrahedralize` in pre-processing.
[GeometrySmoother] [ERROR] Unable to retrieve/initialize triangulation for 'vtkPolyData'

Loguru caught a signal: SIGSEGV
```

JonasLukasczyk:
I can confirm that both issues occur in ctBones. If FTM tree computes the correct result then LTS runs fine and later in the pipeline the geometry smoother is asked to fetch a triangulation for a non-simplex mesh. I will investigate, but I think I remember now that in the past we still returned a triangulation and kept our fingers crossed that everything will run fine (for example if the cells are quads).

JonasLukasczyk:
Right now the Factory complains and return null, and then further down the pipeline some filter does not exit gracefully.

julien-tierny:
alright, thanks for your feedback.
as a side note, when I consider a PR for merging, I expect that its author did run all the pvsm state files from ttk-data.
thanks for double checking that in the future.

JonasLukasczyk:
I checked, but I blamed it on FTM and moved on :P 
The fingers crossed method solves the second issue. Should I open a PR for that or should we actually enforce a tetrahedralization?

JonasLukasczyk:
I could also add VTK_QUAD to the list of acceptable cell types as this seems to work.

julien-tierny:
> Should I open a PR for that or should we actually enforce a tetrahedralization? 
I think you should open the PR right away to restore the states files.
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
this will only work for getVertexNeighbors (which is used here).
that's something that needs to be fixed in the state file (we can open an issue for that), i.e. to call tetrahedralize prior to geometry smoothing.
cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

julien-tierny:
maybe we should open an issue here and put @CharlesGueunet in the loop.
my impression is that it's small thing that could be prevented by a tiny safeguard. that'd be much easier for charles to come up with a quick fix than for any others. 
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

JonasLukasczyk:
The PR for the quick fix is almost ready. But should I not instead open a PR on ttk-data where I run in the three statefiles tetrahedralize first? I think this would actually be the correct solution.

julien-tierny:
both options are fine with me.
(make sure to minimally edit the state files)
cheers
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

pierre-guillou:
The issue occurs in `ttkTriangulationFactory::CreateExplicitTriangulation`. The only difference to before #509 is that a nullptr is returned if cells are not simplices…

JonasLukasczyk:
Thanks Pierre, we decided for now to keep it like this and instead update the statefiles in ttk-data to run tetrahedralize before the GeometrySmoothers. I also opened a PR that adds a sanity check in ttkGeometrySmoother that will let it exit gracefully.

pierre-guillou:
As for `ctBones.pvsm`, I also often experiment this problem in FTM when the global persistence pair seems to be missing, leading to errors downstream.

julien-tierny:
alright, thanks for these fixes!

PR #514, opened 2020-10-09
Exit Gracefully

JonasLukasczyk:
This PR consists of two lines that let the ttkGeometrySmoother exit gracefully if the filter can not fetch a triangulation. Note: the ttkScalarFieldSmoother does this already.

Best
Jonas


JonasLukasczyk:
I guess this PR does not need CI.

julien-tierny:
that was an easy one indeed, thanks!

Issue #505, opened 2020-10-05
Fix MacOs and Windows CI


julien-tierny:
- [x] use ttkHelloWorldCmd instead of ttkBlankCmd

- [x] fix debug level in Windows and MacOS azure tests (issue with static variable)

julien-tierny:
Fixed by #516 and later commits.

PR #516, opened 2020-10-16
Enable shared base libraries to solve Windows log problem

pierre-guillou:
This PR aims at fixing the second #505 issue. As a reminder, the issue was that setting the `ttk::globalDebugLevel_` global variable inside the vtk-c++ example had not the expected effect. Due to our default build process, since the base layer is built as static libraries that are replicated inside the shared libraries in the VTK layer, every global variable is duplicated inside those VTK layer shared libraries. 

After some documentation, it appears that in a Windows process, shared libraries don't share the same memory space, thus making it impossible to modify the duplicated global variables. In Linux, the dynamic loader and the runtime loader seem to handle this situation quite fine (apart from some duplicated welcome messages).

The solution is to leverage work done in #493 to build base libraries as shared libraries instead of static. In this case, the `common.dll` containing the global variables is linked to all VTK layer shared libraries and the global variables have a unique memory address.

This lead to some interesting discoveries on Windows shared libraries: by default, their symbols are hidden, causing link errors. For functions and classes, this is easily fixed by using the `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` option. For global (and static) variables, you have to annotate their declaration and their definition with `__declspec(dllexport)` in the compilation units of the current library, and their extern declarations in other compilation units with `__declspec(dllimport)`. Fortunately, the compilation process identifies every library its builds with a unique macro. I thus leveraged the `common_EXPORTS` macro to annotate the 4 global variables inside the `common` library: `globalDebugLevel_`, `welcomeMsg_`, `goodbyeMsg_` and `globalThreadNumber_` and the static variable `Debug::lastLineMode`. To avoid expanding too much this solution, I removed the also problematic static variables inside the `TopologicalCompression` class (😢). One consequence is that TTK devs should now avoid to use global or static variables in the base layer or else annotate them in a similar (and quite ugly) way.

In the VTK layer, the configuration process generates the `Module.h` headers that define  those `TTKMODULE_EXPORTS` macros automatically, as explained in https://github.com/topology-tool-kit/ttk/pull/502#issuecomment-702186739. Some VTK layer modules (`ttkUserInterfaceBase`, `ttkWRLExporter`) using the wrong annotations have been fixed in the process.

One maybe breaking change of this PR is that the `TTK_ENABLE_SHARED_BASE_LIBRARIES` has been switched to ON by default. This might force TTK devs to expand their `LD_LIBRARY_PATH` environment variable with the install path of the base libraries. This also might solve the same issue on MacOS.

Everything should still work with static libraries on Windows (except the log problem of course).

Enjoy,
Pierre

### DIFF
--- a/CMake/BaseCode.cmake
+++ b/CMake/BaseCode.cmake
@@ -145,6 +145,10 @@ function(ttk_set_compile_options library)
     target_compile_definitions(${library} PUBLIC TTK_ENABLE_DOUBLE_TEMPLATING)
   endif()
 
+  if (TTK_ENABLE_SHARED_BASE_LIBRARIES)
+    target_compile_definitions(${library} PUBLIC TTK_ENABLE_SHARED_BASE_LIBRARIES)
+  endif()
+
   if (TTK_ENABLE_KAMIKAZE)
     target_compile_definitions(${library} PUBLIC TTK_ENABLE_KAMIKAZE)
   endif()
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -606,7 +606,7 @@ jobs:
     displayName: 'Build TTKVTK Example'
 
   - bash: |
-      ./ttkExample-c++ -i ../examples/data/inputData.off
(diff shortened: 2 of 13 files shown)


pierre-guillou:
It seems that the log problem is also solved for macOS 😃

julien-tierny:
very informative, thanks!

PR #517, opened 2020-10-21
Rework installation directories to avoid setting environment variables

pierre-guillou:
This PR reworks the installation process to avoid setting environment variables and to make the installed libraries directly usable inside ParaView, especially for the Ubuntu packages.

To do so, base and VTK shared libraries have been moved from `$prefix/lib/ttk` to `$prefix/lib`, plugin files to `$prefix/bin/plugins` and CMake target files to `$prefix/lib/cmake`. This should ensure that setting PV_PLUGIN_PATH, LD_LIBRARY_PATH and CMAKE_PREFIX_PATH is not needed anymore when installing in the `/usr` prefix.

On Ubuntu, due to the non-standard Python site-package location, setting PYTHONPATH is still needed for the interpreter to find the topologytoolkit module. The problem mainly comes from VTK, which installs its Python modules in a non-(Ubuntu-)standard location (`/usr/lib/x86_64-linux-gnu/python3.6/site-packages` vs `/usr/lib/python3/dist-packages`). Fixing it would need to modify the ParaView CMake code; I would rather avoid it. Fortunately, `pvpython` does not need PYTHONPATH.

Now for the bad part: those are breaking changes. TTK devs will need to change their environment to reflect those changes in installation directories, especially if they use the default prefix (`/usr/local`). If their ParaView has been installed in this prefix, only the LD_LIBRARY_PATH and the CMAKE_PREFIX_PATH have to be modified. The TTK_INSTALL_PLUGIN_PATH CMake cache variable will also be invalidated by these changes: TTK devs will have to remove the corresponding entry to the CMake cache manually (or reconfigure a build directory from scratch).

A second concern of mine, about the Ubuntu packages, is that since we installed all base libraries into `$prefix/lib`, they might collide with libraries from other packages. Indeed, the base libraries are not prefixed with `ttk` and have very common names such as `libauction.so` or `libcommon.so`. I would suggest to assemble all base libraries into a single (but quite tiny) shared library, named by example `libttkbase.so`, since the majority of the code is located in the headers. I don't know how much work this would entail though.

To sum up, these changes make the TTK install process less custom and more user-friendly, with the cost of breaking changes, mostly for the TTK dev.

Enjoy,
Pierre


julien-tierny:
hi pierre,

thanks for the PR.

- for paraview plugins, changing the installation path to the default plugin path for paraview (`$prefix/bin/plugins`) is the way to go. 

- for the libraries, installing everything to `$prefix/lib` does not seem like the right way to go (for various reasons, including the ones you mention). for a reminder, under ubuntu, when everything (base + vtk layers) is installed in `$prefix/lib/ttk` there is no linking issue anymore. for ubuntu packages, the problem comes from the fact that the base shared libraries are installed in `$prefix/lib/x86_64-linux-gnu/ttk` which is not listed in the linking environment variables.

- for PYTHONPATH, I think it's still ok to set it manually.

so to recap, for me, the fix here is to:
- change the installation path for paraview plugins to `$prefix/bin/plugins`
- change the installation path for the base shared objects to `$prefix/lib/ttk` instead of `$prefix/lib/x86_64-linux-gnu/ttk` (an alternative would consist in adding  `$prefix/lib/x86_64-linux-gnu/ttk` to /etc/ld.so.conf.d/ttk.conf and running ldconfig in the package install post-process).

cheers,
julien
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

pierre-guillou:
> for paraview plugins, changing the installation path to the default plugin path for paraview (`$prefix/bin/plugins`) is the way to go.

Well, except if TTK's prefix is different from ParaView's prefix (what is happening in the Azure Pipelines).

> for a reminder, under ubuntu, when everything (base + vtk layers) is installed in `$prefix/lib/ttk` there is no linking issue anymore.

Actually, I just tried to force the installation of the base and VTK shared libraries to `/usr/lib/ttk` in the deb packages in my Ubuntu 18.04 VM. Sadly, I get errors about missing TTK filters while running ParaView on the example visualizations. I still need to set `LD_LIBRARY_PATH=/usr/lib/ttk` to make everything work again.

pierre-guillou:
Concerning the potential collisions with base libraries, actually it is quite easy to use the [OUTPUT_NAME](https://cmake.org/cmake/help/latest/prop_tgt/OUTPUT_NAME.html) CMake property to prefix all base library file names (with `ttkBase` for example). This way, we would get `libttkBasecommon.so` instead of `libcommon.so`.

pierre-guillou:
> Actually, I just tried to force the installation of the base and VTK shared libraries to `/usr/lib/ttk` in the deb packages in my Ubuntu 18.04 VM. Sadly, I get errors about missing TTK filters while running ParaView on the example visualizations. I still need to set `LD_LIBRARY_PATH=/usr/lib/ttk` to make everything work again.

In fact, when I was experimenting this, I was using the default prefix `/usr/local` instead of `/usr`. The packaging operation still installed TTK inside `/usr` but the rpath to the base libraries was mixed up, causing this error.

In fact, packaging with `/usr` as prefix and putting the base + VTK libraries inside `/usr/lib/ttk` works. Sorry for the wrong call here.

pierre-guillou:
OK, so I think I found what causes the mismatch between `/usr/lib/ttk` and `/usr/lib/x86-64-linux-gnu/ttk`. The way to target `/usr/lib/x86-64-linux-gnu/ttk` is to use `${CMAKE_INSTALL_LIBDIR}/ttk` as install directory for libraries, whereas to target `/usr/lib/ttk` I used `lib/ttk` directly in the CMake `install` commands.

What happens behind the scene is that, with `${CMAKE_INSTALL_LIBDIR}/ttk`, CMake still targets `/usr/lib/ttk` but is superseded by dpkg during the packaging step. The latter will silently move the libraries into `/usr/lib/x86-64-linux-gnu/ttk`. It happens that both `/usr/lib/ttk` and `/usr/lib/x86-64-linux-gnu/ttk` are not covered by the dynamic loader `ld.so`; a second system, embedded inside the shared libraries, ensures that from the plugin in `/usr/bin/plugins` loaded by ParaView, the right shared libraries are also loaded. The problem is that during the packaging step, dpkg does not update this path. Thus libraries stored `/usr/lib/x86-64-linux-gnu/ttk` can't be found by the plugin.

This second system is the RPATH or RUNPATH. It's a path written in the libraries that indicates where to look at when searching other linked shared libraries from this one. It can be checked with `objdump -p lib.so`. In both cases, its value is `/usr/lib/ttk`.

Fortunately, CMake offers the `CMAKE_INSTALL_RPATH` variable that can be set to the correct directory (`/usr/lib/x86-64-linux-gnu/ttk`) when building Debian packages. This variable can be set from the command line, for example in the packaging pipelines. This does not solve the issue of building for Ubuntu with `/usr` as prefix without packaging though.

In my opinion, we should copy what ParaView and VTK do, i.e. dumping all our shared libraries into the default destination directory (in the CMake sense), which is `/usr/lib/x86_64-linux-gnu` on Debian derivatives and `/usr/lib` otherwise (and maybe something else on other more obscure Linux platforms). This way, everything is found without workarounds if the prefix is `/usr` (well, except the Python modules on Ubuntu).

In the meantime, the macOS CI has finally completed without errors.

CMakeLists.txt
```diff
@@ -101,21 +103,12 @@ if(TTK_BUILD_PARAVIEW_PLUGINS)
...
-  if(NOT "$ENV{PV_PLUGIN_PATH}" STREQUAL "")
-    set(TTK_INSTALL_PLUGIN_DIR
-      $ENV{PV_PLUGIN_PATH}
-      CACHE
-      PATH
-        "Directory where the ParaView plugin will be installed"
-      )
-  else()
-    set(TTK_INSTALL_PLUGIN_DIR
-      ${ParaView_DIR}/../../../bin/plugins/
```
julien-tierny: thanks for combining your new solution `"${CMAKE_INSTALL_BINDIR}/plugins"` with this old setup (as discussed in the meetup)

pierre-guillou: Done!

julien-tierny:
that'd be neat to check how to avoid the duplicated install of libraries for the pv plugin (as discussed in the meetup).

CMakeLists.txt
julien-tierny: awesome!

pierre-guillou:
I'll try more but it's quite hard to understand where to act: there's a lot of calls to CMake functions from the ParaView CMake configuration. I hope there is a simple way to do it.

julien-tierny:
seems like @lhofmann  found the trick in #523 
that was fast :)

pierre-guillou:
I see that!

julien-tierny:
#523 seems to run into the same issue again. 
maybe a more permanent workaround consists in caching a build of pv 5.8.1 (and not 5.8.0)?

CMakeLists.txt
pierre-guillou: Actually, after some experimentation and considerations, I think keeping this old mechanism is a bad idea (for regular users/devs, not for the package users):
- it introduces an absolute path inside the configuration, which can cause side-effects, especially if you have several installations and build directories. You may overwrite the plugin of a previous installation if you forget to unset `PV_PLUGIN_PATH` before the configuration step.
- the `paraview_plugin_build` function call inside paraview/CMakeLists.txt only accepts relative paths (to `$ttk_prefix`). To support this mechanism, I hard-coded the `$ttk_prefix/bin/plugins` path there. An `install` call in the same file manages the installation in the user-provided `PV_PLUGIN_PATH`, resulting in a duplication of the plugin.
- `PV_PLUGIN_PATH`, like `PATH`, may contain a list of paths. For example, you can load all the ParaView plugins inside `$pv_prefix/lib/paraview-5.8/plugins` + TTK at launch this way.

julien-tierny:
hi pierre, could you take care of the merge with the current dev branch?
cheers

julien-tierny:
it seems like this one is ready to go right?

pierre-guillou:
Well, there is still is these possible issues with the use of PV_PLUGIN_PATH but considering that they were also here before, it should be good.

julien-tierny:
you mean removing PV_PLUGIN_PATH from installation paths?
I could go either way. I understand your point. I just wanted to (at least) offer some compatibility with ktiware's official support.
by default (if unset) it'll go to the right directory anyways.

pierre-guillou:
We could keep it for the time being and remove it afterwards if somebody complains about it louder.

julien-tierny:
sounds good.
thanks!

pierre-guillou:
Ok, so you also need to port the one-liner inside `paraview/patches/main.yml` to the ttk-paraview repo and rebuild the packages. I changed the installation path of the Python VTK modules to avoid setting PYTHONPATH.

julien-tierny:
done! (currently building).
thanks!

PR #519, opened 2020-10-26
New Backend for ttkForEach/ttkEndFor

JonasLukasczyk:
Hi Julien,
this PR changes how the ttkForEach and ttkEndFor do their thing behind the scenes. Before we exploited a hack that allowed us to stream elements over time. This obviously conflicted with datasets that actually have time steps and this mechanism didn't allow nested for loops.

Now these filters actually iterate a subpart of the pipeline, which is much more stable, efficient, and allows nesting.

All statefiles in ttk-data that use for loops produce the same result as before, but please double check.

Best
Jonas Lukasczyk


julien-tierny:
thanks for your PR!
I'll try to merge #520 first though, as it fixes the macos ci.
cheers

julien-tierny:
this one looks good to go, thanks!

PR #520, opened 2020-10-27
Collection of fixes

pierre-guillou:
This PR fixes several bugs in TTK. Among them:

* PersistenceDiagramClustering relaunches its computation when DebugLevel or ThreadNumber are modified (following the same approach as in PersistenceDiagram)
* following some tests with Dipha, the BottleneckDistance VTK layer has been cleaned and explicit messages are now displayed in case of errors
* standalones now exit with the value 1 in case of errors and 0 otherwise (they were following the VTK convention before)
* commented code in ttkAlgorithm has been removed
* manual allocations in ttkTriangulationFactory have been boxed into `std::unique_ptr`
* PersistenceDiagram should now work better on an empty input.

No change has been observed when running all ttk-data states.
Enjoy,

Pierre


pierre-guillou:
Also worth mentioning is an Azure Pipelines fix concerning the macOS CI: the ParaView in the pipeline cache in fact depends on the brew version of Python. A recent update of the python formula broke the cached ParaView.
To not depend on a specific version of Python, I added a new cache key that can be manually modified to bypass the cache and relaunch the ParaView compilation.

core/base/contourTreeAlignment/ContourTreeAlignment.cpp
```diff
@@ -1220,9 +1220,8 @@ std::shared_ptr<AlignmentTree> ttk::ContourTreeAlignment::traceAlignmentTree(
   if(t2 == nullptr)
     return traceNullAlignment(t1, true);
 
-  auto id = [this](std::shared_ptr<BinaryTree> t) {
-    return t == nullptr ? 0 : t->id;
-  };
+  auto id
```
julien-tierny, line 1223: @floWetzels 
can you please confirm you're ok with the edits made to this class?

core/vtk/ttkAlgorithm/ttkTriangulationFactory.cpp
```diff
@@ -76,7 +76,11 @@ struct ttkOnDeleteCommand : public vtkCommand {
     if(this->observee)
       this->observee->RemoveObserver(this);
 
-    auto instance = ttkTriangulationFactory::Instance;
+    auto instance = &ttkTriangulationFactory::Instance;
```
julien-tierny, line 79: @JonasLukasczyk 
can you please confirm you're ok with these edits to this class?

examples/vtk-c++/main.cpp
```diff
@@ -40,72 +41,62 @@ int main(int argc, char **argv) {
...
-  vtkSmartPointer<vtkXMLUnstructuredGridWriter> sepWriter
-    = vtkSmartPointer<vtkXMLUnstructuredGridWriter>::New();
+  vtkNew<vtkXMLUnstructuredGridWriter> sepWriter{};
   sepWriter->SetInputConnection(morseSmaleComplex->GetOutputPort(1));
   sepWriter->SetFileName("separatrices.vtu");
   sepWriter->Write();
 
-  vtkSmartPointer<vtkXMLUnstructuredGridWriter> segWriter
-    = vtkSmartPointer<vtkXMLUnstructuredGridWriter>::New();
+  vtkNew<vtkXMLUnstructuredGridWriter> segWriter{};
```
julien-tierny, line 99: simpler, better. thanks!

paraview/RangePolygon/RangePolygon.xml
```diff
@@ -82,60 +78,13 @@ See also ContinuousScatterplot, FiberSurface, Fiber, ProjectionFromField.
...
-         label="Debug Level"
-         command="SetDebugLevel"
-         number_of_elements="1"
-         default_values="3" panel_visibility="advanced">
-        <IntRangeDomain name="range" min="0" max="100" />
-         <Documentation>
-           Debug level.
-         </Documentation>
-      </IntVectorProperty>
+      ${DEBUG_WIDGETS}
```
julien-tierny, line 81: good catch

julien-tierny (commented):
thanks a lot for the PR and the fixes for the CI!

julien-tierny:
can you summarize your edits to the bottleneck distance vtk layer though?

pierre-guillou:
Most of the changes in BottleneckDistance come from moving some method definitions from the header to the .cpp file. The templated methods were also moved to the .cpp, since they were only used there. Apart from that, I removed useless class members, added meaningful error messages (since the module was crashing or exiting directly when I was using it on the diagrams from Dipha). 

I found one typo in the base layer and one bug in comparing the two input diagrams. Apart from that, the base layer has been left untouched and I believe that the VTK layer remains as featureful as before.

julien-tierny:
alright, sounds good.
waiting for jonas and florian's green flags then.
cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

lhofmann:
> Also worth mentioning is an Azure Pipelines fix concerning the macOS CI: the ParaView in the pipeline cache in fact depends on the brew version of Python. A recent update of the python formula broke the cached ParaView.
> To not depend on a specific version of Python, I added a new cache key that can be manually modified to bypass the cache and relaunch the ParaView compilation.

Nice catch! 
The cache will likely also break when any of the other brew packages are updated (which apparently happens frequently).
The following patch would make the cache depend on the versions of all installed brew packages. Could you include that in your PR while you are at it?

```diff
diff --git a/azure-pipelines.yml b/azure-pipelines.yml
index 5838943a..a417d0a8 100644
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -363,11 +363,12 @@ jobs:
   - script: |
       brew cask install xquartz
       brew install wget python libomp mesa glew boost ccache
+      brew list --versions > brew-versions.txt
     displayName: 'Install dependencies'
 
   - task: Cache@2
     inputs:
-      key: ParaView | $(Agent.OS) | "$(PV_VERSION)"
+      key: ParaView | $(Agent.OS) | "$(PV_VERSION)" | brew-versions.txt
       path: ParaView-$(PV_VERSION)/build
       cacheHitVar: CACHE_RESTORED
     displayName: Cache ParaView build
```

pierre-guillou:
I just did! Thanks @lhofmann

JonasLukasczyk:
The changes in TriangulationFactory seem reasonable. To me it looks like Pierre primarily turned the global Instance to a class member.

pierre-guillou:
The global instance is still a global instance but, instead of being a pointer to a heap-allocated object that seems to never be freed, now it is a fully-fledged object. I just removed the indirection.

In a second modification, I wrapped the Triangulation pointers inside the instance into `std::unique_ptr`. This way, it is automatically freed when the pointers go out of scope.

core/base/contourTreeAlignment/ContourTreeAlignment.h
```diff
@@ -298,7 +298,7 @@ int ttk::ContourTreeAlignment::execute(const vector<void *> &scalarsVP,
   }
   std::srand(seed);
   if(alignmenttreeType != lastMatchedValue)
-    std::random_shuffle(permutation.begin(), permutation.end());
+    std::shuffle(permutation.begin(), permutation.end(), std::random_device{});
```
floWetzels: Does this way of shuffling also use the fixed seed from above? (Is it influenced by std::srand(seed)?)

pierre-guillou: Good call @floWetzels! Indeed, std::shuffle uses a more robust random number generator than std::random_shuffle, that's why I decided to replace it.
However, the new random generator doesn't use the seed set by std::srand. I tried several random engines but it appears to have consequences on the `contourTreeAlignment.pvsm` state file in the `ttk-data` repository.
For now, I think the best way is to revert this change and keep the old (but seeded) random generator.
Thanks for pointing that and sorry for the inconvenience.

floWetzels: No problem, maybe we find a better solution soon.
The other changes in the ContourTreeAlignment class seem reasonable to me and should not change anything. So that's fine then.

julien-tierny:
alright, thank you all for your great feedback :)

PR #521, opened 2020-10-28
Fix HDF5 error in macOS ParaView build

pierre-guillou:
Today, an Xcode update in the macOS image in Azure changed the default compiler flags:

> Clang now reports an error when you use a function without an explicit declaration when building C or Objective-C code for macOS (-Werror=implicit-function-declaration flag is on). This additional error detection unifies Clang’s behavior for iOS/tvOS and macOS 64-bit targets for this diagnostic. (49917738)

(from the [Xcode 12 Release Notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-12-release-notes))

This broke the vtkhdf5 build in our CI pipelines.
This PR aims at solving this issue by removing the offended compiler flag.

Enjoy,
Pierre

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -388,6 +388,7 @@ jobs:
                     -DPARAVIEW_BUILD_TESTING=OFF
                     -DPARAVIEW_ENABLE_PYTHON=ON
                     -DVTK_PYTHON_VERSION=3
+                    -DCMAKE_CXX_FLAGS="-Wno-error=implicit-function-declaration"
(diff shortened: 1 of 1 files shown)


pierre-guillou:
Trying to disable the default warning did not work.
New attempt: follow what's recommanded in [this issue](https://github.com/actions/virtual-environments/issues/1712) to switch to an older Xcode version.
Since we don't own the offended code, I'd rather not patch it.

julien-tierny:
awesome pierre, thanks!

PR #522, opened 2020-10-29
Azure Pipeline: Support re-triggering macOS ParaView build

pierre-guillou:
This PR adds a manual key that can be modified to cause a cache miss and force a ParaView rebuild in the macOS CI, following what's suggested in [this Azure documentation](https://docs.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops#can-i-clear-a-cache).

Enjoy,
Pierre


julien-tierny:
right on, let's keep our fingers crossed for the next pr :)

PR #523, opened 2020-10-29
Make ParaView plugin depend on VTK modules

lhofmann:
- VTK modules are now built first and the ParaView plugin only links against them
- CMake option `TTK_BUILD_PARAVIEW_PLUGINS` now depends on `TTK_BUILD_VTK_WRAPPERS`
- this fixes the issue that VTK modules are installed twice when installing the ParaView plugin

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,8 +6,10 @@ if("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}")
   message(FATAL_ERROR "Build in sources is not supported by TTK, please use a separate build folder")
 endif()
 
+include(CMakeDependentOption)
+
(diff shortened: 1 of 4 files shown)


lhofmann:
For some reason, the MacOS pipeline only now correctly fails, because ttkDimensionReduction is not being built, but still has a definition in Compatibility.xml.
I've added an option to disable the compatibility plugin and turned it off by default.

julien-tierny:
actually, we had a similar error at some point yesterday with pierre:
`/Users/runner/work/1/s/ParaView-v5.8.0/build/lib/libvtkRemotingApplication-pv5.8.1.dylib: unknown file type, first eight bytes: 0x00 `
looks like a weird mixture of pv 5.8.0 and 5.8.1.

the workaround we found for the moment was to force a cache miss on pv's build for macos, see #522.

julien-tierny:
alright, looks good!
thanks a lot lutz!

PR #524, opened 2020-10-30
Bypass ParaView cache for macOS, use GitHub Actions generated .pkg

pierre-guillou:
Another attempt at fixing the macOS caching issue in the Azure CI. This time, we bypass entirely the cache system for the ParaView build and we use directly the generated ttk-paraview package from GitHub Actions.

Hope this works.

Enjoy,
Pierre


julien-tierny:
awesome, we finally prevailed over macos' ci.
thanks pierre!

PR #525, opened 2020-11-02
Fix headers installation folders

pierre-guillou:
Since #523, installation of the VTK layer headers has switched from being managed by the ParaView plugin (`paraview_add_plugin`) to the VTK layer (`vtk_wrap_module`). As a consequence, VTK layer headers are dropped directly in `$prefix/include` . This PR moves them into the `$prefix/include/ttk/vtk` subdirectory, mirroring what is done for the base layer headers.

The `$prefix/include/TTKPVTargets.h` header that was generated by `paraview_plugin_build` has been removed, since it is apparently useless.

Enjoy,
Pierre


julien-tierny:
looks good!

PR #526, opened 2020-11-02
TriangulationFactory: Convert cell array to 64 bits storage

pierre-guillou:
This PR aims at fixing #490 by explicitly converting the input cellArray to 64-bit storage prior to generating the explicit triangulation.

Enjoy,
Pierre


julien-tierny:
thanks pierre!

PR #527, opened 2020-11-02
Support seed in IdentifierRandomizer

### LINKED ISSUE #499: Randomizers should be deterministic
I know it sounds strange at first, but I would actually suggest to make the randomization procedures (for example of IdentifierRandomizer) deterministic for a given seed. By varying the seed parameter we get randomness, with the additional advantage that if set once the output stays consistent.

pierre-guillou:
This PR fixes #499 (at least concerning IdentifierRandomizer) by supporting a random seed in the IdentifierRandomizer module.
This seed can be set in ParaView the same way as in the ContourTreeAlignment module.

The randomization algorithm has also been rewritten (by copying what's done in `AbstractMorseSmaleComplex::setFinalSegmentation`). As a consequence, the new algorithm should be two orders of magnitude faster (on my laptop, `states/imageProcessing.psvm` went from 1.5s to 0.02s).

The ttk-data states using this module should now have a deterministic coloration.

Enjoy,
Pierre

### DIFF
--- a/core/vtk/ttkIdentifierRandomizer/ttkIdentifierRandomizer.cpp
+++ b/core/vtk/ttkIdentifierRandomizer/ttkIdentifierRandomizer.cpp
@@ -8,9 +8,6 @@
 #include <ttkMacros.h>
 #include <ttkUtils.h>
 
-using namespace std;
-using namespace ttk;
-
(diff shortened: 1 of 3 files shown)


julien-tierny:
looks good, thanks!

PR #528, opened 2020-11-03
Add Windows & macOS packaging

pierre-guillou:
This PR adds Windows and macOS packaging jobs alongside Ubuntu 18.04 and 20.04 inside the GitHub Actions YAML workflow file. Similarly to Ubuntu, the jobs are divided in two: in a first job, TTK is build upon the ttk-paraview packages and packaged. In a second part, on a new virtual machine, the packages are installed and tested on the examples and the `ttkHelloWorldCmd` standalone.

As described in #308, the `clang-cl` driver was used as a front-end to the Visual C++ compiler to provide the support of (not so) recent OpenMP parallelism out-of-the-box.

To prevent CMake to find wrong OpenMP flags for the Windows and macOS targets, the OpenMP support in the CMake code has been improved following these [Modern CMake guidelines](https://cliutils.gitlab.io/modern-cmake/chapters/packages/OpenMP.html) (CMake >= 3.9).

A similar effort has been done for Python: I switched from [FindPythonLibs](https://cmake.org/cmake/help/latest/module/FindPythonLibs.html) to [FindPython3](https://cmake.org/cmake/help/latest/module/FindPython3.html), available from CMake 3.12 and onwards, which is the current minimum version of CMake supported in TTK. This also helps a lot CMake finding the NumPy headers. This code has been moved into `config.cmake` alongside the other dependencies.

Some `inline` keywords prefixing methods definitions were removed from `ImplicitTriangulation` and `PeriodicImplicitTriangulation` since they caused link errors in macOS: since the inline methods were not used in the compilation unit, the compiler may have discarded them.

On Windows, the destination folder of ttk and ttk-paraview were hard-coded to `TTK` and `TTK-ParaView` (by default, they would have been `TTK 0.9.9` and `TTK-ParaView 5.8.1`). This should make it easier to set environment variables. The ttk-paraview Windows package should be rebuild with the corresponding patch for the examples not to fail though. To support the Windows packaging system, absolute paths in the CMake `install` calls were also replaced by prefix-relative ones.

Compared to the (current) Ubuntu packages, one drawback of the current approach is that we still need to set environment variables. On Windows, TTK and TTK-ParaView are installed in two different prefixes inside `C:\Program Files\`, thus hindering ParaView to find the TTK plugin. On macOS, both are installed under `/Applications/`, which does not seems to be covered by the standard environment variables (PATH, DYLD_LIBRARY_PATH, PYTHONPATH). See the code running the examples in the YAML file for more details about what variable to use.

A second drawback concerns the TTK  and ParaView run-time dependencies. Contrary to Ubuntu, Windows and macOS don't have a centralized packaging system managing the dependencies. On Windows, we rely on conda to provide us all TTK dependencies (except `graphviz`, its headers are missing). On macOS, I tried to use brew and pip (for scikit-learn). Refer to the workflow file to get a list of the dependencies I used.

Enjoy,
Pierre

### DIFF
--- a/core/base/dimensionReduction/CMakeLists.txt
+++ b/core/base/dimensionReduction/CMakeLists.txt
@@ -11,10 +11,10 @@ install(
   FILES
     ${CMAKE_CURRENT_SOURCE_DIR}/dimensionReduction.py
   DESTINATION
-    "${CMAKE_INSTALL_DATAROOTDIR}/ttk"
+    scripts/ttk
   )
 
-target_compile_definitions(dimensionReduction PUBLIC -DTTK_SCRIPTS_PATH=${CMAKE_INSTALL_FULL_DATAROOTDIR}/ttk)
+target_compile_definitions(dimensionReduction PUBLIC -DTTK_SCRIPTS_PATH=scripts/ttk)
 
 if(TTK_ENABLE_SCIKIT_LEARN)
   target_link_libraries(dimensionReduction PRIVATE Python3::Python Python3::NumPy)
--- a/core/base/dimensionReduction/CMakeLists.txt
+++ b/core/base/dimensionReduction/CMakeLists.txt
@@ -1,5 +1,3 @@
-ttk_find_python()
-
 ttk_add_base_library(dimensionReduction
   SOURCES
     DimensionReduction.cpp
--- a/core/base/dimensionReduction/CMakeLists.txt
+++ b/core/base/dimensionReduction/CMakeLists.txt
@@ -17,4 +17,7 @@ install(
   )
 
 target_compile_definitions(dimensionReduction PUBLIC -DTTK_SCRIPTS_PATH=${CMAKE_INSTALL_FULL_DATAROOTDIR}/ttk)
-target_link_libraries(dimensionReduction PRIVATE ${PYTHON_LIBRARY})
+
+if(TTK_ENABLE_SCIKIT_LEARN)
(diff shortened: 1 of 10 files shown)


core/base/dimensionReduction/CMakeLists.txt
```diff
@@ -13,8 +11,11 @@ install(
   FILES
     ${CMAKE_CURRENT_SOURCE_DIR}/dimensionReduction.py
   DESTINATION
-    ${CMAKE_INSTALL_PREFIX}/scripts/ttk
+    "${CMAKE_INSTALL_DATAROOTDIR}/ttk"
```
julien-tierny: I don't believe this is the right place for scripts. on my system, that'd go under `/usr/local/share/scripts/ttk/` although I believe it should go to `/usr/local/scripts/ttk` (like other packages do, e.g. ospray)

julien-tierny:
thanks a lot pierre, this looks fantastic.
I have this one minor comment about the installation path for the dimension reduction scripts.
also, ttk-paraview is currently rebuilding with the new cpack patch (I'll wait until this is done anyways to move forward with the merge)

core/base/dimensionReduction/CMakeLists.txt
pierre-guillou: Done.

julien-tierny:
wonderful, let's go for it :)

PR #529, opened 2020-11-04
Make PersistenceDiagram stateless

### LINKED ISSUE #515: Persistence Diagram Bug with Multiple Executions
Hi all,
@floWetzels and I noticed that ttkPersistenceDiagram behaves strangely if it is tasked to process a vtkMultiBlockDataSet.

Actually, VTK out of the box checks if a filter can consume a vtkMultiBlockDataSet (by explicitly mentioning it in INPUT_REQUIRED_DATA_TYPE), and if yes the pipeline feeds the entire vtkMultiBlockDataSet into the RequestData method. However, if the filter does not support vtkMultiBlockDataSets (as in the case of ttkPersistenceDiagram), the pipeline will actually iterate over the blocks and feed them one by one into the RequestData method, and collect the individual results in a new vtkMultiBlockDataSet. This is intended behavior.

However, ttkPersistenceDiagram seems to have problems with multiple executions and messes up the outputs. My guess is that the root of the problem is the member variable `std::vector<ttk::PersistencePair> CTDiagram_{};`. Is there a need for this member variable? Theoretically the filter should just compute the output based on its input and then forget everything (be stateless). My guess is that the filter tries to cache some information and that messes up multiple executions.

I attached a minimal statefile that requires the ttk-data repository.  To see the bug, press the execute button of the ttkPersistenceDiagram filter. You should see some lines below the diagonal.

Best
Jonas

[diagram.pvsm.zip](https://github.com/topology-tool-kit/ttk/files/5372059/diagram.pvsm.zip)

bug

pierre-guillou:
This PR fixes #515 by making the `PersistenceDiagram` VTK layer stateless. Switching to the "embedded in domain" representation now requires a full recomputation of the diagram.

To still allow users to quickly switch between diagram representations, I hijacked the `ProjectionFromField` module and put some conversion code there. The transformation is still a bit involving since, for the spatial embedding, the diagonal has to be removed and the point coordinates need to be transformed into the Birth and Death arrays. In a symmetrical manner, the back transformation has also been implemented, the state of input diagrams being automatically detected (presence or absence of the Coordinates, Birth and Death point data arrays). 

From the user's point of view, there is a new checkbox inside `ProjectionFromField` that needs to be checked to switch the diagram representation (unchecked by default).

As a bonus, an unused include was removed from `ftrGraph`.

Enjoy,
Pierre

### DIFF
--- a/core/vtk/ttkProjectionFromField/ttkProjectionFromField.h
+++ b/core/vtk/ttkProjectionFromField/ttkProjectionFromField.h
@@ -57,6 +57,9 @@ public:
   vtkSetMacro(UseTextureCoordinates, bool);
   vtkGetMacro(UseTextureCoordinates, bool);
 
+  vtkSetMacro(Use3DCoordinatesArray, bool);
+  vtkGetMacro(Use3DCoordinatesArray, bool);
+
   vtkSetMacro(ProjectPersistenceDiagram, bool);
   vtkGetMacro(ProjectPersistenceDiagram, bool);
 
@@ -93,4 +96,5 @@ protected:
 private:
   bool ProjectPersistenceDiagram{false};
   bool UseTextureCoordinates{false};
+  bool Use3DCoordinatesArray{false};
 };
--- a/core/vtk/ttkProjectionFromField/ttkProjectionFromField.h
+++ b/core/vtk/ttkProjectionFromField/ttkProjectionFromField.h
@@ -68,10 +68,23 @@ protected:
   int FillOutputPortInformation(int port, vtkInformation *info) override;
 
   /**
-   * @brief Generate the spatial embedding of a given Persistence Diagram
+   * @brief Switch a given Persistence Diagram representation
    */
   int projectPersistenceDiagram(vtkUnstructuredGrid *const inputDiagram,
                                 vtkUnstructuredGrid *const outputDiagram);
+  /**
+   * @brief Generate the spatial embedding of a given Persistence Diagram
+   */
+  int projectDiagramInsideDomain(vtkUnstructuredGrid *const inputDiagram,
+                                 vtkUnstructuredGrid *const outputDiagram);
+  /**
+   * @brief Generate the 2D embedding of a given Persistence Diagram
+   */
+  template <typename VTK_TT>
+  int projectDiagramIn2D(vtkUnstructuredGrid *const inputDiagram,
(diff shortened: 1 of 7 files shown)


core/vtk/ttkProjectionFromField/ttkProjectionFromField.h
```diff
@@ -58,20 +57,40 @@ class TTKPROJECTIONFROMFIELD_EXPORT ttkProjectionFromField
...
-  ~ttkProjectionFromField() override;
-
   int FillInputPortInformation(int port, vtkInformation *info) override;
 
   int FillOutputPortInformation(int port, vtkInformation *info) override;
 
+  /**
+   * @brief Switch a given Persistence Diagram representation
+   */
+  int projectPersistenceDiagram(vtkUnstructuredGrid *const inputDiagram,
```
julien-tierny, line 76: I believe the ProjectionFromField filter should be completely agnostic to what it's projecting.
it should have no explicit relation to persistence diagrams at all.

if the users want to project a persistence diagram, I believe the right use case should be:
- the users check a box in ProjectionFromField for using a pre-defined coordinate array ("Use a pre-defined coordinate array")
- a drop-down menu appears, listing all PointData arrays with 3 components
- the users select the array they want (for persistence diagrams, the one named "Coordinates").

then the rest of the code should just modify the coordinates of the point based on that array and that is it.
what do you think?

paraview/ProjectionFromField/ProjectionFromField.xml
```diff
@@ -35,17 +35,31 @@ point-data scalar fields to be used as 2D coordinates.
         </Documentation>
       </InputProperty>
 
+      <IntVectorProperty name="ProjectPersistenceDiagram"
```
julien-tierny: see remark above.

core/vtk/ttkProjectionFromField/ttkProjectionFromField.h
pierre-guillou: In this case, the diagram diagonal would still be present in the spatial embedding and there would not be any `Birth` and `Death` array, confusing the filters that consume Persistence Diagrams (BottleneckDistance, PersistenceDiagramClustering, …)

pierre-guillou:
I added the agnostic way to project any vtkPointSet in space according to a 3-dimensional point data array.

julien-tierny:
alright, thanks!

PR #530, opened 2020-11-04
Cinema Darkroom

JonasLukasczyk:
Hi Julien,
this is the VTK-Only version of Cinema Darkroom. It is contained in one module, does not require any additional dependencies, and can be disabled via the whitelist. You can also find a python state file in a new ttk-data PR that uses all new filters to reproduce an image from the LDAV paper.

Best
Jonas


julien-tierny:
in the paraview xml files, thanks for adding a label for each filter.
for instance going from:
` <SourceProxy name="ttkCinemaDarkroomColorMapping" class="ttkCinemaDarkroomColorMapping">`
to:
` <SourceProxy name="ttkCinemaDarkroomColorMapping" class="ttkCinemaDarkroomColorMapping" label="TTK CinemaDarkroom - ColorMapping">`

JonasLukasczyk:
Hey Julien,
I added labels, paper references, and some documentation. This comes with an updated statefile in ttk-data.

Best
Jonas

JonasLukasczyk:
Hi Julien,
I added a new CD-Camera source that can be conveniently synchronized with the active ParaView Camera. I now showcase this feature also in the ttk-data statefile.

I don't know about the errors in the CI though. It looks like the configuration is failing. Any idea why this is happening?

@CharlesGueunet : The CinemaDarkroomCamera uses VTK::PythonInterpreter to sync filter parameters with the current active camera in ParaView (this way I don't have to include ParaView headers). Does this mean if someone disables pyhton (and therefore this dependency can not be resolved) then non of the Cinema Darkroom filters will be build? They are all sitting in one module and therefore share the dependencies.

Best
Jonas

julien-tierny:
> I don't know about the errors in the CI though. It looks like the configuration is failing. Any idea why this is happening?
it looks like yet another out-of-the-blue python update from azure.
resetting the paraview cache should be the way to go.
cheers,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

JonasLukasczyk:
How can I force a cache update?

pierre-guillou:
Hi Jonas,
you can force a cache update by modifying the key in the `Cache@2` tasks, for instance by concatenating an arbitrary string.
I would suggest to add a version number, like this
```
key: VTK | $(Agent.OS) | "$(VTKVersion)" | "v1"
```
This way, the next user blocked by an Azure VM update can increment the version number to invalidate the current cache.

julien-tierny:
the ci has been updated but still fails for the windows-vtk configuration.
@pierre-guillou  suggests it might be an issue with boost's support under azure's windows vm (a simple boost upgrade would do the trick).
however, the ci fails at ttk's configure step claiming that it cannot find paraview (which makes sense, since paraview does not need to be installed in the vtk-only configuration). I'm afraid there's another issue in handling paraview/vtk at the cmake level.
@CharlesGueunet, any idea?
@JonasLukasczyk  did you do anything specific at the cmake level with vtk/paraview?

julien-tierny:
alright, pierre was right :) 
I misread the log. that was indeed a simple boost update in azure.
cheers

JonasLukasczyk:
Hi Julien,
I added your feature request (transparent backgrounds), and renamed the filters in ParaView to `TTK Darkroom X` (they are still in the Cinema folder). I also updated the state file in the ttk-data PR. Note that you can save a rendered image by selecting an output port of a Darkroom filter and then use `File/Save Data` to select the PNG writer.

Cheers
Jonas

julien-tierny:
awesome, thank you!

PR #531, opened 2020-11-05
Fix handling of eigen dependency in HarmonicField and Laplacian base modules

pierre-guillou:
Following Christopher Kappe's email on the ttk-dev mailing list, here are the suggested changes. This should allow a user to prevent building the HarmonicField and Laplacian modules even though eigen is present on its system. The `TTK_ENABLE_EIGEN` CMake variable is set in the config.cmake file.

The EigenField module is not concerned by this, as it already depends on `TTK_ENABLE_SPECTRA`.

Enjoy,
Pierre

### DIFF
--- a/core/base/harmonicField/CMakeLists.txt
+++ b/core/base/harmonicField/CMakeLists.txt
@@ -9,7 +9,7 @@ ttk_add_base_library(harmonicField
     triangulation
     )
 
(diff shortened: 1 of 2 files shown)


julien-tierny:
good catch, thanks @kappe-c @pierre-guillou

PR #533, opened 2020-11-15
Cinema Imaging Backend

JonasLukasczyk:
Hi Julien,
this PR adds two additional rendering backends to CinemaImaging (embree and a native raycaster), and updates some Cinema Darkroom filters to be more user-friendly. This PR comes therefore with a drastically simplified cinemaImaging.py state file in ttk-data (see new PR).

Best
Jonas


JonasLukasczyk:
Hi @julien-tierny ,
the Ubuntu config step fails again and I don't know why. In my last commit I appended some text to refresh the cache, but that didn't seem to work. Any ideas?

-Jonas

JonasLukasczyk:
Thx @julien-tierny, problem solved! Did you have a chance to run the new state file?

-Jonas

julien-tierny:
hi jonas,
thanks for the reminder.
can you confirm all the state files are running properly?
cheers

JonasLukasczyk:
Hi Julien,
the old state files run without errors or warnings, but I just saw that the geometry approximation demo has some additional artifacts. I will have a closer look.

-Jonas

JonasLukasczyk:
Hi Julien,
I updated the DepthImageBasedGeometryApproximation to the newest version of the tukl branch. I was planing on doing this next, but it turned out it was already necessary for the current cinema imaging code. Now all state files work and produce the same output as before.

Best
Jonas

julien-tierny:
thanks jonas!
can you try to reset azure's cache (see a5efa9a)
cheers

JonasLukasczyk:
I reset the cache. Btw. what is the point of having a cache if it needs to be reset all the time?

JonasLukasczyk:
Now MacOS can not find the correct python version, arg!

JonasLukasczyk:
@julien-tierny looks like everything is now ready to go.

core/base/cinemaImaging/BVH.h
```diff
@@ -0,0 +1,385 @@
+/// \ingroup base
+/// \class ttk::BVH
+/// \author Rosty Hnatyshyn <rostyslav.hnatyshyn@gmail.com>
```
julien-tierny, line 3: @JonasLukasczyk 
did you take that code from another implementation?

JonasLukasczyk: Rosty is one of my students. The code base does not contain third-party code.

julien-tierny: ok. from ttk's github repository's point of view, you'll be the author of that commit and as far as I'm concerned, its maintainer :) (this means that in case of issue with that code, I'll get back to you)
you might want to consider having that student authoring these commits himself if you want him to be the entry point for maintenance.

julien-tierny: another note: bounding volume hierarchies can be useful for other ttk core components (fiber surfaces could in principle exploit these). I think it would make sense to move that to an independent base code repository `core/base/boundingVolumeHierarchy/boundingVolumeHierarchy.h`

core/base/cinemaImaging/CinemaImaging.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+void ttk::CinemaImaging::subVectors(float *out,
+                                    const float *a,
+                                    const float *b) {
+  out[0] = b[0] - a[0];
+  out[1] = b[1] - a[1];
+  out[2] = b[2] - a[2];
+};
+
+void ttk::CinemaImaging::cross(float *out, const float *a, const float *b) {
```
julien-tierny, line 27: low level geometrical operations should be placed under `core/base/geometry/Geometry.h`.
note that these operations are likely to be available there already (templatized implementations) 
that'd make sense to use them instead (and complement them if features are missing.)

JonasLukasczyk: It is OK if I'm the entry point for now, but Rosty wrote the entire BVH stuff and native renderer so I want to keep him as the author. I merged the cinema imaging code with all the other stuff (embree backend, vtk layer, etc) and "polished" everything. Separating these now again and opening separate PRs is overkill, so I would argue against it. I also agree that rudimentary functions should be moved to a more general class (or use them if they already exist). I would prefer if you merge this in as is and then we can open an issue requesting that, then I can tell Rosty to work on this.

julien-tierny:
alright, based on our discussion, I'll merge this right now and count on your feedback within the next few days regarding the low-level improvements I mentioned above.

PR #534, opened 2020-11-16
Embed Visual C++ redistribuable in Windows installer

pierre-guillou:
This PR improves the Windows packaging by downloading the Visual C++ redistribuable (the Windows C++ standard library) and embeds it into the Windows package. The redistribuable installer is automatically and silently run during the TTK installation. The installer is removed afterwards.

Note that the redistribuable won't be uninstalled when TTK uninstaller is run.

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -307,6 +307,12 @@ jobs:
       run: |
         ttk-paraview.exe /S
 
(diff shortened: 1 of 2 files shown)


julien-tierny:
awesome, thanks!

PR #535, opened 2020-11-18
Use Actions/Tasks to force Python 3.8 on macOS

pierre-guillou:
Following the brew upgrade of Python (3.8->3.9), this PR forces the Python version on macOS to 3.8, since there seems to be an incompatibility between Python 3.9 and CMake.

Following [the update notice](https://github.com/actions/virtual-environments/issues/1929), I used the `setup-python` action for GitHub Actions and the `UsePythonVersion` Task for Azure Pipelines.

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -167,10 +167,14 @@ jobs:
       run: |
         # ParaView dependencies
         brew cask install xquartz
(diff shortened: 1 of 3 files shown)


julien-tierny:
thanks pierre!

PR #536, opened 2020-11-18
Fix Python 3.8 dependencies for macOS packaging

pierre-guillou:
This PR completes #535  by fixing the way CMake finds Python and NumPy in GitHub Actions.

Enjoy,
Pierre


julien-tierny:
thanks!

PR #537, opened 2020-11-20
Packaging: Add tag name to release assets

pierre-guillou:
This PR modifies the GitHub Action uploading a release asset to include the tag name (without the `refs/tags/`) prefix in the package names. To do so, I switched from actions/upload-release-asset to  svenstaro/upload-release-action, which is more convenient.

Some seemingly useless environment variables for the Windows test job have been removed.

Enjoy,
Pierre


julien-tierny:
awesome, thanks!

PR #539, opened 2020-11-24
New GitHub Actions workflow to perform basic checks on the code

pierre-guillou:
This PR adds a new GitHub Action workflow. This new workflow performs some basic checks on the repository:
* test if the code compiles without warning with clang and the default compiler flags, enforced through `-Werror`;
* test if the source code has been formatted with clang-format (disabled since clang-format 10 and clang-format 11 seems to differ on several files);
* use clang-tidy to perform a long list of interesting C++ related checks, controlled with a `.clang-tidy` file at the project root;
* other checks that detect Unicode characters or DOS endlines, empty files, files that don't end with a newline character, files whose names that don't exhibit a "standard" C++ extension (in the TTK sense: .cpp and .h).

Those last checks are only informational, contrary to the first one.

Similarly to the packaging workflow, this workflow can be launched by pushing tags matchings with "v*" and "dev*". Additionally, it will be executed on every pull request.

Some preliminary findings of those checks:
* there are some empty source file in TTK;
* `AuctionActor.cpp` and `Munkres.cpp` are left out of the build system and are not build with TTK.

To avoid confusion, the packaging CI has been renamed from "CI" to "packaging".

As a bonus, the embree dependency has been added to the packaging CI but only for macOS and Windows, since it is missing on Ubuntu 18.04 and badly packaged on Ubuntu 20.04 (missing CMake target file).

Enjoy,
Pierre


julien-tierny:
thanks!
@JonasLukasczyk : now that  `core/vtk/ttkCinemaDarkroom/ttkCinemaDarkroomColorMaps.cxx` is protected from clang-format, maybe it would make sense to change its extension to `.cpp` or `.inl`.
what do you think?

JonasLukasczyk:
> thanks!
> @JonasLukasczyk : now that `core/vtk/ttkCinemaDarkroom/ttkCinemaDarkroomColorMaps.cxx` is protected from clang-format, maybe it would make sense to change its extension to `.cpp` or `.inl`.
> what do you think?

I don't mind if we rename the file. I'm in favor of cpp.

julien-tierny:
thanks, done!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #540, opened 2020-11-24
Support setting environment variables in Windows packages

pierre-guillou:
This PR introduces a Powershell script used in the packaging step for setting environment variables when installing Windows packages. The script is also called at uninstall to remove those environment variables.

Additionally, a clang-cl warning about unused parameters cluttering the Windows packaging CI has been disabled.

Enjoy,
Pierre

### DIFF
--- a/CMake/CompilerFlags.cmake
+++ b/CMake/CompilerFlags.cmake
@@ -23,6 +23,13 @@ if (NOT MSVC) # GCC and Clang
       )
   endif()
 
+elseif(${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" AND "x${CMAKE_CXX_SIMULATE_ID}" STREQUAL "xMSVC")
+  # clang-cl on Windows
+  # c.f. https://stackoverflow.com/questions/50857779/cmake-detects-clang-cl-as-clang
+
+  # warning flags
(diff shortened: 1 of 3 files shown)


julien-tierny:
hi pierre,

thanks a lot for the PR.

it seems that there is a couple of assumptions here:
- that the user selected to install things in general under `C:\Program Files`
- that the user performed a system-wide install of anaconda (again under `C:\ProgramData\`)

am I right?
if so, this may be too restrictive.

pierre-guillou:
I can try to pass the installation directory from the TTK NSIS installer to the Powershell script, but I think it's quite harder to detect the Anaconda installation directory with this installer (which, in fact, is installed system-wide in `C:\ProgramData\Anaconda3`).

julien-tierny:
would it be possible to ask in the nsis installer where anaconda has been installed?

julien-tierny:
(and provide a default value to `C:\ProgramData\Anaconda3`)

pierre-guillou:
Well, I've already tried to modify the CMake-provided NSIS template to set environment variables but I'm afraid that won't be a very robust solution.
But I can try to extend the Powershell script to detect the Anaconda installation automatically. I think this should be way easier.

Note that the Powershell script also makes assumptions on the installation directory of TTK-ParaView.

julien-tierny:
alright, looks good!
thanks

Issue #541, opened 2020-12-07
TTKTopologicalSimplification: LTS unable to converge


lhofmann:
**Describe the bug**
For some datasets, TTKTopologicalSimplification fails to converge with LTS enabled, but works fine without LTS.

```
[LTS] Allocating Memory ....................................... [0.021s|8T|100%]
[LTS] ----------- [Removing unauthorized minima] -------------------------------
[LTS] Inverting Array ......................................... [0.001s|8T|100%]
[LTS] Initializing Memory ..................................... [0.001s|8T|100%]
[LTS] Initializing Propagations (3968|2000000) ................ [0.009s|8T|100%]
[LTS] Computing Propagations (3968) ........................... [0.160s|8T|100%]
[LTS] Finalizing Propagations (1589|0.40|0.82) ................ [0.000s|8T|100%]
[LTS] Computing segments (1589) ............................... [0.041s|8T|100%]
[LTS] Computing Local Order of Segments (1589) .................. [0.000s|8T|0%]
[LTS] [ERROR] Unable to converge within 20 iterations!
[TopologicalSimplification] [ERROR] TopologicalSimplification.execute() error code: 1
```

**To Reproduce**
Example dataset and statefile: [lts.tar.gz](https://github.com/topology-tool-kit/ttk/files/5652310/lts.tar.gz)

**Expected behavior**
While I'm going to admit that I haven't read the paper, I was under the impression, that LTS would yield results equivalent to the sequential algorithm. 
Especially since it is enabled by default, I would expect the same results for TTKTopologicalSimplification with and without LTS.

**System (please complete the following information):**
 - OS: Linux
 - Compiler: gcc 10.2
 - TTK built from dev branch

@JonasLukasczyk

JonasLukasczyk:
Yes, that doesn't mean there are no bugs in the implementation ;-)

I will have a look.

JonasLukasczyk:
You mentioned multiple datasets, can you provide some more?

JonasLukasczyk:
@lhofmann Solved: fortunately it was nothing serious. This dataset just requires 22 local iterations (for a large, heavily malformed region) instead of the fixed 20 ;-). I put this limit in a while ago for debugging purposes, since all tested datasets converged with at most 5 iterations, and when it took longer this usually indicated a bug.

I will open a PR where I remove this limit but maybe print it as a warning to notify the user the algorithm has to deal with a big job (which is the worst case for LTS).

[...]

PR #542, opened 2020-12-07
Removed fixed iteration limit from LTS procedure

JonasLukasczyk:
Hi Julien,
this is the oneliner fix for issue #541. I did not run all state files again, since I just removed the limit that is not reached for these states anyway.

Best
Jonas


julien-tierny:
alright, thanks!

PR #543, opened 2020-12-07
Ci refactor

rostyhn:
Hi @julien-tierny,

I am Rosty Hnatyshyn, a student from @JonasLukasczyk. He asked me to refactor some cinemaImaging code before pushing it to the main branch of ttk.

*I moved the BVH to its own folder that should allow others to use it in their code.
*I added a subtract vectors method in Geometry.h (It is used in calculating triangle-ray intersections)

Thanks,
Rosty


rostyhn:
@JonasLukasczyk also suggested that I test the geometryApproximation.pvsm in ttk-data to verify that the results are similar and they are. They are initially set to render with the opengl backend, so opening them isn't a problem.

julien-tierny:
Looks awesome, thanks rosty!

PR #545, opened 2020-12-17
Fix Azure CI failing on Python dependency when building examples

pierre-guillou:
This PR simplifies the Python dependency management to remove the special handling on macOS.
The shared libraries RPATH has also been fixed.
This should make the Azure CI work again.

Enjoy,
Pierre


julien-tierny:
awesome, thanks!

PR #547, opened 2020-12-20
Fix MSVC std::min overload error

### LINKED ISSUE #546: Error (MSVC C2672, C2780, C2784) for std::min
**What**
std::min is called with `int`, `ttk::SimplexId` and this throws an error when building TTK with `MSVC 19.28.29334`.

**Where**
[Here](https://github.com/topology-tool-kit/ttk/blob/dev/core/base/skeleton/ZeroSkeleton.cpp#L548), [here](https://github.com/topology-tool-kit/ttk/blob/dev/core/base/skeleton/OneSkeleton.cpp#L144), [here](https://github.com/topology-tool-kit/ttk/blob/dev/core/base/skeleton/TwoSkeleton.cpp#L312) and [here](https://github.com/topology-tool-kit/ttk/blob/dev/core/base/skeleton/ThreeSkeleton.cpp#L147)

**How**
```
On Windows, MSVC 19.28.29334
git checkout 0.9.9 
(or)
git checkout dev
mkdir build
cd build
cmake .. -DTTK_BUILD_PARAVIEW_PLUGINS=OFF -DTTK_BUILD_VTK_WRAPPERS=OFF -G Ninja
ninja
```

**Screenshots**
image

 - OS: [Win 10]
 - Compiler [MSVC  19.28.29334 with VS2019 Community]
 - Installed TTK's optional dependencies: NONE.

**Additional context**
Oddly enough, this error appears exclusively on windows. On Linux (`g++ 9.3.0` and `clang++ 10.0.0`), it's a breeze.
This looks like a very simple fix. I'll make a PR soon. Just reporting it to make sure this is not isolated to my environment and hopefully someone can confirm it.

jspanchu:
- Handle multiple template types for std::min
 without a static_cast.
- Call it with wider type ttk::SimplexId
- Contribution `std::min(int, ttk::SimplexId)` -> `std::min<ttk::SimplexId>(a, b)`

Resolves #546

### DIFF
--- a/core/base/skeleton/OneSkeleton.cpp
+++ b/core/base/skeleton/OneSkeleton.cpp
@@ -141,7 +141,7 @@ int OneSkeleton::buildEdgeList(
   printMsg("Building edges", 0, 0, 1, ttk::debug::LineMode::REPLACE);
 
   const SimplexId cellNumber = cellArray.getNbCells();
-  const int timeBuckets = std::min(10, cellNumber);
+  const int timeBuckets = std::min<ttk::SimplexId>(10, cellNumber);
 
 #ifdef TTK_ENABLE_OPENMP
 #pragma omp parallel for num_threads(threadNumber_)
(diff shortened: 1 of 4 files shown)


julien-tierny:
thanks for the quick fix!
cheers

PR #548, opened 2021-01-05
Always add the plugin destination to the RPATH

CharlesGueunet:
Dear Julien,

First of all, happy new year to you and all the TTK team :)
This small PR allows the TTK plugin to be installed in a custom location without having to tweak the `LD_LIBRARY_PATH` by telling cmake to have a look in the right folder during the install step.

The plugin subdir is added to the **RPATH** only once, even if cmake is run multiple time.

Best,
Charles


config.cmake
```diff
@@ -322,7 +322,7 @@ endif()
 if(NOT DEFINED CMAKE_BUILD_WITH_INSTALL_RPATH)
   set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
 endif()
-if(NOT CMAKE_INSTALL_RPATH)
+if(NOT DEFINED CMAKE_INSTALL_RPATH)
   set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
```
vollia, line 326: Are we sure the install is always in **lib** and not in **lib64** in some systems ? Maybe we could use the `CMAKE_INSTALL_LIBDIR` here'

julien-tierny: good catch. @pierre-guillou what's your take? I remember that's been a difficulty at some point for the deb packages.

pierre-guillou: Hi Charles, Julien,

If I remember correctly, I modified 3 weeks ago the libraries RPATH (which was hard-coded to `${CMAKE_INSTALL_PREFIX}/lib/ttk/`) to `${CMAKE_INSTALL_PREFIX}/lib` because I forgot it when moving the TTK libraries default install folder.
I have not (yet) build TTK for a system with `lib64` separated from `lib` but using `CMAKE_INSTALL_LIBDIR` should do the trick.
Pierre

lhofmann: +1 for using `CMAKE_INSTALL_XYZDIR` everywhere.
For example, the conda build requires non-standard paths such as `
    -DCMAKE_INSTALL_LIBDIR="Library/lib"
    -DCMAKE_INSTALL_BINDIR="Library/bin"
    -DCMAKE_INSTALL_INCLUDEDIR="Library/include" ` etc.

There is one more place with a hard-coded path here:
```patch
diff --git a/CMake/BaseCode.cmake b/CMake/BaseCode.cmake
index 01c94cf9..4b4dc8a5 100644
--- a/CMake/BaseCode.cmake
+++ b/CMake/BaseCode.cmake
@@ -47,7 +47,7 @@ function(ttk_add_base_library library)
     )
 
   if(ARG_HEADERS)
-    install(FILES ${ARG_HEADERS} DESTINATION include/ttk/base)
+    install(FILES ${ARG_HEADERS} DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/ttk/base")
   endif()
 endfunction()
```

dsakurai:
Shouldn't this be done per-library /per-target basis?
Depending on the global `CMAKE_INSTALL_RPATH` may conflict with thirdparty apps / libs.

Good work, anyway!

julien-tierny:
alright, merging charles's PR.
I'll process the `CMAKE_INSTALL_XYZDIR` remark in another commit.
thanks all for the discussion :) 
cheers

CharlesGueunet:
Thanks @julien-tierny 
FYI I had started to implement the changes suggested by Lutz. It worked well on my side but I did not try from a clean compilation. Anyway, here is the patch I had:

```diff
diff --git a/CMake/BaseCode.cmake b/CMake/BaseCode.cmake
index 01c94cf9..ed888448 100644
--- a/CMake/BaseCode.cmake
+++ b/CMake/BaseCode.cmake
@@ -25,7 +25,7 @@ function(ttk_add_base_library library)
   target_include_directories(${library}
     PUBLIC
       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
-      $<INSTALL_INTERFACE:include/ttk/base>
+      $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/ttk/base>
     )
 
   if(ARG_DEPENDS)
@@ -47,7 +47,7 @@ function(ttk_add_base_library library)
     )
 
   if(ARG_HEADERS)
-    install(FILES ${ARG_HEADERS} DESTINATION include/ttk/base)
+    install(FILES ${ARG_HEADERS} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/ttk/base)
   endif()
 endfunction()
```
```diff
diff --git a/config.cmake b/config.cmake
index 780dacbc..dec657b7 100644
--- a/config.cmake
+++ b/config.cmake
@@ -323,7 +323,7 @@ if(NOT DEFINED CMAKE_BUILD_WITH_INSTALL_RPATH)
   set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
 endif()
 if(NOT DEFINED CMAKE_INSTALL_RPATH)
-  set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
+  set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
 endif()
 if(NOT DEFINED CMAKE_INSTALL_RPATH_USE_LINK_PATH)
   set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
```

julien-tierny:
these should have already been addressed with the commits c0851610fe45b5acbc4fefbe8f6c8a0bcded3a85 and d8649e1cd8667bf3fff623c93074650907c61ff1.
I have just pushed a quick fix with 9811e53a69e6fda8747fef41fc9965287587e2a9.
thanks for double-checking it's ok.
cheers

CharlesGueunet:
Nice ! It seems all good

PR #550, opened 2021-01-18
Azure Pipelines : Fix macOS ParaView/Python 3.9 compatibility

pierre-guillou:
Once finished, this PR should fix the Azure Pipelines CI on macOS.

Enjoy,
Pierre

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -372,9 +372,13 @@ jobs:
       versionSpec: '3.8'
 
   - script: |
-      wget https://github.com/topology-tool-kit/ttk-paraview/releases/download/$(PV_VERSION)/ttk-paraview.pkg
-      sudo installer -pkg ttk-paraview.pkg -target /
-    displayName: 'Download and install ParaView'
+      wget https://www.paraview.org/files/v$(PV_V)/ParaView-$(PV_VERSION).tar.xz
+      tar Jxvf ParaView-v5.8.1.tar.xz
+      cd paraview/patch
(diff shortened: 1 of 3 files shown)


pierre-guillou:
So, after several tries, I stumbled upon an issue preventing the caching of the ParaView install folder. This issue is twofold:
* VTK from ParaView 5.8 uses a hard-coded path to the Python headers directory in the generated VTK-targets.cmake file (target `VTK::Python`, property `INTERFACE_INCLUDE_DIRECTORIES`). This does not seems to be the case with ParaView 5.9.
* Homebrew Python headers are not located in a fixed directory: sometimes, a parent directory might be renamed (randomly? depending on the virtual machine?).

Those two issues prevent the use of the Pipeline cache system to cache the ParaView installation folder because if the hard-coded path to the Python headers in VTK and their real path in the system mismatch, TTK fails to configure.

For now, I have removed this cache task. We still have the secondary cache with ccache that can fasten the whole build time, sometimes as fast as using the installation folder cache.

To sum up, this PR
* adds the support of Python 3.9 to ParaView 5.8 through the TTK patches
* reworks the macOS Azure CI to build ParaView 5.8 from scratch before TTK (in the same VM) in order to access the same (Homebrew) Python
* uses Ninja instead of Make in the macOS Azure CI
* enable the ttkHelloWorldCmd standalone test in the macOS Azure CI

Hope this will work fine this time.
Enjoy,
Pierre

julien-tierny:
alright, looks good. thanks!

PR #551, opened 2021-01-18
Extend GitHub Action check workflow

pierre-guillou:
This PR extends the GitHub Actions check workflow by refactoring the clang-check warning job using a job matrix. Now this job will be executed with and without `TTK_ENABLE_KAMIKAZE` and `TTK_ENABLE_OPENMP` (4 configurations, up to 256). As a remainder, this job will detect and fail if new (clang) compiler warnings are inserted in new PRs.

This PR also fixes other non-C++ warnings detected by the same workflow. Among them:
* files ending with DOS newlines (`\r\n`) have been converted to Unix `\n` newlines with the dos2unix tool to be consistent with the rest of the codebase. Git on Windows should automatically convert Unix newlines to DOS at checkout and DOS newlines to Unix at commit.
* two unused source files (empty and not referenced by the CMakeLists.txt file in the same directory) were removed.
* some UTF-8 characters in the source files (mostly in comments) were replaced with ASCII counterparts.

Apart for line endings, no proper C++ code has been modified by this PR.

In a future PR, I plan to fix some of the clang-tidy warnings (performance and bug-prone constructs in the C++ source code).

Enjoy,
Pierre


julien-tierny:
could you sync that one with PR #550 to reset the macos ci?

julien-tierny:
looks good, thanks!

PR #553, opened 2021-01-20
Remove unused cmake variable.

jspanchu:
This concerns target `ttkAll`.
 
`${VTKWRAPPER}` is neither set nor used anywhere else in the build system. Perhaps it was leftover from previous commits :smile:


julien-tierny:
good catch, thanks!

PR #555, opened 2021-01-27
Enforce clang-format code formatting with GitHub Actions

pierre-guillou:
This PR uses the latest clang-format (v11), now available on Ubuntu 20.04 in a separate package, to check source code formatting.
This check has been moved in a specific job that fails if the source code in not properly formatted (using this particular version of clang-format).

In case this check proves to fail too often, it can be disabled by removing the `-Werror` argument in the command line.

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -23,9 +23,11 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


julien-tierny:
ok, looks good. thanks!

PR #557, opened 2021-01-30
API Migration: Remove remains of the old API

pierre-guillou:
This PR removes what's remains of the old Debug and Triangulation API in order to bring a final end to the Great TTK Module Migration of 2020.

* The old `Debug::dMsg` methods were removed. Uses of these methods were converted to the new `Debug::printMsg` API
* The `ttkTriangulationAlgorithm` module has been removed, now superseded by `ttkAlgorithm`
* The `Blank` module has also been removed to give place to `HelloWorld`.

No change has been observed while running the ttk-data states nor during the execution of the `ttkGeometrySmootherCmd` standalone.

Enjoy,
Pierre


julien-tierny:
ok, thanks!

PR #558, opened 2021-01-30
Collection of fixes

pierre-guillou:
This PR fixes a collection of bugs I have encountered in the last trimester. Among them:
* use the new `brew install --cask` syntax in the macOS CI pipelines
* refactor the `PSORT` macro in order to make it independent of `BaseClass` derivatives by making the number of required threads a parameter
* removing empty source code files, making the corresponding base modules `ttk_add_base_template_library`
* `makeUnion` is now a member of `UnionFind` instead of a free function
* more explicit log messages for `PersistenceDiagramClustering` (distance per pair type) and `DimensionReduction`
* merging two if statements with the same body in `PeriodicImplicitGrid`
* reducing the number of clang-tidy warnings in shared C++ headers for the GitHub Action lint-code job
  + removing superfluous `std::move`
  + declaring move constructors/assignment operators `noexcept`
  + passing objects owning heap-allocated resources by reference instead of by value (`std::string`, functors…)
  + all the above reduce the clang-tidy job from 5,7k lines to 2,2k lines of warnings
* other potential segfaults and memory leaks fixes.

No change has been observed when running the ttk-data states.

Enjoy,
Pierre

### DIFF
--- a/core/vtk/ttkTopologicalSimplification/ttkTopologicalSimplification.cpp
+++ b/core/vtk/ttkTopologicalSimplification/ttkTopologicalSimplification.cpp
@@ -141,16 +141,16 @@ int ttkTopologicalSimplification::RequestData(
 
   int ret{};
   if(this->UseLTS) {
-    auto lts = new ttk::LocalizedTopologicalSimplification();
-    lts->setDebugLevel(this->debugLevel_);
-    lts->setThreadNumber(this->threadNumber_);
+    ttk::LocalizedTopologicalSimplification lts{};
+    lts.setDebugLevel(this->debugLevel_);
+    lts.setThreadNumber(this->threadNumber_);
 
-    lts->preconditionTriangulation(triangulation);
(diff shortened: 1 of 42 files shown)


core/vtk/ttkTopologicalSimplification/ttkTopologicalSimplification.cpp
```diff
@@ -141,16 +141,16 @@ int ttkTopologicalSimplification::RequestData(
 
   int ret{};
   if(this->UseLTS) {
-    auto lts = new ttk::LocalizedTopologicalSimplification();
-    lts->setDebugLevel(this->debugLevel_);
-    lts->setThreadNumber(this->threadNumber_);
+    ttk::LocalizedTopologicalSimplification lts{};
```
julien-tierny, line 144: nice :)

julien-tierny:
ok, looks good. thanks

PR #556, opened 2021-01-29
Fix TTK compilation with ENABLE_64BIT_IDS=ON

pierre-guillou:
This PR fixes the TTK build with `TTK_ENABLE_64BIT_IDS=ON`. Mainly a matter of switching `std::vector<int>` to `std::vector<SimplexId>` and using `std::abs` instead of the non-templated `abs` alternative.

This should fix #554.

A new warning check configuration has been added to the GitHub Actions CI to make sure that TTK stays able to be built in this configuration without errors.

No change has been observed in ttk-data with the default `TTK_ENABLE_64BIT_IDS=OFF`. I have not tested the states with it enabled though (as I suspect this would cause some crashes).

Enjoy,
Pierre

Issue #337, opened 2020-03-04
Trouble running c++ example with 64bit on


dbanesh:
I'm having trouble running the stable branch of ttk_master with the 64 bit marker turned on for a very large data set. I've also had issues trying to run older version of ttk with 64 bit on (for the same data), but considering that master is the latest, the issue I describe will be based on the output I got with this version. 

0. (On a linux system)
1. I am not compiling TTK with Paraview or VTK, this is for the standalone c++ compilation with 64Bit ON. 
2. Once compiled. I am able to compile and make the c++ example (in examples/c++). I update the 'ints' in the main.cpp file to 'long long int' to support the larger data. Otherwise, the code is 'killed' almost immediately. No other changes are made to the code.
3. The code enters the block for 'Persistence Curve' and breaks. The outputs ttk generates are:

[Common]  _____ _____ _  __                       __  __    ____   ___  _  ___
[Common] |_   _|_   _| |/ /                      / /__\ \  |___ \ / _ \/ |/ _ \
[Common]   | |   | | | ' /                      | |/ __| |   __) | | | | | (_) |
[Common]   | |   | | | . \                      | | (__| |  / __/| |_| | |\__, |
[Common]   |_|   |_| |_|\_\                     | |\___| | |_____|\___/|_|  /_/
[Common]                                         \_\  /_/
[Common] Welcome!
[main::load] Reading input mesh...
[main::load]   done! (read 419430400 vertices, 838778882 triangles)
[OneSkeleton] Edge-list built in 122.062 s. (1258209281 edges, 1 thread(s)).
[ZeroSkeleton] One-skeleton built in 271.815 s. (56 thread(s)).
Killed

**Expected behavior**
I expect the sample code to run to completion.

**Screenshots**
<img width="574" alt="Screen Shot 2020-03-03 at 5 39 42 PM" src="https://user-images.githubusercontent.com/19173085/75900522-4475cf00-5dfa-11ea-9f61-93f1add79c08.png">

**System (please complete the following information):**
 - OS: Linux
 - TTK built with cmake 3.15.3, boot 1.70.0 and gcc 9.2.0
 - No additional TTK optional dependencies installed. All Paraview plugins, standalone apps, vtk wrappers turned off.

julien-tierny:
PR #560 should have fixed the issue.

PR #560, opened 2021-02-02
Fix the VTK array type for 64bit SimplexIds

pierre-guillou:
This PR alters the type of the `ttkSimplexIdTypeArray` from `vtkLongLongArray` to `vtkIdTypeArray` in the 64bit Ids mode.

Some TTK filters compare the offset array type to either `VTK_INT` (when SimplexId == int) or `VTK_ID_TYPE` (when SimplexId == long long): c.f. `ttkPersistenceDiagram.cpp`, `ttkMorseSmaleComplex.cpp` or `ttkTopologicalSimplification.cpp`. If the comparison fails, the filter aborts.

Since `vtkLongLongArray` datatype does not match `VTK_ID_TYPE` but `VTK_LONG_LONG`, those filters always failed in the 64bit Ids mode. The current PR fixes this issue.

I found this issue when testing the TTK examples in the 64bit ids mode after #556. Now, all 4 examples (C++, VTK-C++, Python and pvpython) seems to work correctly and output the correct Morse-Smale segmentation & separatrices. As a consequence, I believe that #337 can now be closed safely.

Enjoy,
Pierre

### DIFF
--- a/core/vtk/ttkAlgorithm/ttkMacros.h
+++ b/core/vtk/ttkAlgorithm/ttkMacros.h
@@ -1,12 +1,12 @@
 #pragma once
 
+#include <vtkIdTypeArray.h>
 #include <vtkIntArray.h>
-#include <vtkLongLongArray.h>
 
 #define TTK_COMMA ,
 
 #ifdef TTK_ENABLE_64BIT_IDS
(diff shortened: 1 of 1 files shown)


julien-tierny:
ok, looks good. 
thanks

PR #561, opened 2021-02-03
MeshGraph with new CellRepresentation

JonasLukasczyk:
Hi Julien,
this PR updates the MeshGraph module to work with the new connectivity/offset format of vtkCells.

It also uses some new Macros that are currently only defined in the MeshGraph.cpp file, but I think these macros have an advantage over the macros we currently use since they mitigate template code explosion. In short, they work as the old ones, but you can limit the supported types per template dimension. So for instance the ttkTypeMacroARI macro expects three template parameters, where the first can be (A)ny type, the second must be a (R)eal, and the third an (I)nteger. This makes sense since coordinates of vtkPoints can only be stored as doubles or floats, whereas connectivity arrays must be integers (actually only int or long). This reduces the number of instances from 14x14x14 to 14x2x2. Lets see what you and @pierre-guillou have to say about this, but I think these macros should make it to the ttkMacros.h file and we should refactor some modules to get rid of template explosions where possible. A build with debug symbols already uses 16GB ram, this getting out of hand...

I will also open a new ttk-data PR that replaces both files that use the MeshGraph module (they still produce the same result).

Best
Jonas

### DIFF
--- a/core/base/meshGraph/MeshGraph.h
+++ b/core/base/meshGraph/MeshGraph.h
@@ -73,77 +73,74 @@ namespace ttk {
     };
 
     inline size_t computeOutputCellSize(const size_t &nSubdivisions) const {
-      return 5
-             + nSubdivisions
-                 * 2; // cellDim + 4 corners + 2 for each subdivision
+      return 4 + nSubdivisions * 2; // 4 corners + 2 for each subdivision
     };
 
     inline size_t
-      computeOutputConnectivityListSize(const size_t &nInputCells,
-                                        const bool &useQuadraticCells,
-                                        const size_t &nSubdivisions = 0) const {
+      computeOutputConnectivityArraySize(const size_t &nInputCells,
+                                         const bool &useQuadraticCells,
+                                         const size_t &nSubdivisions
+                                         = 0) const {
       return useQuadraticCells
                ? nInputCells
-                   * 18 // 2*cellDim + 8 corners (a0,m0,m1,a1, m0,b0,b1,m1) + 8
+                   * 16 // 8 corners (a0,m0,m1,a1, m0,b0,b1,m1) + 8
                         // mid-edge nodes (a0m0,c,m1a1,a, m0b0,b,b1m1,c)
                : nInputCells * this->computeOutputCellSize(nSubdivisions);
     };
 
     // Mesh graph with quadratic quads
-    template <typename idType, typename dataType>
+    template <typename IT, typename CT, typename DT>
     int execute(
       // Output
-      float *outputPoints,
-      idType *outputConnectivityList,
+      CT *outputPoints,
+      IT *outputConnectivityArray,
+      IT *outputOffsetArray,
 
       // Input
-      const float *inputPoints,
-      const idType *inputConnectivityList,
+      const CT *inputPoints,
+      const IT *inputConnectivityArray,
       const size_t &nInputPoints,
       const size_t &nInputCells,
 
-      const dataType *inputPointSizes,
-      const float &sizeScale,
+      const DT *inputPointSizes,
+      const CT &sizeScale,
       const size_t &sizeAxis) const;
 
     // Mesh graph with linear polygon
-    template <typename idType, typename dataType>
+    template <typename IT, typename CT, typename DT>
     int execute2(
       // Output
-      float *outputPoints,
-      idType *outputConnectivityList,
+      CT *outputPoints,
+      IT *outputConnectivityArray,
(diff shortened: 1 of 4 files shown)


julien-tierny:
hi jonas, 

thanks for your quick fix. 
this looks good to me.

@CharlesGueunet 
is there something more long term that should be considered here instead?
(similarly to your core/base/skeleton/CellArray abstraction)

regarding the macros, that looks like a reasonable way forward.
can you open a new PR for that?
@pierre-guillou  what do you think?

in the meantime, I'll merge the fix.
cheers

CharlesGueunet:
I have created the `CellArray` utility which is heavily used in *Skeleton classes and hand both the old and new layout. We may want to update classes that explicitly create the cell array using the class so various memory layout can be supported.

CharlesGueunet:
The class is here: https://github.com/topology-tool-kit/ttk/blob/dev/core/base/skeleton/CellArray_new.h

julien-tierny:
hi charles, thanks a lot for your quick input.
are you suggesting jonas should re-factor his fix using the `CellArray` class instead?

CharlesGueunet:
If we want these classes to still support VTK prior to version 8, that is required. And in any case this would be a cleaner solution so we have a unified way to write cell arrays in TTK.

JonasLukasczyk:
I see. I got the feeling that at multiple points we already depend heavily on VTK9. Does TTK even build with VTK<9?

julien-tierny:
my impression is that the backward compatibility is not the most important aspect here (I'm not sure we're building against vtk8 anymore). I guess the `CellArray` is a nice intermediate layer that will help us in the future whenever kitware changes their connectivity modeling. ideally, we'd have only one class to touch (`CellArray`) instead of all the classes manipulating vtk's connectivity buffers directly (see for instance all the `core/base/skeleton/` classes).

JonasLukasczyk:
In most filters I proceed like this:
1) Compute Number of output cells (if possible to deduce beforehand)
2) Init Memory
3) Populate Values (direct memory access assuming AOS)
4) Pass as output

So there is no mem copy. If I understand correctly, I would have to instantiate the CellArray class and then pass this class to the base code to populate values, but not via direct memory access but through function calls. Will this be inlined in some way?

pierre-guillou:
Hi Jonas

I looked at your macros but I'm not very sure of which filters these may be applied (apart from `MeshGraph`).

The filters that spend the most build time are, at the moment, `FiberSurface` and `ReebSpace`. Those filters are double-templated over the input scalar field type (+ the 3-4 triangulation types). To mitigate this, I've introduced the `TTK_ENABLE_DOUBLE_TEMPLATING` CMake key (ON by default) that can be set to OFF to reduce the double templating to a simple templating.

The next source of build time should be all the `DiscreteGradient` users, i.e. `PersistenceDiagram`, `MorseSmaleComplex`,`TopologicalSimplification`and derivatives. Prior to the order field introduction (6 month ago), those were templated over the input scalar field, an offset field and the triangulation. Now, except for some top-level functions (e.g. `MorseSmaleComplex` need the scalar field values to put them on the Separatrices), the base of the `DiscreteGradient` is only templated over the triangulation type: the scalar field is redundant with the order field, and the order field is a `SimplexId` buffer. I'm still working on removing the last dependencies in `DiscreteGradient` to the scalar field type (`reverseGradient`), so this should be better soon.

Since I'm working quite a lot on those filters and the build times are still high, I've reduce in my Git work branch the number of template instantiations over the scalar field type by adding the following to `ttkMacros.h`:

```
#ifdef vtkTemplateMacro
#undef vtkTemplateMacro
#define vtkTemplateMacro(call)                    \
  vtkTemplateMacroCase(VTK_DOUBLE, double, call); \
  vtkTemplateMacroCase(VTK_FLOAT, float, call);   \
  vtkTemplateMacroCase(VTK_INT, int, call);       \
  vtkTemplateMacroCase(VTK_LONG_LONG, long long, call);

#endif
```

This is sufficient for my use-cases and reduces the build times further. I have to remove it every time I check the ttk-data states though because it messed, among others, with the `MeshGraph` filters.

With those two tricks, I'm able to build a working version of TTK in my laptop (16GB RAM) without swapping.

pierre-guillou:
Concerning the new VTK cell layout, I used it explicitly in the `MorseSmaleComplex` filter in order to fill directly and in parallel the Separatrices (1d and 2d) geometry, with lots of pre-computations to reduce the number of allocations. I don't think this is compatible with VTK < 8 anymore.

PR #563, opened 2021-02-15
Minor fix for isnan use

CharlesGueunet:
Dear Julien,

Following an update on my system, I needed this minor fix to compile TTK.

Best,
Charles


julien-tierny:
looks good to me, thanks!

PR #566, opened 2021-02-19
header cleanup; icosphere 2.0

JonasLukasczyk:
Hi Julien, this PR:
* removes some unnecessary headers from the ttkAlgorithm which then had to be included in one or two submodules.
* overhauls the icosphere code to write directly into vtk9 cell arrays and now produces vtkPolyData output instead of vtkUnstructuredGrid (big memory/performance improvement)
* adds a new function to ttkUtils which can be used to access the pointer of an array directly for a given type. This is very convenient as I use this all over the place. The new call
`ttkUtils::GetPointer<float>(array)`
is equivalent to
`static_cast<float*>(ttkUtils::GetVoidPointer(array))`

Best
Jonas

### DIFF
--- a/core/base/icosphere/Icosphere.h
+++ b/core/base/icosphere/Icosphere.h
@@ -44,69 +44,70 @@ namespace ttk {
       return 1;
     };
 
-    template <typename dataType, typename idType>
+    /**
+     * Translates an icosphere
+     */
+    template <typename DT, typename IT>
     int translateIcosphere(
       // Output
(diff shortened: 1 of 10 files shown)


JonasLukasczyk:
Mh, seems like the CI uses another VTK version than me (vtk from PV 5.9) so I'm working through some missing include statements.

julien-tierny:
looks good, thanks!

PR #567, opened 2021-02-20
Allow setting TTK_CELL_ARRAY_LAYOUT from superprojects

dsakurai:
This PR allows a superproject integrating TTK using `add_subdirectory()` to customize the value of `TTK_CELL_ARRAY_LAYOUT`.

Exception: if integration of TTK to VTK or ParaView is requested, the value will be set to the appropriate value.

[Here's a use case.](https://github.com/topology-tool-kit/ttk/files/6013909/CMakeLists.txt)

(Sorry that my initial PR message was accidentally empty)

### DIFF
--- a/config.cmake
+++ b/config.cmake
@@ -11,10 +11,12 @@ endif ()
 
 # Options & dependencies
 
-# CellArray use legacy single array by default.
-# Do not use FORCE so that this settings can be overriden
-# when integrating TTK with add_subdirectory(). In such scenarios, set the value before the
-# add_subdirectory() call.
+# the 'TTK_CELL_ARRAY_LAYOUT' drive the behavior of TTK to store cell points.
+# This variable has two possible values "SingleArray" and "OffsetAndConnectiviy".
+# * "SingleArray" use a layout compatible with VTK < 9 were the cell array store the
+#   cells and their connectivity in a flat array
+# * "OffsetAndConnectivity" use a layout comatible with VTK >= 9, having two arrays 
+#   (see https://vtk.org/doc/nightly/html/classvtkCellArray.html#details for more info)
 set(TTK_CELL_ARRAY_LAYOUT "SingleArray" CACHE STRING "Layout for the cell array.")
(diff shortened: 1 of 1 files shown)


julien-tierny:
thanks daisuke.
@CharlesGueunet could you review this really quick?

config.cmake
```diff
@@ -11,50 +11,54 @@ endif ()
 
 # Options & dependencies
 
-# CellArray use legacy single array by default
-set(TTK_CELL_ARRAY_LAYOUT "SingleArray" CACHE STRING "Layout for the cell array." FORCE)
+# CellArray use legacy single array by default.
+# Do not use FORCE so that this settings can be overriden
+# when integrating TTK with add_subdirectory(). In such scenarios, set the value before the
+# add_subdirectory() call.
```
vollia: I am not sure this comment is meaningful outside of the current MR. We may replace it in order to explain the behavior of `TTK_CELL_ARRAY_LAYOUT` instead:

```suggestion
# the 'TTK_CELL_ARRAY_LAYOUT' drive the behavior of TTK to store cell points.
# This variable has two possible values "SingleArray" and "OffsetAndConnectiviy".
# * "SingleArray" use a layout compatible with VTK < 9 were the cell array store the
#   cells and their connectivity in a flat array
# * "OffsetAndConnectivity" use a layout comatible with VTK >= 9, having two arrays 
#   (see https://vtk.org/doc/nightly/html/classvtkCellArray.html#details for more info)
```

CharlesGueunet (approved):

CharlesGueunet:
Except a minor remark concerning a comment, LGTM

config.cmake
dsakurai: Excellent. As TTK prefers `CACHE FORCE` in many places, though, I recommend shortly mentioning why FORCE is not desired here. @CharlesGueunet, what do you think?

dsakurai: Something like
~~~
# FORCE is not used so that parent projects can override the configuration.
~~~

vollia: Yep, this make sense too

julien-tierny:
alright, let's go then

PR #568, opened 2021-02-22
CinemaQuery: Fix vtkCharArray/vtkSignedArray support

pierre-guillou:
According to VTK, applying the `ToString()` method on a `vtkCharArray`/`vtkSignedCharArray` element returns its character representation. Since we use primarly those array types to store small integers irrelevant to their character representation, this leads to bugs when the `CinemaQuery` module converts arrays of those types to SQL. For instance, applying the `CinemaQuery` filter on a `vtkTable` conversion of a `ttkScalarFieldCriticalPoints` Point output returns a SQLite error caused by the CriticalType and the IsOnBoundary `vtkSignedCharArray`.

This PR makes sure that the corresponding array elements are kept into their integer representation when sending them to SQLite.
Also, to help debug, SQLite error messages are now more explicit about the corresponding step in which they fail.

No modifications observed on the relevant ttk-data states.

Enjoy,
Pierre


julien-tierny:
looks good, thanks!

PR #569, opened 2021-02-23
ttkExtract update

JonasLukasczyk:
Hi Julien,
this PR contains an upgrade of the ttkExtract filter. In the past the filter was already able to extract a vtkImageData object from a vtkMultiBlockDataSet, but due to vtk's pipeline concept one has to specify the available image extent already during the RequestInformation pass. Unfortunately, deducing this during the RequestData pass is too late. To cope with this the filter guessed during the RequestInformation pass the extent based on the bounds of the input object. This works as long as the bounds are equal to the extent. With this PR we can actually specify the extent directly with some help of the PV UI instead of using the bounds to do some guessing. The filter will work as before but now it will also support extractions of grids that are deformed, e.g., lat-lng grids.

Note, the corresponding property `ImageBounds` is now renamed to `ImageExtent` which effects some state files in ttk-data. I will open a PR for those.

TLDR; The ttkExtract filter can now extract any vtkImageData object based on a specified extent (initially autocompleted by PV). This is very useful for any pipeline that utilizes the Cinema filters since they produce vtkMultiBlockDataSet outputs. For convenience, block iterations of the ttkForEach filter also supports this by explicitly setting its output type to vtkImageData.

Best
Jonas

### DIFF
--- a/core/vtk/ttkExtract/ttkExtract.cpp
+++ b/core/vtk/ttkExtract/ttkExtract.cpp
@@ -90,22 +90,13 @@ int ttkExtract::FillOutputPortInformation(int port, vtkInformation *info) {
 // RequestInformation
 // =============================================================================
 int ttkExtract::RequestInformation(vtkInformation *,
-                                   vtkInformationVector **,
+                                   vtkInformationVector **inputVector,
                                    vtkInformationVector *outputVector) {
+
   if(this->ExtractionMode == 0 && this->GetOutputType() == VTK_IMAGE_DATA) {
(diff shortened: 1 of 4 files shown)


julien-tierny:
looks good to me, thanks!

julien-tierny:
a quick remark though: ttkExtract starts to look like a very useful swiss army knife but I have the impression that if it were better documented, people would use it much more.
(it happened several times here that people wanted to use it but didn't because they could not figure out how to use it)

that'd be great if the paraview xml and doxygen documentations could be expended.

PR #575, opened 2021-02-25
removed empty files

JonasLukasczyk:
Hi Julien,
this PR just deletes some empty files that I just kept around because I did not know about the `ttk_add_base_template_library` cmake function.

Best
Jonas

### DIFF
--- a/core/base/localizedTopologicalSimplification/LocalizedTopologicalSimplification.h
+++ b/core/base/localizedTopologicalSimplification/LocalizedTopologicalSimplification.h
@@ -34,7 +34,7 @@
 
 namespace ttk {
 
-  std::string toFixed(const float &number, const int precision = 2){
+  std::string toFixed(const float &number, const int precision = 2) {
     std::stringstream vFraction;
     vFraction << std::fixed << std::setprecision(precision) << number;
     return vFraction.str();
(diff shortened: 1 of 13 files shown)


core/base/localizedTopologicalSimplification/LocalizedTopologicalSimplification.h
```diff
@@ -34,7 +34,11 @@
 
 namespace ttk {
 
-  std::string toFixed(const float &number, const int precision = 2);
+  std::string toFixed(const float &number, const int precision = 2) {
```
pierre-guillou: Hi Jonas,
I moved this function's body into a .cpp file several PRs ago to avoid ODR violations (I had a warning with clang-tidy, check for "ODR" in the "Use clang-tidy" step of the "lint-code" CI job). 
I think that if you want to keep it in the header file, it might be better to declare it as inline (as also suggested by clang-tidy).

JonasLukasczyk: You are right, I will at least inline/constexp it. I always wanted to move it to the debug class. Any thoughts on that?

pierre-guillou: Why not, it may be useful outside of LTS. I always find cumbersome to use `setprecision` directly.

julien-tierny:
looks good!

PR #578, opened 2021-03-02
DiscreteGradient: Move allocations outside main loop

pierre-guillou:
This PR moves some memory allocations in the DiscreteGradient module:

- allocating the `gradient_` vector is now done during the preconditioning step (previously done in the `buildGradient` method).
- the `lowerStar` and the priority queue variables have been hoisted out of the parallel loop in order to reduce the dynamic allocations by cleaning & reusing their storage between iterations by a same thread (with the help of OpenMP's firstprivate clause).

Depending on the dataset size and the number of threads used, a small speedup is observed.

No modification observed in the ttk-data state files.

Enjoy,
Pierre


julien-tierny:
awesome, thanks!

PR #579, opened 2021-03-03
BlockTuple Extraction Mode; Auto ForEach Update

JonasLukasczyk:
Hi Julien,
this PR has two contributions:
1) The `ttkExtract` filter now makes it possible to extract so called block tuples. In many tracking use cases you encounter `vtkMultiBlockDataSets` of the following form:
  MB containing three lists:
  > List of Merge Trees over time
  > List of Merge Tree Domain Segmentations over time
  > List of Persistence Diagrams over time
Before it was not possible to simply extract the `MT+S+PD` tuple that corresponds to the i-th timestep. Now the `ttkExtract` filter understands this common data structure and makes it possible to extract tuples by index. The `ForEach` filter can not also be used to iterate over all tuples.

2) Right now if users change something inside a `ForEach` loop they need to manually restart the `ForEach` iterations. With this PR the `EndFor` filter detects if an iteration is executed twice (if a change inside the loop occurred), so for convenience it will automatically restart the iterations.

Best
Jonas

### DIFF
--- a/core/vtk/ttkEndFor/ttkEndFor.cpp
+++ b/core/vtk/ttkEndFor/ttkEndFor.cpp
@@ -68,11 +68,19 @@ int ttkEndFor::RequestData(vtkInformation *request,
   // get iteration info
   int i = forEach->GetIterationIdx() - 1;
   int n = forEach->GetIterationNumber();
-  this->printMsg("Iteration ( " + std::to_string(i) + " / "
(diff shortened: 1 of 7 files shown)


JonasLukasczyk:
I also tested with ttk-data and it seems nothing is affected.

julien-tierny:
looks good, thanks!

PR #580, opened 2021-03-08
Azure: Build ParaView without Qt on macOS

pierre-guillou:
Qt 6 just arrived in macOS brew and, in a foreseeable manner, it is not compatible (yet) with the ParaView 5.8.1 we build in the macOS job of the Azure CI.

To avoid similar issues in the future and to not be stuck into a specific version of Qt, this PR builds ParaView without the Qt interface. This should be enough for this particular CI job (and should also make this job faster, as downloading and installing Qt via brew takes a long time).

Enjoy,
Pierre

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -364,7 +364,7 @@ jobs:
   steps:
   - script: |
       brew install --cask xquartz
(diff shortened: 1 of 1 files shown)


julien-tierny:
awesome, thanks!

PR #581, opened 2021-03-08
Improved number parsing in ArrayEditor based on std::stod

JonasLukasczyk:
Hi Julien, the PR title says it all. Just a minor improvement so that the ArrayEditor can handle more edge cases such as trailing white space and so on.

Best
Jonas


CharlesGueunet (approved):

julien-tierny:
looks good, thanks!

PR #584, opened 2021-03-10
(Explicit)Triangulation: Use vectors of arrays instead of vectors of vectors for additional performance

pierre-guillou:
The explicit triangulation data are stored inside `std::vector<std::vector<SimplexId>>` which causes unnecessary pointer indirections when accessing data.

This PR replaces several of those vectors of vectors with vectors of `std::arrays` when their size is known at compile-time. This concerns:

* triangle vertices (3)
* tetrahedron vertices (4)
* triangle edges (3)
* tetrahedron edges (6)
* tetrahedron triangles (4)
* edge vertices (`std::pair<SimplexId, SimplexId>` has been replaced with `std::array<SimplexId, 2>` to keep a unified style)

The Triangulation API has been, for the most part, left untouched, apart from the `getEdges` and the `getTriangles` methods whose return type has been modified accordingly.

Since the `cells` can represent either triangles or tetrahedrons, 3 new vectors of vectors members have been introduced. They are to be filled by the triangulation implementations at each `getCells`, `getCellEdges` and `getCellTriangles` calls from the relevant vectors of arrays. This can cause memory duplication for explicit grids but since using those methods is not advised, this should be fine.

A conceptual change also occurred in the AbstractTriangulation between cells and triangles in 2D. Before this PR, triangle methods were redirected to cell methods whereas now cell methods redirect to triangle ones in 2D. Confer `getCellEdge`, `getCellEdgeNumber`, `getCellEdges` and triangle equivalents. Without this change, the explicit triangulation would request the wrong vector of arrays in 2D: `geTriangleEdge`  would call `getCellEdge` which would access the `cellEdgeList_`/`tetraEdgeList_` member, empty in 2D.

A little cleanup has also been done in the `skeleton` module to accelerate the preconditioning methods. In particular, the `buildTriangles` method gained an 3x speedup while staying (for the most part) sequential.
A C++ RAII wrapper around OpenMP locks has also been introduced in the `common` module. It is currently used in `buildVertexTriangles` and enables another 3x speedup with 16 threads.
Some unused because inefficient parallel implementations were removed (e.g. `OneSkeleton::buildEdgeList`).

To keep up with the cell being tetras in 3D and triangles in 2D, three skeleton methods were templated:
* `OneSkeleton::buildEdgeLinks`
* `ThreeSkeleton::buildCellEdges`
* `ZeroSkeleton::buildVertexLinks`

Explicit template instantiations were used to keep the method definitions in the .cpp files.

Performance improvements for the discrete gradient computation range from +30% to +90% on explicit triangulations. The discrete gradient triangulation preconditioning also gained a x2 speedup.

In addition, the log verbosity of `ZeroSkeleton::buildVertexStars` was reduced from DETAIL to VERBOSE.

In fine, this PR reworks the memory usage of the explicit triangulation by removing a lot of pointer indirections and reducing memory fragmentation, leading to better performances overall.

No change has been observed in the ttk-data states files.

Enjoy,
Pierre

### DIFF
--- a/core/base/abstractTriangulation/AbstractTriangulation.cpp
+++ b/core/base/abstractTriangulation/AbstractTriangulation.cpp
@@ -40,9 +40,9 @@ int AbstractTriangulation::clear() {
   boundaryTriangles_.clear();
   boundaryVertices_.clear();
 
-  cellEdgeList_.clear();
+  tetraEdgeList_.clear();
   cellNeighborList_.clear();
-  cellTriangleList_.clear();
+  tetraTriangleList_.clear();
 
   edgeLinkList_.clear();
   edgeList_.clear();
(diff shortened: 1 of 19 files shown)


JonasLukasczyk:
Cool stuff!

julien-tierny:
thanks!

PR #586, opened 2021-03-12
PersistenceDiagramClustering: Fix segfault with non-default parameters

### LINKED ISSUE #564: TTK Persistance Diagram Clustering crash when displaying cluser as star
## Description

Applying the `TTK PersistanceDiagramClustering` filter using the "Display Clusters as star" option for the first computation makes the filter crash (Segmentation Fault [on this line](https://github.com/topology-tool-kit/ttk/blob/b886f56201f2e4ea4572e9bfa888410373391f1b/core/vtk/ttkPersistenceDiagramClustering/ttkPersistenceDiagramClustering.cpp#L699)).
If the filter is first called using the default "Compact" view then the option is switched to "Cluster as star" it works as expected. 

## To reproduce

On a multi-block containing only persistence diagrams:
* Select the `TTK PersisitanceDiagramClustering` filter
* Change the `Displaying method` to **Clusters as star**
* Change the Spacing to **1** (not sure this is required)
* **Apply**

 - OS: Artix Linux
 - Compiler GCC 10.2.0
 - TTK dev branch, up to date.

@julesvidal have you already experienced this ?

pierre-guillou:
When modifying the DisplayMethod prior to applying the PersistenceDiagramClustering filter in ParaView, the `needUpdate` member was always set to false, thus skipping the application of the base layer routines on the input diagrams and leading in fine to segfaults further down in the pipeline.

This PR adds check around the modification of the `needUpdate` member to allow a first pass on the input data while still skipping later computations. This should fix #564.

Enjoy,
Pierre

### DIFF
--- a/core/vtk/ttkPersistenceDiagramClustering/ttkPersistenceDiagramClustering.h
+++ b/core/vtk/ttkPersistenceDiagramClustering/ttkPersistenceDiagramClustering.h
@@ -121,7 +121,10 @@ public:
     Spacing = spacing;
     oldSpacing = spacing;
     Modified();
-    needUpdate_ = false;
+    if(!intermediateDiagrams_.empty()) {
+      // skip clustering computation only if done at least once before
+      needUpdate_ = false;
+    }
   }
   vtkGetMacro(Spacing, double);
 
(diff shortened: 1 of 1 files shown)


julien-tierny:
looks good, thanks!

PR #587, opened 2021-03-16
Enums; InSitu Cinema Writer; ttkExtract improvements

JonasLukasczyk:
Hi Julien,
here comes a package of several improvements that were necessary for my latest in situ project. At the core of the update is an improvement of the cinema writer which can now synchronize multiple mpi processes that want to write to the same cinema database. Writing some in situ code at the vtk api level also made it apparent that the interface of the `ttkExtract, ttkForEach, ttkCinemaWriter,...` made heavy use of integer enums, where one always has to carefully lookup which number represents what. This is obviously bad, so with this update I also introduces strict class enums. Unfortunately the paraview UI can not handle those, so I added a macro in `ttkMacros` called `ttkSetEnumMacro` that generates two functions: one setter with the strict enum, and one setter with the underlying data type which PV can use. It would be cool if at some point we refactored other VTK filters to use strict enums too.

@pierre-guillou: As I was updating the CinemaWriter I noticed you introduces an option that controls if the filter passes its input or keeps an empty output. What is the usecase for this? I think that performing a shallow copy is negligible  and should always be performed to be more consistent with VTK.

Best
Jonas


pierre-guillou:
Hi Jonas,

I indeed introduced this `ForwardInput` option to make the CinemaWriter filter act as a sink. The use-case is, in a Catalyst in-situ pipeline, I wanted to store the data via CinemaWriter and not propagate it afterwards to a regular ParaView writer (e.g. `XMLPUnstructuredGridWriter`). In my mind, this reduced the I/O operations (we don't write the data twice). Do you think it's a valid use-case?

Pierre

JonasLukasczyk:
Hi Pierre,
I see, so this was necessary to circumvent some automatic write operations with catalyst. If there is no way around this we can of course keep this flag. What do you think?

pierre-guillou:
From what I know of the Catalyst API (https://kitware.github.io/paraview-docs/latest/python/paraview.coprocessing.html), you need to use a ParaView "write" operation (`coprocessing.RegisterWriter`, `coprocessing.RegisterView`) to trigger a given branch of a Catalyst pipeline at a given frequency. In particular, `coprocessing.RegisterWriter` does not with CinemaWriter as input, you need to use a "regular" ParaView writer.
With this small `ForwardInput` hack, I was able to feed the ParaView writer with empty data and thus reducing the I/O operations. I am not aware of an other way of controlling a Catalyst pipeline branch.

JonasLukasczyk:
Ok, that is a good reason to keep it for now.

core/vtk/ttkCinemaWriter/ttk.module
```diff
@@ -8,4 +8,4 @@ DEPENDS
   ttkAlgorithm
   ttkTopologicalCompressionWriter
   Boost::boost
-  Boost::system
+  Boost::system
```
julien-tierny: are you sure about Boost::system?
@pierre-guillou wrote in the merged PR #590 Boost::system was no longer mandatory...

julien-tierny: note: that must be the conflict I now see down there :)

julien-tierny:
have you run the ttk-data states to make sure nothing's broken there?
thanks!

JonasLukasczyk:
Hi, I didn't notice any differences in the ttk-data statefiles. In the next dev meetup we should discuss automatic testing^^

pierre-guillou:
Hi Jonas,
I've begun to look into automatic testing of the ttk-data states in GitHub Actions. While it works on Ubuntu VMs with offscreen rendering, I'm still struggling a bit on Windows. Finally, on macOS, there are still RPATH issues I think that prevent ParaView to load the TTK libraries.

JonasLukasczyk:
We could also have test programs like VTK has.

core/vtk/ttkCinemaWriter/ttk.module
pierre-guillou: Well, with my last changes, `Boost::system` should in fact not work any longer…

JonasLukasczyk: Note: this file only appears in the change log because my editor automatically removed the newline symbol at the end of the file xD. So in this PR I did not really touch the ttk.module file. However, I use `boost::interprocess::file_lock` / `#include <boost/interprocess/sync/file_lock.hpp>`. I'm not 100% sure if this is included in the `Boost::boost` dependency, but it should be. So I guess we can drop `Boost::system`

pierre-guillou: I just checked and, on my system, interprocess is a header-only library and it is included in `Boost::boost`.

JonasLukasczyk:
Hi @julien-tierny,
I guess this PR is ready now.

julien-tierny:
alright, thanks!

PR #588, opened 2021-03-17
Add missing vtkCellArrays to vtkUnstructuredGrid outputs

pierre-guillou:
Several TTK filters generate a point cloud output: `DiscreteGradient`, `MorseSmaleComplex`, `ScalarFieldCriticalPoints`, `GaussianPointCloud`. To do so, these filters fill their `vtkUnstructuredGrid` output with `vtkPoints` only. While this seems to be fine inside ParaView, this is causing segfaults in the corresponding standalones, because the writer expects cells inside the `vtkUnstructuredGrid`.

This PR adds a `vtkCellArray` to the output of the four filters previously cited. Among other things, this fixes the segfaults in the standalones and allow to Threshold directly the output without applying a `SphereFromPoints`.

In addition, VTK structure are now filled in parallel in DiscreteGradient and ScalarFieldCriticalPoints. Furthermore, the MorseSmaleComplex standalone has been modified to skip writing empty outputs.

No modification has been observed in the ttk-data state files.

Enjoy,
Pierre

### DIFF
--- a/core/vtk/ttkDiscreteGradient/ttkDiscreteGradient.cpp
+++ b/core/vtk/ttkDiscreteGradient/ttkDiscreteGradient.cpp
@@ -61,16 +61,20 @@ int ttkDiscreteGradient::dispatch(vtkUnstructuredGrid *outputCriticalPoints,
   // critical points
   {
     this->setCriticalPoints<scalarType>(triangulation);
+    const auto nPoints = outputCriticalPoints_numberOfPoints_;
 
     vtkNew<vtkPoints> points{};
+    points->SetNumberOfPoints(nPoints);
 
     vtkNew<vtkSignedCharArray> cellDimensions{};
     cellDimensions->SetNumberOfComponents(1);
     cellDimensions->SetName("CellDimension");
+    cellDimensions->SetNumberOfTuples(nPoints);
 
     vtkNew<ttkSimplexIdTypeArray> cellIds{};
     cellIds->SetNumberOfComponents(1);
     cellIds->SetName("CellId");
+    cellIds->SetNumberOfTuples(nPoints);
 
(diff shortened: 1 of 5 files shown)


JonasLukasczyk:
Perfect, I was working on that as well xD. What do you think about switching from vtkUnstructuredGrid to vtkPolyData? Theoretically this should optimize quite a few things (rendering, memory, etc).

JonasLukasczyk:
I recently changed the `ttkIcosphere` to output vtkPolyData output.

pierre-guillou:
I don't/didn't know exactly the difference between vtkUnstructuredGrid and vtkPolyData and the implications of such a change. Since TTK produced and acts mainly on Image Data and Unstructured Grids, I wanted to modify as few code as possible while fixing the segfaults.

JonasLukasczyk:
In general `vtkPolyData` can only represent either vertices, lines, or triangles (no mixed types). `vtkUnstructuredGrid` lets you do whatever you want, but `PolyData` therefore uses a lot less memory and rendering can be done quicker because represented primitives can directly be mapped (forwarded) to the rendering engine, whereas `vtkUnstructuredGrid` has some overhead. Some filters such as the `PersistenceDiagram` and the `ScalarFieldCriticalPoints` only produce a fixed type that can be represented by `vtkPolyData`, so to me it seems it is the better data container. TTK also supports triangulations of `vtkPolyData`.

pierre-guillou:
Thanks for this explanation, it's clearer now.

julien-tierny:
alright, looks good, thanks!

ps: the switch to vtk poly data is not systematic, some outputs can only be expressed with unstructured grids (e.g. segmentations). some others (e.g. separatrices of the morse-smale complex) can indeed.

PR #589, opened 2021-03-17
DilateErode: Add grayscale operators, opening + closing

pierre-guillou:
The actual DilateErode filter performs, from what I've understood, a binary dilation or erosion on the level set of user-provided value `PivotLabel`.

Since I'm more familiar with the grayscale implementation of these operators with the default structuring element (dilation = max over neighbors, erosion = min over neighbors), this PR adds them into this filter. The default implementation is still the "binary" one.

On top of that, I've also added the morphological opening (erosion then dilation) and closing (dilation then erosion). They can be performed using either the "binary" or the grayscale implementations of the elementary operators.

Enjoy,
Pierre


pierre-guillou:
Hi @JonasLukasczyk,
What do you think of these additions? I tried to make the fewest modifications to your own code while expanding onto it.

JonasLukasczyk:
Hi Pierre,
this looks great! I implemented the binary operator to actually dilate/erode only a specific feature label, but these general cases are also extremely useful. But since we now have more modes, I would suggest to make the API a bit more comprehensible by either replacing the mode integer with a strict class enum or explicitly provide methods such as Close, Open, Dilate, Erode, etc.

If you opt for the enums, I would suggest to work on that after the enum PR goes through.

pierre-guillou:
I saw your last PR on using enum classes more broadly and it was exactly my thoughts to use it on the DilateErode filter once both PRs are merged. If your are fine doing the work, I'll let you!

JonasLukasczyk:
xD After the VIS dealine I can work on this. I wanted to provide a new module `ttkTrackingFromLabels` that uses the ttkDilateErode module. Do you think we should rename it now to something like `ttkMorphologicalOperators`?

pierre-guillou:
Yeah, `ttkMorphologicalOperators` may be more precise here. It might also draw in more operators (morphological gradient for instance).

JonasLukasczyk:
Do you want to rename the filter in this PR?

pierre-guillou:
I'll do that now

julien-tierny:
ok, looks good!

PR #590, opened 2021-03-17
Azure: Manually install Boost on Windows

pierre-guillou:
This PR fixes the Azure CI Windows build, which fails because Boost was recently removed from the Windows image.

To do so, a Boost 1.72 installer is downloaded from SourceForge and installed to the image prior to the TTK CMake configuration.


pierre-guillou:
So, after several tries, I managed to get Boost working again on Windows. However, CMake doesn't seem to detect the Boost libraries, so I removed the (unused) dependency to `Boost::system`. Everything seems to work fine nonetheless. I guess this PR also fixes #562.

Everything seems fine with ttk-data too.

julien-tierny:
the boost::system had to be there for a reason at some point (can't remember now).
I'll trust the CI for now :)

(have you run ttk-data states?)

pierre-guillou:
Yes, I've run all ttk-data states on my computer.

PR #591, opened 2021-03-26
ContourTreeAlignment - New metric and input structure

floWetzels:
Added overlap metric for matching on time dependent data sets to the ContourTreeAlignment module. The overlap metric for contour tree arcs needs the corresponding segmentations. Therefore the filter now has to input options:
- A multiblock of contour trees (old mode, does not work with overlap metric, but is kept to still have the lightweight option without segmentation)
- A multiblock of multiblocks, where the inner multiblocks contain a contour tree and its segmentation

Also fixed some memory leaks from the conversion of raw pointers to smart pointers in the base layer.

### DIFF
--- a/core/base/contourTreeAlignment/ContourTreeAlignment.cpp
+++ b/core/base/contourTreeAlignment/ContourTreeAlignment.cpp
@@ -20,8 +20,8 @@ void ttk::ContourTreeAlignment::computeBranches() {
 
   // find path to global max
   std::shared_ptr<AlignmentNode> nextNode
-    = minNode->edgeList[0]->node1 == minNode ? minNode->edgeList[0]->node2
-                                             : minNode->edgeList[0]->node1;
+    = minNode->edgeList[0]->node1.lock() == minNode ? minNode->edgeList[0]->node2.lock()
+                                             : minNode->edgeList[0]->node1.lock();
   std::vector<std::shared_ptr<AlignmentNode>> maxPath_
     = pathToMax(nextNode, minNode).second;
   std::vector<std::shared_ptr<AlignmentNode>> maxPath;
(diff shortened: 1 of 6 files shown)


julien-tierny:
hi @floWetzels ,
thanks for your pr.

@pierre-guillou: it seems that florian is struggling a bit with the clang-format step of the ci as well as the windows build.
could you help him a bit on that? (for instance, what version of clang-format should he use, etc.)

thanks!

pierre-guillou:
Yes!
So, several month ago, I introduced a code formatting check step in TTK's CI using GitHub Actions. To do so, we rely on `clang-format`, a tool under the CLang/LLVM umbrella. Unfortunately, it seems that `clang-format`'s formatting rules change between major updates. To alleviate this issue, we try to stick to a precise version of `clang-format`, which is by convention the one on Julien's computer(s). Currently, it's `clang-format` v.11 (https://github.com/topology-tool-kit/ttk/blob/414ae8d02785cfe21e34a3e2dde1232784a2c635/.github/workflows/check.yml#L26), which can be installed as a standalone package on Ubuntu 20.04. I don't think it is available on older Ubuntu versions but official CLang/LLVM binaries are distributed via GitHub (https://github.com/llvm/llvm-project/releases/tag/llvmorg-11.0.1) and `clang-format` should be usable from there too.

Concerning the Windows issue, it seems that the CI just happen during a downtime of the Boost server. Unlucky. But the CI can be restarted with a new git commit or just with an amend on the latest commit (it will change the commit date) and a `push --force`.

Hope this helps,
Pierre

floWetzels:
Thank you, @pierre-guillou @julien-tierny!
I used version 10 of clang-format, which is the default version you get by installing with apt. For version 11, you need to install the clang-format-11 package. So this should solve the problem, I hope.

pierre-guillou:
Exactly, I use the `clang-format-11` apt package on Ubuntu 20.04 on the CI to keep code formatting consistent.

julien-tierny:
looks good, thanks!

PR #592, opened 2021-03-30
Switch some filters output to PolyData, rework memory in Quadrangulation, DiscreteGradient filters

pierre-guillou:
As a follow-up to #588 and per Jonas' suggestion, this PR changes the output of point cloud generating filters from `vtkUnstructuredGrid` to `vtkPolyData`. The concerned filters are:

* the two outputs (critical points, glyphs) of the DiscreteGradient,
* the critical points and separatrices outputs of the MorseSmaleComplex,
* the quadrangulations generated by MorseSmaleQuadrangulation and QuadrangulationSubdivision,
* the MatrixToHeatMap output,
* the point clouds outputs of ScalarFieldCriticalPoints and GaussianPointCloud.

TriangulationRequest, forgotten in #588, now also outputs cells for vertex requests. Since `vtkPolyData` only works with 2D cells and can't represent tetrahedrons, the `vtkUnstructuredGrid` output has been kept.

In addition to this work on filters output, I also did a bit of cleaning in the DiscreteGradient and the Quadrangulation filters:

* the critical points and glyphs outputs of the DiscreteGradient have been refactored in specific methods,
* the glyphs are now computed in parallel,
* the gradient structure has been changed from `std::vector<std::vector<std::vector<SimplexId>>>` to `std::array<std::vector<SimplexId>, 6>` to reduce indirections. The documentation of this type has been modified accordingly.
* the `setCriticalPoints` and `setManifoldSize` methods have been refactored to use references, thus enabling to remove some state in the DiscreteGradient class,
* the critical points `cellScalars` point data array is now filled directly in the DiscreteGradient and the MorseSmaleComplex VTK layer, which avoids passing void pointers and templating some base methods according to the scalar field type,
* the quadrangulation data structures have also been slightly modified: instead of a struct, I now use `std::array` for quadrangles. The "interleaved" cell structure has also been dropped: the quadrangles now don't embed their number of points,
* some leftovers of the Great Migration have been fixed in the Quadrangulation filters (using the correct macro calls, replacing `vtkSmartPointers` with `vtkNew`…)

No modification observed in the ttk-data state files.

Enjoy,
Pierre

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -315,7 +315,7 @@ jobs:
 
   - script: |
       ./ttkExample-vtk-c++ -i ../examples/data/inputData.vtu
-      if [ ! -f curve.vtk ] || [ ! -f separatrices.vtu ] || [ ! -f segmentation.vtu ] ; then
+      if [ ! -f curve.vtk ] || [ ! -f separatrices.vtp ] || [ ! -f segmentation.vtu ] ; then
         exit 1
       fi
     workingDirectory: 'build-example-vtk/'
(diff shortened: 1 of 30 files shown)


julien-tierny:
looks good, thanks!

PR #593, opened 2021-03-30
ExplicitTriangulation: Replace vectors of vectors with flat vectors

pierre-guillou:
Following #584, this PR replaces the remaining `std::vector<std::vector<SimplexId>>` in the ExplicitTriangulation with a combination of two `std::vector<SimplexId>`, one for the offsets and one for the data, similarly to VTK and its new `vtkCell` connectivity and offsets arrays.

To do so, a new class, `FlatJaggedArray`, has been introduced in the `common` base module. Every vector of vector generated by the `skeleton` module has been converted to use this class instead, leading to massive performance improvements in the explicit triangulation precondition step (up to x3 from #584) due to fewer memory allocations. Performance improvements outside the precondition step and inside filters hot paths amount to 15-20%, due to cache-friendliness.

To get there, a lot of code inside the `skeleton` module has been rewritten to fill this new data structure. While the result might lose some parallelism, it still outperforms the previous parallel version. Notable changes include:
* links are computed from stars,
* `buildEdgeStars` and `buildCellEdges` have been merged inside `buildEdgeList` to be consistent with `buildTriangleList`,
* cell neighbors are no longer computed by intersecting the vertices stars but from triangle stars in 3D and edge stars in 2D,
* edge triangles are computed by returning the triangle edges data,
* `boost::container::small_vector` (Boost >= 1.60, [description](https://www.boost.org/doc/libs/1_75_0/doc/html/boost/container/small_vector.html)) has been used to reduce allocations in some specific places (`buildEdgeList`, `buildTriangleList`) were the use of a `vector<vector>>` was unavoidable. Since they were used only in the source file, the dependency does not leak outside the skeleton module,
* the precondition methods in the ExplicitTriangulation module were moved from the header to the source file. This prevents the skeleton module to leak outside ExplicitTriangulation and might lead to a small improvement in build times, at the expense of their inlining (missing `map` headers had to be added since skeleton headers used `std::map`).

This time, the AbstractTriangulation API has been kept (mostly) unchanged: the new buffers are converted automatically to `std::vector<std::vector<SimplexId>> when calling the methods to get a raw access to the internal buffers (which is advise against). Apart from that, `AbstractTriangulation::footprint` method has been adapted to also return the size of these new buffers when called from an ExplicitTriangulation and, for convenience, two missing methods were added: `getEdgeVertexNumber` and `getTriangleVertexNumber` return constants.

No ttk-data state change has been observed with those commits.

Enjoy,
Pierre


julien-tierny:
ok, thanks!

PR #595, opened 2021-04-13
WebSocketIO

JonasLukasczyk:
Hi Julien,
* This PR adds a new module called `ttkWebSocketIO` that makes it possible to send `vtkDataObjects` from `ParaView (VTK)` to the browser via a websocket. I used this several times to outsource interaction from ParaView to js-based visual analytic interfaces.
* This introduces a new optional dependency to a header-only library: [WebSocketPP](https://github.com/zaphoyd/websocketpp)
* In the standalone folder I provide html/js code examples, and I will open a new PR in ttk-data that runs a small demo.

If you think this is only useful for a small ttk-target audience, I could also provide this filter as a standalone ParaView plugin. What are your thoughts on this?

Best
Jonas


julien-tierny:
looks cool, thanks!

PR #597, opened 2021-04-14
Fix ttkIdentifiers/GenerateIds compatibility

pierre-guillou:
The `ttkIdentifiers` filter is soon to be deprecated in favor of `GenerateIds`/`GenerateGlobalIds`. However, the latter generate `vtkIdTypeArrays` while `ttkIdentifiers` only generates a `ttkSimplexIdTypeArray` (an 32-bit `int` by default, instead of a 64-bit `long long`). Since the client filters (`HarmonicField`, `DistanceField`, `IntegralLines` and `TopologicalSimplification`) expect a SimplexId buffer, we can't replace `ttkIdentifiers` directly.

This PR introduces a new mechanism to ensure compatibility between the Identifiers array type. A new method, `ttkAlgorithm::GetIdentifierArrayPtr`, returns a `SimplexId` pointer to either the Identifier array or to a spare storage if the Identifier array is not a `ttkSimplexIdTypeArray`. This new method replicates the signature of `ttkAlgorithm::GetOptionalArray` plus the spare storage (a vector of SimplexIds).

The aforementioned client filters have been modified to use this new method instead of `GetOptionalArray`. Other filters that extract an identifier field from a Persistence Diagram or a Morse-Smale Complex have been left untouched.

No modification observed when running the ttk-data states.

Enjoy,
Pierre

### DIFF
--- a/core/base/distanceField/DistanceField.h
+++ b/core/base/distanceField/DistanceField.h
@@ -52,7 +52,7 @@ namespace ttk {
       return triangulation->preconditionVertexNeighbors();
     }
 
-    inline void setVertexIdentifierScalarFieldPointer(void *data) {
+    inline void setVertexIdentifierScalarFieldPointer(SimplexId *const data) {
       vertexIdentifierScalarFieldPointer_ = data;
     }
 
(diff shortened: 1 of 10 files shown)


julien-tierny:
awesome, thanks!

PR #598, opened 2021-04-14
Write & read preconditioned explicit triangulation to file

pierre-guillou:
This PR adds two new modules:
* `ttkTriangulationWriter` writes a preconditioned explicit triangulation to disk from a VTU,
* `ttkTriangulationReader` loads the corresponding file and applies it on a VTU, skipping preconditioning.

Generated files end with the `tpt` extension (for TTK Preconditioned Triangulation). They use an ad-hoc binary format that stores contiguously every vector used by the Explicit Triangulation.

The triangulation does not need to be fully preconditioned to be written to disk. Besides, if two VTUs have an identical mesh (e.g. tetrahedralized VTIs with identical grid dimensions), they should be able to share a triangulation file.

Benchmarks on an hard drive disk (not an SSD) show that reading a triangulation file is about twice as fast as preconditioning the same triangulation (for the `ttkMorseSmaleComplex` filter, whose preconditioning step is very consequent). For reference, the writing step takes a little more time than the reading one (about 2/3 of the preconditioning step duration).

Enjoy,
Pierre


julien-tierny:
it looks good, although I'd still suggest to move the core (``core/base/``) IO to the abstract triangulation instead of just the explicit one.

PR #599, opened 2021-04-14
Fix more clang-tidy warnings

pierre-guillou:
This PR fixes some of the clang-tidy warnings found in the `check_code` job. As a reminder, [clang-tidy](https://clang.llvm.org/extra/clang-tidy/) is a tool under the CLang/LLVM umbrella that detects faulty/bug-prone or inefficient code patterns in large codebases. It features numerous types of [warnings](https://clang.llvm.org/extra/clang-tidy/checks/list.html) that can be controlled through a `.clang-tidy` file at the root of a C/C++ project.

This PR focuses on bug-prone and performance warnings:
* passing by reference has been preferred over passing by copy for heap-allocated objects (mainly for `std::string`),
* vectors have been pre-allocated before being filled in trivial loops (without conditionals),
* identical branches have been merged,
* one-character strings were replaced with chars,
* explicit set to vector conversions have been modified to use the relevant `std::vector` constructor.

Additionally, some warnings found by clang's static analyzer have also been fixed:
* some variables have been initialized,
* more checks have been put in place to avoid segfaults.

No modification has been observed when running the ttk-data states.

Enjoy,
Pierre


julien-tierny:
ok, looks good.

PR #600, opened 2021-04-15
Remove assert using q.n

jspanchu:
Quad was reimplemented as `std::array` container in #592. It does not have a member named 'n'.

### DIFF
--- a/core/base/quadrangulationSubdivision/QuadrangulationSubdivision.h
+++ b/core/base/quadrangulationSubdivision/QuadrangulationSubdivision.h
@@ -723,7 +723,6 @@ int ttk::QuadrangulationSubdivision::subdivise(
   }
 
   for(auto &q : outputQuads_) {
-    assert(q.n == 4); // magic number...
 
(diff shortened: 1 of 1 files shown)


pierre-guillou:
Good catch! I forgot to remove this `assert` when I wrote #592. I tend to always compile TTK in Release mode, so the asserts are rarely put to test. Guess I'll try to avoid them in the future.
Pierre

jspanchu:
> Good catch! I forgot to remove this `assert` when I wrote #592. I tend to always compile TTK in Release mode, so the asserts are rarely put to test. Guess I'll try to avoid them in the future.
> Pierre

Ah! It's okay.  :)

I compiled in Debug mode for development and think asserts are useful, especially in code that implements complex computational geometry algorithms.

I guess this is one reason the CI builds could also compile TTK in Debug mode too?

pierre-guillou:
Good suggestion, I'll add the Debug mode to my GitHub Actions matrix.

julien-tierny:
thanks!

PR #601, opened 2021-04-16
GitHub Actions: Also test Debug builds

pierre-guillou:
Following #600, this PR adds Debug checks in the GitHub Actions workflow. This should catch invalid assert errors disabled by the Release mode.

Enjoy,
Pierre


julien-tierny:
awesome, thanks

PR #602, opened 2021-04-16
Fix #585

### LINKED ISSUE #585: TODO: ttkMacros.h imports headers which can be replaced with forward declarations
The `ttkMacros.h` file imports two headers which should be replaced with forward declarations. This, however, will require to update a lot of modules which now explicitly have to import these headers. The corresponding headers already import the `vtkObjectFactory.h` header, so we also have to manually add this header in a lot of cpp files. Since this a larger task that effects a lot of files I put this here on the Todo-Stack.

Best
Jonas

pierre-guillou:
This PR replaces the includes in `ttkMacros.h` with forward declarations. The includes were put back in the `.cpp` files when needed. It fixes #585.

No modification observed in the ttk-data states.

Enjoy,
Pierre

### DIFF
--- a/core/vtk/ttkAlgorithm/ttkMacros.h
+++ b/core/vtk/ttkAlgorithm/ttkMacros.h
@@ -1,7 +1,7 @@
 #pragma once
 
-#include <vtkIdTypeArray.h>
-#include <vtkIntArray.h>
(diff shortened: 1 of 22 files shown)


julien-tierny:
looks good.

PR #603, opened 2021-04-16
TopologicalCompression: Use ZFP fixed-accuracy mode, headers

pierre-guillou:
This PR improves the TopologicalCompression filter by using the ZFP fixed-accuracy mode instead of the fixed-rate mode. ZFP headers are also inserted in the ZFP buffer to encode the compression parameters more accurately. As a consequence, the main ZFP user parameter has been changed from "ZFP Bit Budget" to "ZFP Absolute Error Tolerance", with a default value of 50. With this parameter, the new compressed files should be smaller than before.

Additionally, in a (quite futile) attempt to reduce compilation times, some methods were de-templated and moved from headers to `TopologicalCompression.cpp`.

Please note that this PR breaks the backward compatibility with the old `.ttk` files. Old files won't be read by this updated filter.

The `persistenceDrivenCompression` state from ttk-data has been remade accordingly. A ttk-data PR will shortly follow.

Enjoy,
Pierre


julien-tierny:
looks good, thanks!

PR #604, opened 2021-04-16
Do not modify input with FTM Tree

### LINKED ISSUE #570: FTMTree - Segmentation Arrays broken
Hi,
@JonasLukasczyk and I found some weird behaviour in the FTMTree filter, which we would consider a bug.

**Describe the bug**
The segmentation output of FTMTree has point data arrays "SegmentationId" and "RegionSize" along with others. In Paraview, it does not have any cell data arrays attached (see Screenshot below). However, when printing the vtk object from the C++ layer, the two arrays "SegmentationId" and "RegionSize" exist as cell data and point data (screenshot of output attached as well). While trying to find the problem, Jonas and I removed lines 187 to 192 from the ttkFTMTree.cpp, so the output should only be a shallow copy of the input. But there was still a point data array "ttkVertexScalarField" added. This seems to be a violation of the pipeline paradigm.
(@CharlesGueunet )

**To Reproduce**
You need some other (maybe C++) filter applied after FTMTree on the segmentation, which prints its input vtk object. For the pipeline violation, it should (hopefully) be enough to run FTMTree with the described changes on any input.

**System:**
 - OS: macos 11.2.1
 - Compiler GCC 9.3.0
 - TTK dev branch on my fork, currently 10 commits behind.

**Screenshots**
Screenshot 2021-02-24 at 11 00 44
Screenshot 2021-02-24 at 11 07 04

CharlesGueunet:
Dear Julien,

The way the FTM wrapper is done requires to add an identifier array on the data to process.
As noted in #570, the way it was done lead to a violation to the pipeline mechanism where FTM added a new array on the input (for unstructured meshes). This MR solve the issue.

Best,
Charles

fix #570

### DIFF
--- a/core/vtk/ttkFTMTree/ttkFTMTree.cpp
+++ b/core/vtk/ttkFTMTree/ttkFTMTree.cpp
@@ -76,10 +76,12 @@ int ttkFTMTree::RequestData(vtkInformation *request,
     // This data set may have several connected components,
     // we need to apply the FTM Tree for each one of these components
     // We then reconstruct the global tree using an offest mecanism
-    identify(input);
+    auto inputWithId = vtkSmartPointer<vtkUnstructuredGrid>::New();
+    inputWithId->ShallowCopy(input);
+    identify(inputWithId);
 
     vtkNew<vtkConnectivityFilter> connectivity{};
-    connectivity->SetInputData(input);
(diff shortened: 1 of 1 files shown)


julien-tierny:
looks good, thanks.

PR #606, opened 2021-04-21
Morse-Smale Complex optimizations

pierre-guillou:
This PR brings two optimizations to the Morse-Smale Complex module:
* the `separatrixFunctionMin`, `separatrixFunctionMax` and `separatrixFunctionDiff` cell arrays are now partially computed in the VTK layer, in order to remove the dependency of base layer methods to the scalar field type. Now, the base layer computes for each separatrix the vertex id of maximum and minimum order, storing the result per separatrix instead of per separatrix cell. The VTK layer expands these compact vectors to separatrices cells (the diff is also computed here). Scalar field template parameters have been removed in relevant methods, leading to a (small) decrease in compilation times. Also, the base layer allocations for these vectors is reduced.
* the VTK separatrices cells are now using the 32-bit storage (when `ttk::SimplexId` is 32-bits). As a consequence, less time is spent allocating those vectors in the base layer, especially for 2-separatrices.

The full Morse-Smale Complex computation of `ctBones.vti` (256^3) using the standalone saw a 15% performance improvement (in the base layer, +20% on the descending 2-separatrices and +25% on the ascending 2-separatrices). And my 64GB RAM desktop station doesn't swap anymore...

No changes observed in the ttk-data states using the Morse-Smale Complex filter.

Enjoy,
Pierre


julien-tierny:
ok, looks good.
thanks

PR #607, opened 2021-04-21
CMake: Don't display warning if optional dependency is not found

### LINKED ISSUE #571: Avoid non-warning messages for embree (and other dependencies)
**Problems**

If CMake does not find embree (or other libs like zlib for that matter) it throws warning.

The warning message is this, which comes from [config.cmake:113](https://github.com/topology-tool-kit/ttk/blob/02d084deff10af209c43079bf7d61a596bc182f1/config.cmake#L113):
~~~
CMake Warning at thirdparty/ttk/ttk-vcs/config.cmake:113 (find_package):
  By not providing "FindEMBREE.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "EMBREE", but
  CMake did not find one.

  Could not find a package configuration file provided by "EMBREE" (requested
  version 3.4) with any of the following names:

    EMBREEConfig.cmake
    embree-config.cmake

  Add the installation prefix of "EMBREE" to CMAKE_PREFIX_PATH or set
  "EMBREE_DIR" to a directory containing one of the above files.  If "EMBREE"
  provides a separate development package or SDK, be sure it has been
  installed.
~~~

This makes it harder to find other warnings, which could be critical.
For example, my IDE panel currently alerts me with a yellow warning icon, and I have to scroll down to make sure there's no other warning:
image

Steps to reproduce:
1. No embree visible to CMake
2. Run CMake

**The solution you'd like**

As TTK happily chooses to proceed without embree, the warning shall not be displayed.
What do you guys think?

**Alternatives you've considered**
Living with this non-warning messages is of course possible. But it is a waste of an opportunity to find more important warnings.

Thank you for your attention :-)

pierre-guillou:
TTK uses optional dependencies through CMake. However, due to how CMake's `find_package` works, CMake either displays some message such as

> Found ZFP: /usr/lib64/cmake/zfp/zfp-config.cmake (found version "0.5.5")

if the package is found, a warning if not, or nothing.

Since those dependencies are optional, the CMake warnings clutter the CMake log and can hide errors.
This PR sets all `find_package` calls relative to optional dependencies to `QUET`: no warning will be displayed if the dependency is not found.

Additionally, new/improved messages are displayed when those dependencies are found: the dependency version and a path to either the/one (shared) library or the include directory for header-only dependencies.

The dependencies affected are:
* Boost,
* ZFP
* ZLib,
* Embree,
* Eigen & Spectra,
* WebsocketPP,
* Python/NumPy

This PR fixes #571.

Enjoy,
Pierre


julien-tierny:
looks great, thanks!

PR #608, opened 2021-04-23
[assignmentSolver] Add module

MatPont:
This PR adds the assignmentSolver module, it includes the Munkres implementation of TTK and add an Auction implementation (working directly on a cost matrix) and a Exhaustive one.

It renames the auction of TTK as persistenceDiagramAuction since this implementation is a particular case of the Auction working only for persistence diagrams.

It removes Munkres in BottleneckDistance and replace it with a call to AssignmentSolverMunkres in the assignmentSolver module.

Mathieu.


core/base/assignmentSolver/AssignmentAuction.h
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    inline std::vector<double> &getPrices() {
+      return goodPrices;
+    }
+
+    template <class vecType>
+    void printVector(std::vector<vecType> &vec) {
+      for(auto valTemp : vec)
+        std::cout << valTemp << " ";
```
julien-tierny: please use the ttk log stack (``printMsg``) instead of ``std::cout`` or ``std::cerr``.
see ```core/base/scalarFieldCriticalPoints/ScalarFieldCriticalPoints.cpp``` for an example.

core/base/assignmentSolver/AssignmentExhaustive.h
```diff
@@ -0,0 +1,362 @@
...
+        std::cout << "{";
+        for(unsigned int i = 0; i < vecTemp.size(); ++i) {
+          auto valTemp = vecTemp[i];
+          std::cout << valTemp;
+          if(i != vecTemp.size() - 1)
+            std::cout << ",";
+        }
+        std::cout << "}, ";
+      }
+      std::cout << "}";
```
julien-tierny: same remark as before (use ``printMsg``).

core/base/assignmentSolver/AssignmentSolver.h
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    virtual inline std::vector<std::vector<dataType>> *getCostMatrixPointer() {
+      return &costMatrix;
+    }
+
+    void printTableVector(std::vector<std::vector<dataType>> &table) {
+      for(auto vecTemp : table) {
+        for(auto valTemp : vecTemp) {
+          std::cout << valTemp << " ";
```
julien-tierny: same comment about ``printMsg`` (to propagate through the entire PR).

core/base/assignmentSolver/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+ttk_add_base_library(assignmentSolver
+  SOURCES 
+    AssignmentSolver.cpp
+  HEADERS 
+    AssignmentSolver.h
+    AssignmentAuction.h
+    #AssignmentExhaustive.h
```
julien-tierny: Why not building the exhaustive version?

julien-tierny (changes requested):
hi mathieu, 
thanks a lot for this PR. this looks great and nearly ready to go!
could you just address my small remarks regarding the logging and the exhaustive solver?
thanks!

core/base/assignmentSolver/CMakeLists.txt
MatPont: Indeed, it is a mistake, will be fixed in the next commit!

core/base/assignmentSolver/AssignmentSolver.h
MatPont: Ok thanks! I will fix that in the next commit!

core/base/assignmentSolver/AssignmentAuction.h
```diff
@@ -0,0 +1,356 @@
...
+    inline std::vector<double> &getPrices() {
+      return goodPrices;
+    }
+
+    template <class vecType>
+    void printVector(std::vector<vecType> &vec) {
+      std::stringstream ss;
+      for(auto valTemp : vec)
+        ss << valTemp << " ";
+      printMsg(ss.str(), debug::Priority::VERBOSE);
```
julien-tierny, line 84: thanks!

core/base/assignmentSolver/AssignmentAuction.h
```diff
@@ -0,0 +1,356 @@
...
+      printMsg(" ------------------------------ ", debug::Priority::VERBOSE);
```
julien-tierny: actually, you have pre-specified separators available for you. instead of this line, you can use for instance:
``this->printMsg(ttk::debug::Separator::L1);`` which will have a similar effect (but with a better integration in the overall logs). checkout the ``Debug.h`` file to see the available separators. if you could spread that through the PR, that'd be awesome.

julien-tierny (changes requested):
awesome thanks, if you could use the dedicated mechanism for separators, that'd be awesome :)

core/base/assignmentSolver/AssignmentAuction.h
MatPont: Ok thanks for these information about printMsg, I will fix this in the next commit!

julien-tierny:
@pierre-guillou can you give us some feedback regarding the ci failures?
thanks!

pierre-guillou:
Yes! It's a problem with a third party Action I'm using to fetch pre-compiled ttk-paraview packages. I was tracking the `master` branch and a commit yesterday broke everyone of my GitHub Actions tests. The solution is to rollback to a tagged version of this Action (for the time being). PR soon.

julien-tierny:
@MatPont 
I have just merged pierre's hotfix for the CI.
could  you synchronize your branch with ttk's dev branch?
thanks!

julien-tierny (approved):

julien-tierny:
looks ready to go. rock'n'roll!
thanks!

PR #609, opened 2021-04-29
WIP: Merge tree planar layout

MatPont:
This PR adds a merge tree planar layout as an option of the planarGraphLayout filter of TTK.

By selecting the nodes and arcs output (while pressing ctrl) of the FTMTree filter of TTK you can call the planarGraphLayout filter and set the parameter "Input is a merge tree" to 1.
It will display a planar layout customized for merge tree (with some parameters, like a threshold of persistence for selecting important pairs that will be much more visible, or spacing parameters for example).


core/base/planarGraphLayout/FTMTreePPCompute.h
```diff
@@ -0,0 +1,248 @@
+/// \ingroup base
+/// \class ttk::FTMTreePPCompute
+/// \author Charles Gueunet <charles.gueunet@lip6.fr>
```
julien-tierny: are these functions taken from charles's code?
are there code duplicates?

core/base/planarGraphLayout/MergeTreeVisu.h
```diff
@@ -0,0 +1,2060 @@
...
+
+    int numInputs = trees.size();
+    int NumberOfBarycenters = barycenters.size();
+    bool clusteringOutput = (NumberOfBarycenters != 0);
+    NumberOfBarycenters
+      = std::max(NumberOfBarycenters, 1); // to always enter the outer loop
+    if(not clusteringOutput)
+      numInputs = 1;
+
+    vtkSmartPointer<vtkUnstructuredGrid> vtkMatching
```
julien-tierny, line 279: yikes :(
this is unfortunately breaking the API structure :(
the base code classes should *NOT* depend on VTK!

core/base/planarGraphLayout/MergeTreeVisu.h
```diff
@@ -0,0 +1,2060 @@
...
+    }
+
+    if(verbose > 0) {
+      std::stringstream ss;
+      ss << "AVOID CROSSING  = " << t_avoid.getElapsedTime();
+      printMsg(ss.str(), debug::Priority::VERBOSE);
+    }
+
+    if(verbose > 0)
+      printMsg("================================", debug::Priority::VERBOSE);
```
julien-tierny, line 1814: same remark as in PR #608 regarding separators.

julien-tierny (changes requested):

julien-tierny:
hi mathieu, 

thanks a lot for this PR!
this is great stuff :)
however, it seems it still requires some discussion and a few things will need to be reworked (code distribution with FTM, improve the vtk/base separation, etc.). let's talk about it together soon!
best,

core/base/planarGraphLayout/MergeTreeVisu.h
julien-tierny: @pierre-guillou @CharlesGueunet 
the fact that the CI let that go through is a little concerning.
we may want to enforce that test (for instance, by building the vtk-less version of the library in the CI).

pierre-guillou: Well, it's difficult if the faulty headers are only included by source files in the VTK layer and not in the base layer, as this seems to be the case here. I'm afraid that even if we build TTK without VTK (only the base layer), these headers won't be included anywhere and there won't be any build error (I need to test that though).

One way to detect that is to automatically generate a header that includes every base layer header and use it, for instance, in the c++ base-only example. Unfortunately, due to the modularity of the base layer, this might be hard to automate.

core/base/planarGraphLayout/FTMTreePPCompute.h
MatPont: Indeed, it was a quick fix in my local version, I needed to modify this class in order to be used in my code, we need a better solution indeed!

core/base/planarGraphLayout/MergeTreeVisu.h
MatPont: Yes, actually at first it was in the vtk layer, but then I moved it in the base layer when merging with planarGraphLayout, we could split the file or let it in the vtk layer. Like you said, we wil discuss about it soon!

core/base/planarGraphLayout/MergeTreeVisu.h
MatPont: Ok I will fix this in both PR, thanks!

JonasLukasczyk:
Hi all,
thank you @MatPont for this PR!

I use the PlanarGraphLayout filter all the time for merge trees. In this case I select as the sequence array the scalars, and then I compute a spatial embedding with the calculator with the following formula: `ihat*Layout_y*SomeConstantScalingFactor + jHat*Scalar`. To be fair, this does not produce awesome (but ok) layouts. However, if nodes also have a branchID then the computed layout is really good. Unfortunately FTM tree does not compute those by default, so I compute branch decompositions with a Python filter. So the question is what does this PR do differently/better than this approach? I wonder if we should improve other parts of the code first (like adding branch decompositions to the FTM tree) instead of specializing the layout code.

Best
Jonas

PR #610, opened 2021-05-04
GitHub Actions: Rollback to a tagged version of dsaltares/fetch-gh-release-asset

pierre-guillou:
A recent commit in the `dsaltares/fetch-gh-release-asset` repository broke the Action. This PR switch from tracking the `master` branch to an older tagged version to get the working code back.

Enjoy,
Pierre


julien-tierny:
awesome, thanks!

PR #611, opened 2021-05-18
update Docker build scripts

chgarth:
Thanks for contributing to TTK!

Before submitting your pull request, please:

- [x] Review our [Contributor Guidelines](https://github.com/topology-tool-kit/ttk/blob/dev/CONTRIBUTING.md), in particular regarding code formatting (with clang-format) and continuous integration.

- [x] Please provide a quick description of your contributions below:

This updates the docker container build scripts to include newer Paraview versions, plus some other improvements.

PR #612, opened 2021-05-18
Remove GCC 11 from the GitHub Actions CI

pierre-guillou:
GCC 11 just arrived in the GitHub Actions `ubuntu-latest` virtual environment and it broke everything…

An include has been removed in a `libstdc++` header, I think, causing domino effect build errors: a VTK header is missing the `limits` include and, as a consequence, TTK fails to build (see #611 CI jobs).

To alleviate this issue, this PR removes GCC 11 (compiler, headers + libraries) from these CI jobs. This should work for the time being (but does not fix the underlying compatibility issue with GCC 11).

As a bonus, this PR sets a fixed version of `clang-check` and `clang-tidy`. And also introduces a new job for performing static analysis, also via `clang-tidy` (informational only).

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -78,6 +78,8 @@ jobs:
     - name: Install Ubuntu dependencies
       run: |
         sudo apt update
+        # workaround: remove GCC 11
+        sudo apt purge gcc-11 libgcc-11-dev libgfortran-11-dev libstdc++-11-dev
         # TTK dependencies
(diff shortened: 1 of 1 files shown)


pierre-guillou:
Related VTK issue: https://gitlab.kitware.com/vtk/vtk/-/issues/18194

julien-tierny:
great stuff, thanks!

PR #615, opened 2021-05-27
Persistence Sensitive Simplification

JonasLukasczyk:
Hi all,
this PR finally brings the `persistence sensitive localized topological simplification` procedure to TTK. In addition to the speedup due to synergies, it is now no longer necessary to first compute all persistence pairs and then pass filtered pairs to the simplification module. To this end I made the second input (the constraints) of the simplification module optional, and I added a persistence threshold. If no constraints are present, then the module uses the persistence sensitive simplification procedure to remove all pairs below the persistence threshold. If constraints are present then the module behaves as before.

Some notes however:
1. While working on the new code, I also had to modify some old LTS routines. I tested the module and it seems to work correctly, but since this touches a crucial part of TTK I recommend for several people to test this out before we merge it in.
2. Currently the module always shows the persistence threshold input in the paraview ui, even when constraints are present. Does anyone know how to automatically hide the threshold input if a second input is present?
3. If the part of the domain that needs to be simplified takes up the majority of the domain, then it would be faster to do the "trunk trick". This is not implemented right now, but I can add it in the future.

Enjoy!


lhofmann:
>   Does anyone know how to automatically hide the threshold input if a second input is present?

Something like this should work (not tested, see https://gitlab.kitware.com/paraview/paraview/-/blob/v5.9.1/Qt/ApplicationComponents/pqGenericPropertyWidgetDecorator.cxx#L89):

```xml
<DoubleVectorProperty name="PersistenceThreshold" ...>
    <Hints>
        <PropertyWidgetDecorator 
            type="GenericDecorator"
            mode="visibility"
            property="Constraints"
            value="null" />
    </Hints>
</DoubleVectorProperty>
```

JonasLukasczyk:
> > Does anyone know how to automatically hide the threshold input if a second input is present?
> 
> Something like this should work (not tested, see https://gitlab.kitware.com/paraview/paraview/-/blob/v5.9.1/Qt/ApplicationComponents/pqGenericPropertyWidgetDecorator.cxx#L89):
> 
> ```
> <DoubleVectorProperty name="PersistenceThreshold" ...>
>     <Hints>
>         <PropertyWidgetDecorator 
>             type="GenericDecorator"
>             mode="visibility"
>             property="Constraints"
>             value="null" />
>     </Hints>
> </DoubleVectorProperty>
> ```

Works, thank you!

julien-tierny:
hi jonas,

thanks a lot for the PR, this is great stuff!

I've been playing a bit with your branch:
- running the states from ttk-data works fine, nice!
- the direct persistence simplification works fine (a few adjustments would be needed for the GUI, but that's small-scale stuff)
- however, it is not clear to me how the original, generic use case can now been triggered in paraview. consider the following use case: the user computes the persistence diagram and proceeds to a specific selection of features to preserve (for instance using distinct thresholds for minima and maxima). how should they manipulate paraview's gui to provide this selection to TopologicalSimplification? (somehow, it seems that this branch is indeed backward compatible with ttk-data, but rebuilding these states from scratch does no longer seem supported).

thanks for your feedback.
julien

JonasLukasczyk:
Hi Julien,
since now the second input is optional PV will not automatically show you the input selection dialog. If you want to feed explicit constraints to the simplification then you need to right click the filter and select `Change Inputs`. As soon as a second input is present the filter will work as before.

If you do not like this, we could provide two xml wrappers on the same vtk class: one for the classical simplification (now called SimpWithConstraints or something), and one with only one input and a persistence threshold (SimpByPersistence). However, I prefer the way it is right now.

julien-tierny:
hi jonas,

thanks a lot for the feedback and sorry for the late reply.

actually, I'd vote for the dedicated xml for `PersistenceSimplification`. 

changing the input of `TopologicalSimplification` requires to commit the computation first without constraints (by hitting the apply button) before the input changing option becomes available. 
that does not seem optimal. it requires users to perform an unwanted computation (without constraints) before having a chance to fix it.

I'd suggest to:
- rollback the original `Topological Simplification` xml file
- add a new `Persistence Simplification` xml file
- add a few options to `Persistence Simplification`: 
    - drop down menu to choose the extremum type to simplify (minima, maxima, both)
    - checkbox to express the persistence threshold as a ratio of the function range (useful when you have no clue)
- add a warning (in the base code) in the original use case of `Topological Simplification` (when constraints are detected, irrespective of the backend) to promote `Persistence Simplification`. something like "Did you know? If you just need to simplify your data by persistence, simply use `Persistence Simplification`. It is both easier to use and faster!"

if you want I can push that to your branch later this week. just let me know.

best,

julien-tierny:
PS: having a separate vtk class (using the same base layer class) is another option (may be even cleaner)

JonasLukasczyk:
Hi Julien,
you are right about the "first I need to apply before I can change the inputs" issue. That is a showstopper. This issue doesn't occur in the node editor, which is btw currently integrated into the PV source ;-). So there is a PV limitation that now forces us to provide two topological simplification filters (which I strongly dislike, but currently see no way around it). I guess the most consistent way of naming the filters would be to preserve the old name for the constraint-based simp, and naming the new filter `TopologicalSimplificationByPersistence`. If you and @pierre-guillou agree with that I will make the changes.

julien-tierny:
congrats on the node editor in PV sources! (I didn't know)

julien-tierny:
the naming convention works for me!

JonasLukasczyk:
The people at Kitware also think of this as an issue and suggested to open a discussion on discourse, which I did: https://discourse.paraview.org/t/change-filter-inputs-before-apply/7346

The question is should we keep one filter and deal with the fact that people with fixed constraints have to run the filter once in persistence mode, or should we introduce two filters where we delete one as soon as PV supports this. In the second option I'm worried about backwards compatibility.

JonasLukasczyk:
@julien-tierny let me know what you think because this PR is basically ready except these cosmetic changes.

julien-tierny:
hi jonas,

thanks for the feedback.
I'd vote again for the two xml files for the time being. 
at the moment, we have no way to know when and how this will be changed in paraview and we don't want to drag that double-computation showstopper until then.

thanks a lot!

julien-tierny:
hi @JonasLukasczyk,

I gave a try again at your branch with the separate module and I find it wonderful :)

any chance you could add a drop down menu to `TopologicalSimplificationByPersistence` in order to choose the type of critical point to cancel (minimum, maximum or both)? (there are use cases where it could be handy to cancel only one type of extrema).
thanks!

JonasLukasczyk:
Hi Julien,
as requested I added the pair type selection. I tested this with a lot of data (inc ttk-data) and it seems to work, but I had to touch again a lot of code. So please test this yourself (at least with ttk-data) before we merge this in.

Best
Jonas

julien-tierny:
hi jonas, 

thanks a lot for this addition!
this module is really great :)

when testing it, I observed a regression though.
for the same constraints, the output of this new LTS is no longer identical to the version of LTS that sits in the dev branch of ttk.

for instance, consider the example from ttk-data `states/BuiltInExample1.pvsm`, then change the persistence threshold from 0.02 to 0.05 (in the `PersistenceThreshold` node of the pipeline) and you will observe two different results between these two versions of LTS:

- with ttk:dev's LTS:
lts_ttkDev

- with this PR's LTS:
lts_pltsBranch

do you have any idea of what could be the reason?

best,

julien-tierny:
PS: looking at the console, it looks like you switched the processing order for minima and maxima. could that simply be the reason?

JonasLukasczyk:
Hi Julien,
thanks for the check. This was a bug in the scalar flattening procedure that processed the field in min-max order despite that the field is now simplified in max-min order (which circumvents some unnecessary operations). It should be fine now, at least all test cases look correct (e.g., by swapping the simp module in the build in example with PLTS and only processing numerically perturbed scalar field (ditching the order field for testing)). However, since we changed the order in which we simplify the field, it is possible that the locations of saddles changes, but this is well understood.

Best
Jonas

julien-tierny:
hi jonas,

thanks a lot for fixing that so quickly
I confirm this is fixed :)

now I did the following test:
- compute the persistence diagram
- simplify the original field with `TopologicalSimplification` (with the LTS backend) and rename the output filed "LTS" (with the calculator)
- simplify the original field with `TopologicalSimplification` (with the legacy backend) and rename the output field "legacy" (with the calculator)
- append the attributes of the two simplifications and compare the arrays "LTS" and "legacy" with `TTK LDistance`. the reported distance is 0: the two backends generate exactly the same thing.

now if I simplify with `TopologicalSimplificationByPersistence` and I compare its output with one of the above fields (with the same procedure), the distance reported by `TTK LDistance` is no longer 0 (the output actually differ)

my impression is maybe because you decided to switch the order around for the simplification of the mins and maxes. 
but this is not clear for me why you did this.

if there's some performance gains, are these worth breaking the strict compatibility with the other backends?
would that make sense to switch that order as well in the other backends?

I'm sorry for being picky here but as you know, data simplification is a rather central component of TTK and I want to make sure that everything is well consistent there. I hope you understand.

thanks!

JonasLukasczyk:
Hi Julien,
I didn't fix the previous issue by reverting to the old order, instead I updated the scalar flattening procedure to respect the new order (see commit changelog). So the order has no impact on the results. The short answer why LTS simplifies maxima first now is:
* All LTS procedures are written for maxima simplification.
* To simplify minima one first needs to invert the order, compute the simplification, and then revert the order.
* Now that one can skip the minima computation, the invert operations can also be skipped. So I rearranged procedure calls (it is cleaner now).

The new issue you observe results from the fact that the PLTS procedure is not strictly equal to LTS. I don't even know if it can be made strictly the same. My intuition says no since PLTS needs to compute persistence on the fly (so scalars values need to be perturbed after each propagation type, in contrast to LTS where the scalars can be flattened together at the end). I can investigate this further, but I think this is a known limitation of PLTS.

JonasLukasczyk:
Oh, also note that the PLTS persistence threshold must be strictly exceeded, whereas the common LTS pipeline assumes the persistence threshold to be included.

julien-tierny:
hi jonas,

thanks a lot for the clarifications. it makes sense now, although we might want to carefully note that for the future, as the difference may be a source of confusion later on for users.

I added a few TIP messages to promote this new module.
I'll merge after CI completion.

thanks a lot!

JonasLukasczyk:
hi Julien,
please hold off for one more day. I want to make sure that the LDistance issue you reported is really not a bug.

julien-tierny:
you got it, thanks for you careful efforts :)

JonasLukasczyk:
Hi Julien,
I double checked and I can not reproduce your issue. For all my test cases the PLTS procedure produces the exact same output (order and scalar) as the LTS procedure with equal constraints (recall that the threshold given to PLTS needs to be strictly exceeded, whereas the threshold filter includes the lower bound). So my guess is the threshold edge case caused the difference, but if not please let me know.

Although everything looks good on my end, I came across the following oddity:
Using topo simp (LTS or Legacy) on the foot dataset with a threshold > 10 (not >= 10) will still leave a pair with persistence 10. It is weird that both LTS and Legacy are not able to remove that. PLTS with >10 also keeps that one pair (at least everything is consistent xD). Please check if you can reproduce that issue. My guess is that here we actually encounter the rare situation where the persistence of a maxima-saddle pair (that before exceeded the threshold) is lowered below the threshold after the minima removal. Running PLTS/LTS/Legacy again removes that pair. Lets keep an eye on that.

I personally do not like the warning in the simp module that one could use PLTS every time it is executed. In PV the modules are already right next to each other (also in the shortcut window), so I would suggest to remove them again.

Thanks again for all your input! This drastically improved the quality of the module/code. Let me know how your tests go!

julien-tierny:
hi jonas,

> I double checked and I can not reproduce your issue. 

actually, I ran that experiment again and I confirm this was a >10 vs >= 10 issue (shifting the thresholding with the generic module to 10.0001 solved the issue).

> So my guess is the threshold edge case caused the difference, but if not please let me know.

that's confirmed then and we can move on with this :)

> Using topo simp (LTS or Legacy) on the foot dataset with a threshold > 10 (not >= 10) will still leave a pair with persistence 10.

I investigated that and it happened to be a maximum-2saddle pair with an original persistence of 11. I believe it got lowered down to ten during the minimum removal phase, which is an understood aspect of the approach. 
so for me, that's all good :)

> I personally do not like the warning in the simp module that one could use PLTS every time it is executed. In PV the modules are already right next to each other (also in the shortcut window), so I would suggest to remove them again.

ok, I removed that but then updated the doxygen documentation to add pointers between the two classes.

so all in all, that PR looks like it's ready to be merged.
thanks for confirming you're ok with this and I go ahead with the merge.

thanks again for this awesome work :)

best,

JonasLukasczyk:
This now looks good to me. Also thanks to you for helping out!

julien-tierny:
alright, let's rock'n'roll then :)

PR #616, opened 2021-05-28
MorseSmaleComplex: Fix boundary critical points numbering on separatrices

pierre-guillou:
This PR fixes the numbering of boundary critical points on 1- and 2-separatrices.
Previously, saddle connectors were treated separately from ascending and descending 1-separatrices. 
For 2-separatrices, the source was not counted.

The ttk-data morseMolecule is affected by this change. A PR is coming soon.

Enjoy,
Pierre


julien-tierny:
looks great, thanks a lot!

PR #617, opened 2021-05-28
ExplicitTriangulation: Skip precondition if dataset empty

pierre-guillou:
This PR fixes potential segfaults by skipping all explicit triangulation preconditions if the input dataset has no vertices.

No modification detected on the ttk-data states.

Enjoy,
Pierre


julien-tierny:
thanks!

PR #618, opened 2021-05-28
CMake: per-module optional dependency management, FindSQLite3

pierre-guillou:
This PR modifies the CMake code to do per-module dependency management. Previoulsy, enabling or disabling an optional dependency with CMake (via the `TTK_ENABLE_` keys) would cause, through preprocessor defines, a rebuild of the entire library. With this PR, only the relevant modules should be rebuilt.

Also, since we now use CMake 3.14 and above, I switched the code handling the SQLite dependency to the corresponding CMake FindSQLite3 module.

As a bonus, the display of Spectra headers directory during CMake configuration has been fixed (bug introduced in #607).

Enjoy,
Pierre


julien-tierny:
ok, looks good. thanks.

PR #620, opened 2021-06-03
ContourTreeAlignment: Use shared_ptr references to boost performance

pierre-guillou:
The ContourTreeAligment module uses `std::shared_ptr` (a reference-counted smart pointer) extensively.
Passing them by copy incur a performance penalty in the form of an atomic increment (at the beginning of the function call) and a decrement (at the end of the function scope).

This PR replaces all the `std::shared_ptr` copies in this module by constant references, as suggested by the clang-tidy lints [performance-for-range-copy](https://clang.llvm.org/extra/clang-tidy/checks/performance-for-range-copy.html) and [performance-unnecessary-value-param](https://clang.llvm.org/extra/clang-tidy/checks/performance-unnecessary-value-param.html).

The output of the ttk-data `contourTreeAligment.pvsm` state file is the same, while the time spent in the TTKContourTreeAlignment filter decreases from 120ms to 45ms.

Also, some run-time checks have been added to please the clang static analyzer.

@floWetzels Are you ok with these modifications? Is there a use-case of this module that requires the `std::shared_ptr` copies?

Enjoy,
Pierre

### DIFF
--- a/core/base/contourTreeAlignment/ContourTreeAlignment.cpp
+++ b/core/base/contourTreeAlignment/ContourTreeAlignment.cpp
@@ -1,10 +1,5 @@
 #include "ContourTreeAlignment.h"
 
-#include "float.h"
-#include <algorithm>
-#include <limits.h>
-#include <list>
-
 ///=====================================================================================================================
 /// branch decomposition
 ///=====================================================================================================================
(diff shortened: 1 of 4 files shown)


floWetzels:
> @floWetzels Are you ok with these modifications? Is there a use-case of this module that requires the `std::shared_ptr` copies?

I don't think so, as I can't think of any case where it would be necessary. But I would would have to look deeper into the code to be completely sure. If the example still works, it should be fine.

Thank you for the improvements!

julien-tierny:
alright, thanks!

PR #614, opened 2021-05-20
Progressive Topology integration

julesvidal:
Dear Julien,

This PR  integrates the work related to the following publication : 

A Progressive Approach to Scalar Field Topology
Jules Vidal, Pierre Guillou, Julien Tierny
IEEE Transactions on Visualization and Computer Graphics, 2021

It adds the option to choose a progressive algorithm for the computation of the critical points or the saddle-extremum persistence diagram when the input is defined on an implicit regular grid. The progressive approach is interruptible and resumable.

Added base modules:
multiresTriangulation
dynamicTree
progressiveTopology

The `progressiveTopology` base class is called from the base modules `persisenceDiagram` and `scalarFieldCriticalPoints.`

PR #621, opened 2021-06-03
ProgressiveTopology: Improve integration

pierre-guillou:
Following #614, this PR improves the integration of the ProgressiveToplogy module in TTK:
* dead code has been removed,
* `std::cout`, `std::cerr` were replaced with the Debug module log API,
* ProgressiveTopology now use the parallel sort and the OpenMP lock wrapper implementation from the Common module
* by using order fields of type `SimplexId` instead of templated scalar field and offset field, the majority of the ProgressiveTopology methods were detemplated, leading to build time improvements: a full rebuild of the `ttkPersistenceDiagram` target (base layer + VTK layer) decreased from 5min to 4min.
* clang-tidy warnings in MultiresTriangulation were fixed by using zero-initialized `std::array`s instead of C-arrays.

No change has been observed on the ttk-data states using the `ScalarFieldCriticalPoints` filter. Persistence diagrams still seems identical to FTM.

Enjoy,
Pierre


julien-tierny:
ok, looks good.
thanks!

PR #623, opened 2021-06-08
own namespace for lts

JonasLukasczyk:
Hi Julien,
this PR just adds a namespace around LTS data structures.

### DIFF
--- a/core/base/localizedTopologicalSimplification/LocalizedTopologicalSimplification.h
+++ b/core/base/localizedTopologicalSimplification/LocalizedTopologicalSimplification.h
@@ -1476,4 +1476,4 @@ namespace ttk::lts {
     };
 
   }; // class
-} // namespace ttk
+} // namespace ttk::lts
--- a/core/base/localizedTopologicalSimplification/LocalizedTopologicalSimplification.h
+++ b/core/base/localizedTopologicalSimplification/LocalizedTopologicalSimplification.h
@@ -32,1448 +32,1466 @@
 
 #include <Propagation.h>
 
-namespace ttk::lts {
+namespace ttk {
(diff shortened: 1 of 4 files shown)


JonasLukasczyk:
Hi @julien-tierny,
my original code was rejected because I used the abbreviated version of a nested namespace (supported since C++17):
`namespace ttk::lts {` which is equivalent to `namespace ttk { namespace lts {`.
However, my version was rejected by the clang format check (explicitly stating C++17 required). Is it not safe to require that for ttk? This would safe a lot of unnecessary indent.

julien-tierny:
the support for clang-formatting in the CI has been added by @pierre-guillou 
pierre, could you comment on that?
should we upgrade the version of clang-format we use (I remember it's already a more recent version than the default one provided on github's CI VMs)

pierre-guillou:
Well, if I understand correctly, it was `clang-check` that rejected the nested namespace construct, since it's a C++17 feature (and we use C++11 in TTK). I use `clang-format` v11 for checking the code formatting but the v12 may be available. Yesterday, I upgraded my laptop with Clang 12 and tested `clang-format` on TTK but fortunately, no change occurred on the code.

JonasLukasczyk:
Should I revert to the original abbreviated version and then we change the CI until it accepts it?

pierre-guillou:
While I would gladly use C++17 instead of C++11 in TTK,  I fear that accepting C++17 features might break on old compilers, especially when we explicitly require C++11 (at the top of `config.cmake`)

JonasLukasczyk:
Then it seems this PR is ready.

julien-tierny:
alright, let's go then.
thanks!

Issue #619, opened 2021-06-03
ParaView Layer CMake Cleanup


JonasLukasczyk:
Hi all,
in the PV layer we basically have per filter "always" two files, for example:
CinemaProductReader.xml
TTKFilter.cmake

but the TTKFilter.cmake seems to be superfluous since its content (`ttk_register_pv_filter(ttkCinemaProductReader CinemaProductReader.xml)`) could be generated automatically based on the folder/xml name. I realize there are some folders that do not have an xml file, in which case the content is reduced to `ttk_register_pv_filter(ttkUserInterfaceBase "")`, which can also be generated automatically.

My suggestion is to get rid of all these TTKFilter.cmake files and handle that automatically in cmake. We could even get rid of the folders and just list all xml files directly in the paraview folder.

In my opinion this would cleanup the code base. Any thoughts on this?

julien-tierny:
simpler looks better :)
@CharlesGueunet: what do you think?

PS: `ttk_register_pv_filter(ttkUserInterfaceBase "")` looks like a hiccup (`ttkUserInterfaceBase` has nothing to do with paraview plugins)

CharlesGueunet:
agreed, and having a collection of xml instead of the actual folder hierarchy seems really cool.

We will need to be careful how to process ttk modules with no ParaView counterpart (like containers or utilities for example)

JonasLukasczyk:
This issue is resolved with PR #624

PR #624, opened 2021-06-08
WIP: new paraview layer structure

JonasLukasczyk:
Hi @julien-tierny, @pierre-guillou, and @CharlesGueunet,
this PR addresses issue #619.

In short, it removes all module folders from the paraview directory and instead provides a single directory called `paraview/xmls` that contains all `module.xml` files. I did not directly put these files into the paraview dir since there are two special folders: `singlePlugin` and `patch`. So the pv folder is now nice an clean. Note, there are some modules that did not have an xml file, but they still need to be registered at this step. I solved this issue by providing an "empty" xml file for these modules (it must at least contain the following string `<ServerManagerConfiguration></ServerManagerConfiguration>`). Similar to before where every module had to have a directory in the paraview folder, now every module has to have an xml file in the `paraview/xmls/` folder.

There are two TODOs before this can be merged in:
1. The create/rename/delete/clone scripts need to be updated
2. There was a `Compatibility.xml` file used to print some deprecation warnings that needs to be handled in a special way. I could do this, but I thought this might be a good time to remove these messages. So currently this xml file is removed.

Let me know what you think!

Best
Jonas


paraview/CMakeLists.txt
```diff
@@ -57,3 +54,4 @@ install(
   DESTINATION
     "${TTK_INSTALL_PLUGIN_DIR}"
   )
+
```
vollia: avoid newline at the end of the file

CharlesGueunet (approved):

paraview/CMakeLists.txt
JonasLukasczyk: done ;-)

CharlesGueunet:
This new mechanism is way cleaner. I just find a bit strange the almost empty xml files for `ProgramBase`, ...

JonasLukasczyk:
I agree with the empty files, they are necessary to still let the PV layer know which modules to build. Alternatively we could implement the following (more complicated logic):
For each directory in core/vtk register a pv_module and add the xml file if one exists. This way we do not need the empty files.

CharlesGueunet:
let's keep the files like that, the simpler the better.

paraview/xmls/UserInterfaceBase.xml
```diff
@@ -0,0 +1 @@
+<ServerManagerConfiguration></ServerManagerConfiguration>
```
julien-tierny: I don't believe that file is even needed. 
(it's supposed to be used by standalone gui programs)

paraview/xmls/ProgramBase.xml
```diff
@@ -0,0 +1 @@
+<ServerManagerConfiguration></ServerManagerConfiguration>
```
julien-tierny: same here (this vtk-layer class should only be used by standalone programs).
I think you can remove that.

julien-tierny (commented):
could you remove the two entries I mentioned and update the script files as well?

julien-tierny:
hi jonas,
thanks a lot for the PR.
ok for removing the compatibilty xml file.
if you could update the scripts and remove the two files I mentioned in my review, that'd be ready to ship.
thanks!

JonasLukasczyk:
I deleted the two files and tests run fine. I also updated the clone and delete scripts, but the rename script is too complicated for me to quickly make changes. Can someone else have a look?

julien-tierny:
@lhofmann 
to my understanding, the script `scripts/rename_plugins.py` was just a one-timer right?
(it had been used to migrate the name of the filters in the paraview plugin at some point)
I am not sure we need it anymore, do we?

JonasLukasczyk:
If we do not need it I will delete it and then I guess this PR is ready to go.

lhofmann:
Yes, `scripts/rename_plugins.py` is no longer needed.

JonasLukasczyk:
Maybe we should let the developers know that if they are merging their branch with the dev then it is possible that they need to manually place the xml files of untracked modules in the xmls directory.

julien-tierny:
alright, thanks for the confirmation @lhofmann and thanks @JonasLukasczyk for the cleanup!

PR #627, opened 2021-07-03
Extend Jacobi sets with Pareto flag

julien-tierny:
With this PR, Jacobi sets can be filtered out if they do not belong to Pareto sets.


julien-tierny:
ok!

PR #628, opened 2021-07-05
Update Spectra dependency to v1.0.0

pierre-guillou:
[Spectra ](https://github.com/yixuan/spectra/releases/tag/v1.0.0) has recently been updated to v1.0 with breaking API changes. 

Since the ArchLinux (AUR) package has been updated accordingly, this PR uses the new Spectra API in the EigenField module. This mostly consists in changes in enums and how to call the solver. The `config.cmake` file has also been modified to refuse old(er) Spectra versions.

A bug in the EigenField CMakeLists.txt has also been fixed: it used the old `SPECTRA_INCLUDE_DIR` variable which has been empty since switching from a user-installed Spectra to the CMake `find_package` scheme. The bug has been undetected because, on ArchLinux, Spectra headers are installed in `/usr/include` directly.

Enjoy,
Pierre


julien-tierny:
ok, looks good. thanks!

PR #629, opened 2021-07-09
Fix the way Graphviz is found and used

CharlesGueunet:
Dear Julien,

Here is a small change in the handling of Graphviz, which comes from: https://gitlab.kitware.com/paraview/paraview/-/merge_requests/5007
This change is required to use Graphviz in the ParaView superbuild.

Best regards,
Charles

@JonasLukasczyk FYI


julien-tierny:
looks good, thanks!

PR #630, opened 2021-07-10
Stable manifold persistence

julien-tierny:
New vtk filter which attaches persistence (and pair type) to the stable manifolds of the Morse-Smale complex.


julien-tierny:
rock'n'roll

PR #631, opened 2021-07-12
ArrayEditor XML fix

JonasLukasczyk:
Hi Julien,
this PR just fixes a warning that is shown in PV when someone uses the second input of the `ArrayEditor`. I didn't notice any impact on the ttk-data statefiles.

Best
Jonas


julien-tierny:
alright, looks good. thanks!

PR #632, opened 2021-07-12
PersistenceDiagramClustering: vtkMultiBlockDataSet ouputs, distances as FieldData

pierre-guillou:
This PR reworks the VTK layer of the PersistenceDiagramClustering module to put every diagram in the "Clustered Diagrams" and every centroid of the "Clustered Centroids" outputs into their own `vtkMultiBlockDataSet` block. The matchings have also been put into blocks, one per input diagram.

To do so, a common method converting a diagram or a centroid in its intermediate form of a vector of tuples to a VTU ensures that diagrams and centroids share the same PointData and CellData arrays. Those VTU are later translated for the "Inspect Matchings" and the "Clusters as Stars" display options using `vtkTransformFilter`. The diagonal has been added to each diagram/centroid.

Additionally, the distance results are now accessible in the VTK layer and have been put as FieldData of the Matchings output (duplicated per block since multi-block FieldData are not visible in ParaView).

Since the diagonal has been added in the diagrams, there will be some small graphical modifications in the two ttk-data states using this filter. 

As a bonus, a unused include has been removed from PointSetToCurve.

Enjoy,
Pierre

### DIFF
--- a/core/base/persistenceDiagramClustering/PDClustering.h
+++ b/core/base/persistenceDiagramClustering/PDClustering.h
@@ -60,6 +60,10 @@ namespace ttk {
       execute(std::vector<std::vector<diagramTuple>> &final_centroids,
               vector<vector<vector<vector<matchingTuple>>>> &all_matchings);
 
+    std::array<double, 3> getDistances() const {
+      return {this->cost_min_, this->cost_sad_, this->cost_max_};
+    }
+
     dataType getMostPersistent(int type = -1);
     vector<vector<int>> get_centroids_sizes();
     dataType getLessPersistent(int type = -1);
--- a/core/base/persistenceDiagramClustering/PersistenceDiagramClustering.h
+++ b/core/base/persistenceDiagramClustering/PersistenceDiagramClustering.h
@@ -66,6 +66,10 @@ namespace ttk {
       std::vector<std::vector<diagramTuple>> &centroids,
       std::vector<std::vector<std::vector<matchingTuple>>> &all_matchings);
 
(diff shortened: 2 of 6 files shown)


julien-tierny:
looks good, thanks a lot!
looking forward to giving it a try :)

julien-tierny:
it looks like applying `PersistenceDiagramDistanceMatrix` on the output `Clustered Diagrams` still segfaults :(

julien-tierny:
also, I have the impression that the distance that's attached in FieldData to the `Matching` output is incorrect
(it reports the same distance for all diagrams in the ttk-data example `persistenceDiagramClustering`)

pierre-guillou:
> it looks like applying `PersistenceDiagramDistanceMatrix` on the output `Clustered Diagrams` still segfaults :(

Yes, sorry... I did not test that after I switched from forwarding the input diagrams to reconstructing them from the intermediate representation... It seems that the `ttkVertexScalarField` array was missing there, causing a KAMIKAZE segfault and a silent error without in `PersistenceDiagramDistanceMatrix`. I have patches.

> also, I have the impression that the distance that's attached in FieldData to the `Matching` output is incorrect
> (it reports the same distance for all diagrams in the ttk-data example `persistenceDiagramClustering`)

For this, I have taken the global distance/cost result that is provided in the terminal. I don't have a way to compute it per diagram for more than two diagrams (this is what `PersistenceDiagramDistanceMatrix` is for) and I have to attach it to individual vtkBlocks to be able to see them in the ParaView interface: this is why they are duplicated.

Issue #626, opened 2021-06-22
FTMTree weird segmentation


dsakurai:
**Describe the bug**

The segmentation generated from FTM contour tree is wrong (as far as I understand).

<img width="2091" alt="debug-contourtree-latest-commit" src="https://user-images.githubusercontent.com/16434308/122849277-29f49b80-d346-11eb-9c17-808de069c019.png">

In the image above, the segmentation is color-coded on the point gaussian of the mesh.
(Edit: the numbers on the vertices are the vertex IDs.)
The gray cylinders are the skeleton arcs of the contour tree.

Point 9 shares the segmentation with point 10.
This is a bug (if my understanding is correct) because the yellow contour containing point 4 should separate point 9 from 10.

I dug in the code to find out that the FTM tree somehow adds point 9 in the superarc between points 4 and 10.
I'm not perfectly sure about this observation because I don't fully understand the code. Please have a look if needed.

In addition, point 11 should also not belong to the same segmentation as point 10.

**To Reproduce**

Run the ParaView state file below:
[debug-contourtree.pvsm.zip](https://github.com/topology-tool-kit/ttk/files/6690438/debug-contourtree.pvsm.zip)
I used ParaView 5.9.0 with the latest dev branch of the TTK (commit 3fbb9e98fc2d9f4e80a8b79e09262b1cc74c3345 ).

**Expected behavior**
Point 9 and 11 should belong to a segmentation different from point 10.

**System:**
 - OS: macOS 10.15.7 Catalina
 - Compiler: clang 10
 - Installed TTK's optional dependencies:
~~~
 TTK_ENABLE_EIGEN                 ON                                                                                                                                                                    
 TTK_ENABLE_GRAPHVIZ              ON                                                                                                                                                                    
 TTK_ENABLE_OPENMP                ON                                                                                                                                                                    
 TTK_ENABLE_SCIKIT_LEARN          ON                                                                                                                                                                    
 TTK_ENABLE_SQLITE3               ON                                                                                                                                                                    
 TTK_ENABLE_ZLIB                  ON                                                                                                                                                                    

 TTK_ENABLE_EMBREE                OFF                                                                                                                                                                   
 TTK_ENABLE_MPI                   OFF                                                                                                                                                                   
 TTK_ENABLE_SPECTRA               OFF                                                                                                                                                                   
 TTK_ENABLE_WEBSOCKETPP           OFF                                                                                                                                                                   
[...]

PR #634, opened 2021-07-13
ftmTree: Put back loose comparisons

pierre-guillou:
Following @CharlesGueunet's comment in https://github.com/topology-tool-kit/ttk/commit/f6f21b0206cdbac6ba802d5afc5da31debae43ca#r53373160, the strict order comparisons where reverted back to loose comparisons.

@CharlesGueunet is this OK for you?

No change observed in the ttk-data states.

Enjoy,
Pierre


CharlesGueunet:
seems right, did you try on the dataset mentioned in #626 ?

pierre-guillou:
I get the following result, which seems to solve #626.

Capture d’écran du 2021-07-13 16-27-57

CharlesGueunet (approved):

julien-tierny:
awesome, thanks for the quick fix.

PR #635, opened 2021-07-13
ftmTree: Fix Archer2 HPC cluster GCC6 build error

pierre-guillou:
@julesvidal has been trying to build TTK on [Archer2](https://docs.archer2.ac.uk/index.html) an HPC cluster at Edinburgh. Sadly, the default compiler (GCC6) returns a build error on an OpenMP atomic section of the ftmTree module. 

This simple fix has been proposed by people at University of Edinburgh. I tested it on ttk-data and could not observe any difference.

@CharlesGueunet While we're at it, do you think this change can have unforeseen consequences with relation to the semantics of the atomic section?

Pierre

### DIFF
--- a/core/base/ftmTree/FTMTree_MT_Template.h
+++ b/core/base/ftmTree/FTMTree_MT_Template.h
@@ -382,12 +382,13 @@ namespace ttk {
 
       // is last
       valence oldVal;
+      valence &tmp = (*mt_data_.valences)[currentState.vertex];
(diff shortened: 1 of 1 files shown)


CharlesGueunet (approved):

CharlesGueunet:
By using a reference, this version should have the same result as the previous one

julien-tierny:
cool, thanks!

PR #636, opened 2021-07-15
PersistenceDiagramClustering: Fix compatibility with PersistenceDiagramDistanceMatrix

pierre-guillou:
This PR fixes the compatibility between the "Clustered Diagrams" output of the `PersistenceDiagramClustering` filter and the `PersistenceDiagramDistanceMatrix` filter after #632.

First, error handling in `PersistenceDiagramDistanceMatrix` has been overhauled to avoid segfaults when the input diagrams are malformed. Instead, error messages are printed.

To ensure that the correct distances are computed, the original diagrams that are fed to `PersistenceDiagramClustering`, are shallow copied to the "Clustered Diagrams" output. Reconstructing the diagrams from the intermediate representation (vector of tuples) is too error prone (especially around the global min-max pair).

No additional changes observed on the two concerned ttk-data states.

Enjoy and sorry for the trouble,
Pierre


julien-tierny:
alright, thanks!
regarding the distances to attach to the "Matching" outputs, these should already be available in principle from the base code. to discuss with @julesvidal

PR #640, opened 2021-07-28
TopologicalCompression: Use a relative tolerance for ZFP

pierre-guillou:
This PR modifies the way the ZFP Tolerance parameter is used in the TopologicalCompression and TopologicalCompressionWriter filters.

Since #603, ZFP is used with an absolute tolerance rate, set by default to the value 50. While this is fine for images with scalar field ranging from 0 to 255, compressing datasets with smaller scalar field values need updated tolerance values.

With this PR, the default tolerance value is now relative to the range of the scalar field (in percent). This way, the default tolerance value of 50(%) is sufficient to get good compression results.

Since only the compressors have been modified and since the absolute tolerance value is stored in the headers of the compressed files, the ttk-data state `persistenceDrivenCompression` is not modified by this change (apart from the displayed tolerance value, which is now not 100% correct anymore).

Enjoy,
Pierre

### DIFF
--- a/CMake/topological_compression.xml
+++ b/CMake/topological_compression.xml
@@ -72,14 +72,15 @@
 
 <DoubleVectorProperty
     name="ZFPTolerance"
-    label="ZFP Absolute Error Tolerance (extra)"
+    label="ZFP Relative Error Tolerance (extra)"
     command="SetZFPTolerance"
     number_of_elements="1"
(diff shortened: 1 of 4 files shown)


julien-tierny:
awesome, thanks!
(the state file can be easily adjusted)

PR #641, opened 2021-07-28
DiscreteGradient: Fix vtkCells offset field for gradient glyphs

pierre-guillou:
An error in the DiscreteGradient filter is causing segfaults when trying to threshold the GradientGlyphs output. This PR fixes this error by providing the correct value at the end of the offset array used to define the `vtkCells`.

No other misuse of the offset array for building `vtkCells` spotted so far.

Enjoy,
Pierre


julien-tierny:
thanks!

PR #637, opened 2021-07-20
Fix unused warning in ttkOFFReader

mwestphal:
Thanks for contributing to TTK!

Before submitting your pull request, please:

- [X] Review our [Contributor Guidelines](https://github.com/topology-tool-kit/ttk/blob/dev/CONTRIBUTING.md), in particular regarding code formatting (with clang-format) and continuous integration.

- [X] Please provide a quick description of your contributions below:

Fix unsused variable warnings in ttkOFFReader

PR #642, opened 2021-07-30
Enable -Wextra compilation flag, fix warnings

pierre-guillou:
As suggested in #637, this PR enables the -Wextra flag in CMake for GCC and Clang. Since -Wtype-limits is included in GCC's -Wextra but not in Clang, this flag is also added.

Note that this PR does disable the -Wunused-parameter warning flag, included by -Wextra, which was causing the warning in #637. A second PR will focus on fixing and enabling -Wunused-parameter.

All warnings caused by -Wextra -Wno-unused-parameters (-Wtype-limits) have been fixed by this PR. Since most of them have been already fixed in previous PRs, warning fixes are mostly located in newest modules.

This PR also contains some very small fixes for clang-tidy (self-assignment warnings) and the clang static analyzer (silencing warnings on Eigen and CPython). A third PR will fix all remaining clang-tidy/clang static analyzer warnings.

No change so far in the validation states screenshots.

Enjoy,
Pierre

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -5,6 +5,7 @@ Checks: >-
   -bugprone-too-small-loop-variable,
   -bugprone-macro-parentheses,
   -bugprone-reserved-identifier,
+  cert-oop54-cpp,
   misc-*,
   -misc-non-private-member-variables-in-classes,
   -misc-unused-parameters,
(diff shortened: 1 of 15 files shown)


julien-tierny:
alright, thanks!

PR #643, opened 2021-08-02
Enable and fix -Wunused-parameters

pierre-guillou:
Following #642 and #637, this PR enables the `-Wunused-parameters` compiler flag under the `-Wextra` umbrella and fixes all subsequent warnings.

The `-Wunused-parameters` warning flag highlights a function parameter that is not currently used in the function body. To fix it, 3 actions are possible:
* it the most simple cases, the offending parameter can be removed from the function definition & declaration signatures and in every calling sites.
* when the current function has to upheld a specific API that constrain its parameter (for instance, the `request` parameter of the `RequestData` member functionr in the VTK layer), following VTK's approach, I introduced and used a first new macro `ttkNotUsed()` defined in the base layer. This simple macro removes the name of the parameter in the function definition signature, kind of like commenting it (but still keeping the fact that it's unused).
* when the offending parameter is used in a specific configuration in the function, meaning under some `#ifdef` (for instance when passing thread number as parameters with `TTK_ENABLE_OPENMP` is `OFF`), I trick the compiler by inserting a no-op statement involving the parameter in the `#else` branch or anywhere else in the function body. To do so, I introduced a second macro, `TTK_FORCE_USE()` which basically casts the parameter to `void`, resulting in a no-op.

As a bonus, I fixed some `-Wextra-semi` warnings (not enabled by default) that flags useless semicolons (usually after function definitions) and replaced some empty user-defined constructors or destructors with `default`.

No change observed in the ttk-data states.

Enjoy,
Pierre


julien-tierny:
ok (haven't checked completely though)

PR #645, opened 2021-08-26
BarycentricSubdivision: Fix #644

### LINKED ISSUE #644: Barycentric subdivision should quit properly in dimension 3
In the current state, applying a TTK Baricentric subdivision on a 3D data set will emit a warning telling dim 3 is not supported ... then seg fault. We should try to avoid this second part :)

Best,
Charles

pierre-guillou:
This PR reworks the BarycentricSubdivision filter to properly exit with an error message when applied on 3D datasets (instead of segfaulting).

Enjoy,
Pierre


julien-tierny:
ok, thanks!

PR #646, opened 2021-08-27
Timer: Replace platform-specific code with std::chrono

pierre-guillou:
This PR replaces the platform-specific code (POSIX `gettimeofday` for Linux and MacOS, `QueryPerformanceFrequency` and `QueryPerformanceCounter` for Windows) with the `std::chrono` standard library calls (introduced in C++11).

Alternative TDA libraries, such as Gudhi and Oineus, already use `std::chrono` functions for computing timings.

The only drawback that I see of this new implementation is that it exposes `std::chrono` in the Timer struct fields instead of hiding them. I think this is worth the removal of the platform-specific code.

Struct size is identical (8 bytes, size of double). No change observed in the timings so far.

The Timer code has been moved in a separate header in the `base/common` directory.

Enjoy,
Pierre

### DIFF
--- a/core/base/common/CMakeLists.txt
+++ b/core/base/common/CMakeLists.txt
@@ -13,5 +13,6 @@ ttk_add_base_library(common
         OrderDisambiguation.h
         Os.h
         ProgramBase.h
+        Timer.h
         Wrapper.h
         )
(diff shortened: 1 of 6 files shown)


julien-tierny:
ok

PR #647, opened 2021-08-30
New MergeBlockTables filter

pierre-guillou:
In #632, PersistenceDiagramClustering output has been moved into `vtkMultiBlockDataSets`, one for every diagram. Clustering information on diagrams, passed as Field Data on the "Clustered Diagrams" output, has also been split into blocks. Now, applying DatasetToTable on these Field Data produces a vtkMultiBlockDataSet of vtkTables instead of one unique vtkTable. Filters that take a vtkTable (CinemaQuery...) are then unable to operate on this multi-block structure. 

ParaView provides a filter called "Merge Blocks" to merge datasets in blocks into one dataset but it only acts on Unstructured Grids or Polygonal Meshes.

Since we rely on this clustering information in the VESTEC project, this PR introduces a new filter, MergeBlockTables, that aims to bridge the gap between the new PersistenceDiagramClustering output format and vtkTables filters. This new filter takes a vtkMultiBlockDataSet of vtkTables as an input and merges then into one vtkTable, similarly to ParaView's "Merge Blocks".

Enjoy,
Pierre

### DIFF
--- a/core/vtk/ttkMergeBlockTables/CMakeLists.txt
+++ b/core/vtk/ttkMergeBlockTables/CMakeLists.txt
@@ -1,0 +1,1 @@
+ttk_add_vtk_module()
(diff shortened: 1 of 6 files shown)


julien-tierny:
alright, looks good. thanks!

PR #648, opened 2021-08-30
ExplicitTriangulation: Fix segfault calling buildEdgeList with wrong template argument

pierre-guillou:
The  `OneSkeleton::buildEdgeList` method has been templated in #593 with the dataset dimension as template parameter to unify implementations for 2D and 3D datasets. 

However, using the wrong template call can lead to segfault: if a filter calls `preconditionVertexNeighbors` followed by `preconditionEdgeStars` on a 3D tetrahedron dataset for instance, `preconditionVertexNeighbors` will initialize the `edgeList` in the skeleton module, skipping the call to `preconditionEdges` in the AbstractTriangulation. The latter call to `preconditionEdgeStars` will segfault due to the wrong default parameter when calling `OneSkeleton::buildEdgeList` in `ExplicitTriangulation::preconditionEdgeStarsInternal`.

While getting to this segfault may seem hard, just applying ScalarFieldSmoother then DiscreteGradient on any 3D tetrahedron dataset is enough to trigger it.

To fix this issue, this PR reworks the handling of the `OneSkeleton::buildEdgeList` calls in the skeleton and ExplicitTriangulation modules:
* the `buildEdgeList` method has been refactored not to accept optional parameters. This may increase performance (no more null pointer checks in the main loop) at the expense of memory: calling buildEdgeList will now compute and keep in memory the edgeList, the edgeStarsList and the cellEdgesList even though the latter are not used,
* intra-skeleton module calls to `buildEdgeList` were replaced by calls to `preconditionEdgesInternal` in ExplicitTriangulation and passing the edgeList from there,
* `buildEdgeList` calls in the ExplicitTriangulation module were replaced with `preconditionEdgesInternal` calls, which dispatch calls to `buildEdgeList` according to the dataset dimension.

In the end, the calls to `buildEdgeList` have been reduced to `ExplicitTriangulation::preconditionEdgesInternal`, pointers to edgeList, edgeStarsList and cellEdgesList have been replaced with references (and calls to `preconditionEdgesInternal`).

And since TTK filters are sometimes used on 1D datasets (in ttk-data/morseMolecule.psvm, GeometrySmoother is applied on 1D separatrices), this PR also enables the support of 1D datasets in the ExplicitTriangulation (at least the bits necessary for the DiscreteGradient and the ttk-data states to run normally).

No changes detected in the ttk-data states.

Enjoy,
Pierre

### DIFF
--- a/core/base/explicitTriangulation/ExplicitTriangulation.cpp
+++ b/core/base/explicitTriangulation/ExplicitTriangulation.cpp
@@ -289,11 +289,11 @@ int ExplicitTriangulation::preconditionEdgesInternal() {
     oneSkeleton.setWrapper(this);
     // also computes edgeStar and triangleEdge / tetraEdge lists for free...
     if(getDimensionality() == 2) {
-      return oneSkeleton.buildEdgeList(vertexNumber_, *cellArray_, &edgeList_,
-                                       &edgeStarData_, &triangleEdgeList_);
+      return oneSkeleton.buildEdgeList(vertexNumber_, *cellArray_, edgeList_,
+                                       edgeStarData_, triangleEdgeList_);
     } else if(getDimensionality() == 3) {
-      return oneSkeleton.buildEdgeList(vertexNumber_, *cellArray_, &edgeList_,
-                                       &edgeStarData_, &tetraEdgeList_);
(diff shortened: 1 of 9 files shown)


julien-tierny:
ok, thanks!

PR #650, opened 2021-09-02
GitHub Actions: New workflow to build & test TTK (+ run ttk-data states on Ubuntu)

pierre-guillou:
This PR introduces a new GitHub Actions workflow that builds TTK on Ubuntu (18.04 & 20.04), macOS & Windows and tests the TTK examples. On Ubuntu, the ttk-data validation state files are also run and compared to the reference.
  
While similar to the Azure Pipelines CI, this new workflow has two main advantages:
  * for all 3 platforms, it will build the whole ParaView (minus the Qt UI) and fetch all TTK optional dependencies;
  * the workflow can be triggered by GitHub users / TTK developers on their public forks with tags matching the regexps 'ci*', 'build*' and 'test*' (while being automatically triggered by PRs).
  
Non-packaged optional dependencies (Spectra, ZFP, Websocket++) build & install steps have been factored into a "composite" action (Ubuntu, macOS) that can be shared between this workflow and the packaging workflow. Same thing for building & testing TTK examples (also for Ubuntu & macOS).
  
On Ubuntu, ttk-data is also fetched and the state files are run using the validation scripts introduced in https://github.com/topology-tool-kit/ttk-data/pull/69. For now, if the validation detects errors, the faulty screenshots will be uploaded as a CI product (this might use GitHub storage) but the CI won't fail (except an error annotation in the job summary page). This in order to still cache the generated ParaView package in case of failure (and also due to non up-to-date reference screenshots for bivariateToyCspPeeling, PersistenceDiagramClustering & PersistenceDiagramDistance).
  
As a remainder, the ttk-data validation runs only on Ubuntu because the generated screenshots depend on:
  * OSMesa-built ParaView (no hardware rendering on GitHub Actions CI),
  * the compiler/standard library version (GCC11) for identical pseudo-random outputs (IdentifierRandomizer)
  * and the pip version of `scikit-learn`for DimensionReduction.
  
The packaging workflow has also been renamed (main.yml -> package.yml). A  unique tag regex have been added to trigger every workflows separately ('pack*' for the packaging workflow, 'check*' for the check_code workflow, either 'test*' or 'build*' for the new test_build workflow).
  
The workflow should take around 30/40min to complete once the ParaView build is cached, otherwise around 3h.

A small modification in the websocketio module (a Boost define) is necessary to fix link errors on Ubuntu.
  
  Enjoy,
  Pierre


julien-tierny:
ok, looks good!

PR #652, opened 2021-09-13
UncertainDataEstimator: Fix segfaults

pierre-guillou:
This PR fixes segfaults that were lingering in the UncertainDataEstimator module since the Great Migration.

Enjoy,
Pierre


julien-tierny:
ok, thanks!

PR #653, opened 2021-09-13
[planarGraphLayout] fix merge tree layout matching cost

MatPont:
This PR removes the comment indicator of the line adding the cost to each cell in the matching output with merge trees.

Otherwise, the cost field will not appears..


julien-tierny:
alright, looks good. 
thanks!

PR #655, opened 2021-09-17
Add new FlattenMultiBlock module

pierre-guillou:
This PR adds a new TTK filter that flattens one level of a vtkMultiBlockDataSet hierarchy.

It can be used on the (Group Dataset of) Clustered Diagrams and Cluster Centroids output of the PersistenceDiagramClustering module to bring input diagrams and centroids into the same MultiBlockDataSet.

The old parent block is identified as a Field Data array, named "ParentBlockId", attached to every block.

Enjoy,
Pierre


julien-tierny:
ok, thanks!

PR #656, opened 2021-09-23
Merge xml common parameters of merge tree filters

MatPont:
The filter MergeTreeClustering and MergeTreeDistanceMatrix have parameters in common in the filter GUI of paraview. For the moment some code is duplicated in their xml files. 

This PR, inspired dy debug_widgets.xml, merge these parameters in common into "shared" xml files such that there is no duplicated code anymore.


julien-tierny:
ok, looks good. thanks!

PR #657, opened 2021-09-24
[mergeTreeClustering] fix join/split mixture coefficient

MatPont:
The GUI parameter allowing to controls the weight of join or split tree in the clustering algorithm of the MergeTreeClustering filter is not correctly given to the base code. 

Hence, at that moment, changing this parameter in the GUI does not change anything. This PR fixes this problem.


julien-tierny:
nice.

PR #658, opened 2021-09-24
PersistenceDiagramClustering: Fix log, compatibility with FlattenMultiBlock

pierre-guillou:
This PR reduces the scope of a warning message in the base layer of the PersistenceDiagramClustering module.

Furthermore, the output diagrams and centroids format has been reworked for both to be compatible with FlattenMultiBlock (no more segfault when applying PersistenceDiagramDistanceMatrix after FlattenMultiBlock). The "DiagramID" point data array has been removed from the Clustered Diagrams output (diagrams are now in individual blocks and can be filtered with Extract Block/TTKExtract) and the "ClusterId" field data has been added to the Cluster Centroids outputs.

No change in the two ttk-data states using PersistenceDiagramClustering.

Enjoy,
Pierre


julien-tierny:
ok, thanks!

PR #659, opened 2021-09-24
[planarLayout] merge tree min and max number of important pairs

MatPont:
The most important parameter of the merge tree planar layout is the important pairs threshold that allows to consider as "important" the pairs having a persistence above this threshold (they will be much more visible).

However, when many trees are present, this threshold (in percentage) can be complicated to use to have a good planar layout for all trees (the same threshold can give a very high number of important pairs for a tree but very few for another one).

To tackle this problem we can add two parameters (in advanced) than bound the minimum and maximum number of possible important pairs for all trees. Hence ensuring that the planar layout of all trees are not too much imbalanced.

(in a future PR, the (planar layout) parameters in common in the merge tree filters will be merged)


julien-tierny:
ok!

PR #660, opened 2021-09-24
[mergeTreeClustering] fix compatibility with FlattenMultiBlock

MatPont:
As in #658 the MergeTreeClustering filter followed by a FlattenMultiBlock and a MergeTreeDistanceMatrix can cause a segmentation fault.

This is because the clustered trees will have a cluster ID field data and not the centroids tree (hence making an error while copying this field data in the distance matrix).

As in #658, this PR fixes the problem by adding the cluster ID to the centroids.


julien-tierny:
thanks!

PR #661, opened 2021-09-27
GitHubActions: Replace TTK-ParaView cache with release assets

pierre-guillou:
This PR replaces the cache used in the "test_build" GitHub Actions workflow with release assets.

The last PRs on the TTK repositories showed that the GitHub Actions cache scope is restricted to a branch and PRs that target it. Since the CI is not triggered after the merge but only when PRs are opened, this means that currently no cache is associated with the "dev" branch.

Another workaround would be to enable this workflow on push events on the "dev" branch. However, this might have an impact on forks of the reference repository with "dev" branches, triggering sometimes unwanted CI pipelines. Besides, in case of a cache miss during a PR, a second cache miss would also occur after the merge because the PR-triggered workflow cache is associated to the PR branch, not the "dev" branch. The current version, by using Release Assets instead of cache, solves those two shortcomings.

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,21 +31,6 @@ jobs:
     - uses: actions/checkout@v2
       name: Checkout TTK source code
 
-    - name: Cache ttk-paraview package
-      uses: actions/cache@v2
-      id: cache-ttk-paraview
-      with:
-        path: "ttk-paraview/build/ttk-paraview.deb"
-        key: ttk-paraview-${{ matrix.os }}
-
-    - uses: actions/checkout@v2
-      if: steps.cache-ttk-paraview.outputs.cache-hit != 'true'
-      with:
-        repository: "topology-tool-kit/ttk-paraview"
-        ref: "5.9.0"
-        path: "ttk-paraview"
-      name: Checkout TTK-ParaView source code
-
     - name: Install Ubuntu dependencies
       run: |
         sudo apt update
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -194,52 +158,13 @@ jobs:
...
-        cmake --build . --parallel
-        sudo cmake --build . --target install
-
-    - name: Create ParaView package
-      if: steps.cache-ttk-paraview.outputs.cache-hit != 'true'
+    - name: Fetch TTK-ParaView headless macOS binary archive
       run: |
-        cd ttk-paraview/build
-        sudo cpack -G TGZ
+        wget https://github.com/pierre-guillou/paraview-ttk/releases/download/5.9.1-headless/ttk-paraview-headless.tar.gz
```
julien-tierny: this asset should be available on `https://github.com/topology-tool-kit/ttk-paraview` instead.
thanks PRing this repository accordingly.

.github/workflows/test.yml
```diff
@@ -310,41 +220,16 @@ jobs:
...
-    - name: Create ParaView package
-      if: steps.cache-ttk-paraview.outputs.cache-hit != 'true'
-      shell: bash
+    - name: Fetch TTK-ParaView headless Windows installer
       run: |
-        cd ttk-paraview/build
-        cpack -G NSIS64
+        Invoke-WebRequest `
+        -OutFile ttk-paraview-headless.exe `
+        -Uri https://github.com/pierre-guillou/paraview-ttk/releases/download/5.9.1-headless/ttk-paraview-headless.exe
```
julien-tierny: same remark as above.

julien-tierny (changes requested):

julien-tierny:
this looks like low priority.

pierre-guillou:
Also replaced the ccache cache on Ubuntu & macOS with release assets to further cut down the CI time to about 30/40min (if the manually-generated ccache archives are up-to-date).

Next: investigate sccache for MSVC (for a future PR).

.github/workflows/test.yml
pierre-guillou: Fixed through #669

.github/workflows/test.yml
pierre-guillou: idem

pierre-guillou:
The "check_code" workflow now also uses the ParaView 5.9.1 headless packages.

julien-tierny (approved):

julien-tierny:
ok, looks good now.
thanks

PR #662, opened 2021-09-28
Using CMake Targets for Eigen includes

HollowSun0:
Uses the CMake target Eigen3::Eigen to link to the Eigen library.

Requires CMake 3.0 (or later).


julien-tierny:
looks good, thanks!

PR #663, opened 2021-09-30
Homogenize merge tree temporal reduction

MatPont:
This PR merges the preprocessing parameters shared with the other merge tree filters (using a shared xml file).

The input options of the encoding are now saved as field data in the output and are retrieved in the decoding.

### DIFF
--- a/core/vtk/ttkMergeTreeTemporalReductionDecoding/ttkMergeTreeTemporalReductionDecoding.cpp
+++ b/core/vtk/ttkMergeTreeTemporalReductionDecoding/ttkMergeTreeTemporalReductionDecoding.cpp
@@ -118,7 +118,8 @@ int ttkMergeTreeTemporalReductionDecoding::RequestData(
         ->GetPointData()
         ->GetArray("Scalar")
         ->GetDataType();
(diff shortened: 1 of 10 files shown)


core/vtk/ttkMergeTreeTemporalReductionDecoding/ttkMergeTreeTemporalReductionDecoding.cpp
```diff
@@ -118,6 +118,10 @@ int ttkMergeTreeTemporalReductionDecoding::RequestData(
         ->GetPointData()
         ->GetArray("Scalar")
         ->GetDataType();
+  auto assignmentSolverArray
+    = blocks->GetFieldData()->GetArray("AssignmentSolver");
+  if(assignmentSolverArray)
```
julien-tierny, line 123: else?
it looks like `assignmentSolverID_` is not initialized otherwise, is it?

julien-tierny (changes requested):

julien-tierny (approved):

julien-tierny:
ok, looks good. let's go!

PR #664, opened 2021-09-30
Merge tree planar layout merge common parameters

MatPont:
The filters MergeTreeClustering, MergeTreeTemporalReductionXX and PlanarGraphLayout have parameters in common in the filter GUI of paraview regarding the merge tree planar layout. For the moment some code is duplicated in their xml files.

This PR, inspired dy debug_widgets.xml, merge these parameters in common into "shared" xml files such that there is no duplicated code anymore.


julien-tierny:
looks good, thanks!

PR #665, opened 2021-10-01
nullptr check in GetVoidPointer; new util function SliceArray

JonasLukasczyk:
Hi Julien,
this PR adds a sanity check in `ttkUtils::GetVoidPointer` to prevent a segfault when a `nullptr` is passed to  the function. This is convenient, because now one can write:
```
auto arrayData = ttkUtils::GetPointer<double>(somePointData->GetArray("Test"));
if(!arrayData)
  return ERROR;
```
Before one had to first check if the array exists, and then feed it into the `GetVoidPointer` method. Otherwise a seg fault would occur.

I also added a new Utils function called `SliceArray` which takes a `srcArray` and a `tupleIdx` as input. The function then creates a new `dstArray` with the same name and number of components, but the new array only contains the tuple at the specified index. This is convenient for many Cinema related stuff.

Best
Jonas

### DIFF
--- a/core/vtk/ttkAlgorithm/ttkUtils.cpp
+++ b/core/vtk/ttkAlgorithm/ttkUtils.cpp
@@ -234,9 +234,10 @@ void *ttkUtils::GetVoidPointer(vtkPoints *points, vtkIdType start) {
   return GetVoidPointer(points->GetData(), start);
 }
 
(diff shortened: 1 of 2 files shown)


julien-tierny:
alright, looks good! 
thanks :)

PR #666, opened 2021-10-07
Use TTK version for the ParaView plugin

CharlesGueunet:
Dear Julien,

This is a really small PR that will make the ParaView plugin use the same version as the TTK package, so users won't think we have hit a 1.0 without celebration :)

Best,
Charles


julien-tierny:
looks good, thanks.

PR #667, opened 2021-10-07
mergeTreeTemporalReduction remove encoding visu parameters

MatPont:
This PR removes the visualization gui parameters of the mergeTreeTemporalReductionEncoding filter.


julien-tierny:
looks good!

PR #668, opened 2021-10-07
DiscreteGradient: Fix simplexes comparisons in priority queues

pierre-guillou:
This PR fixes the simplexes comparison order when filling the priority queues of the processLowerStar algorithm in the DiscreteGradient. Currently, comparing simplexes of dimension > 1 in a lower star do not take the filtration order into account. This should be fixed in this PR with the following changes:
* for each simplex in a lower star, instead of storing the lower vertices, we store their order value sorted in decreasing order (the higher order value first). -1 values are used to fill missing vertex order values for edges and triangles. Comparing two simplexes becomes a simple lexicographic comparison between two `std::array`s.

This change causes, however, some modifications in the ttk-data states. 1manifoldLearning, 2manifoldLearning, morseMolecule, morseSmaleQuadrangulation, tectonicPuzzle and tribute (all states that call the Morse-Smale complex and display 2-saddles, maxima or ascending 1-separatrices) are slightly affected.

Also included in this PR:
* an extended use of the `firstprivate` clause in the MorseSmaleComplex (3D). It improves performance (+8/10% when computing saddle connectors + 2-separatrices) compared to a shared vector between threads in which each thread access its own element,
* DimensionReduction now exits with an helpful error message when the input matrix has 0 rows or columns,
* DistanceField does not segfault anymore when the input dataset is not a valid triangulation (preconditionTriangulation now called after checking triangulation != nullptr).

Those last changes do not affect the ttk-data states.

Enjoy,
Pierre

### DIFF
--- a/core/vtk/ttkDistanceField/ttkDistanceField.cpp
+++ b/core/vtk/ttkDistanceField/ttkDistanceField.cpp
@@ -47,70 +47,44 @@ int ttkDistanceField::RequestData(vtkInformation *ttkNotUsed(request),
   vtkPointSet *sources = vtkPointSet::GetData(inputVector[1]);
   vtkDataSet *output = vtkDataSet::GetData(outputVector);
 
-  ttk::Triangulation *triangulation = ttkAlgorithm::GetTriangulation(domain);
-  this->preconditionTriangulation(triangulation);
-  Modified();
-
-#ifndef TTK_ENABLE_KAMIKAZE
-  if(!triangulation) {
-    this->printErr("wrong triangulation.");
-    return -1;
+  auto triangulation = ttkAlgorithm::GetTriangulation(domain);
+  if(triangulation == nullptr) {
+    this->printErr("Wrong triangulation.");
+    return 0;
   }
-#endif
+
+  this->preconditionTriangulation(triangulation);
 
   std::vector<ttk::SimplexId> idSpareStorage{};
   auto *identifiers = this->GetIdentifierArrayPtr(ForceInputVertexScalarField,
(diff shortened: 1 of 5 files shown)


core/vtk/ttkDistanceField/ttkDistanceField.cpp
```diff
@@ -47,70 +47,44 @@ int ttkDistanceField::RequestData(vtkInformation *ttkNotUsed(request),
...
-    this->printErr("wrong triangulation.");
-    return -1;
+  auto triangulation = ttkAlgorithm::GetTriangulation(domain);
+  if(triangulation == nullptr) {
+    this->printErr("Wrong triangulation.");
+    return 0;
   }
-#endif
+
+  this->preconditionTriangulation(triangulation);
```
julien-tierny, line 56: triangulation preconditioning should only occur at the base level.
thanks for moving that there.

julien-tierny (changes requested):

core/vtk/ttkDistanceField/ttkDistanceField.cpp
pierre-guillou: The corresponding triangulation preconditioning code is located in the base layer: https://github.com/topology-tool-kit/ttk/blob/bca49aaa62e725f5866fb409165b6980b3548c07/core/base/distanceField/DistanceField.h#L51

julien-tierny: my bad.

julien-tierny (approved):

julien-tierny:
thanks.

PR #669, opened 2021-10-11
GitHub Actions: Fix packaging for ParaView >= 5.9.0, generate ccache archives for CI

pierre-guillou:
This PR updates the packaging workflow to be compatible with ParaView >= 5.9.0. It complements https://github.com/topology-tool-kit/ttk-paraview/pull/3 on the TTK side while also importing the modified ttk-paraview GitHub Actions workflows inside paraview/patch.

The packaging workflow has been modified as follows:
* only Ubuntu 20.04 is supported (since ParaView 5.9 is not compatible anymore with Ubuntu 18.04)
* the embree in Ubuntu 20.04 repositories (3.3) is too old to work with TTK (which needs >= 3.4)
* a new dependency to `libgl1-mesa-dev` on Ubuntu, needed by some ParaView headers
* the macOS job uses brew's Python and Qt 5 is explicitly selected
* the Windows job uses conda's Python and every dependency is pulled from the conda-forge repository (including graphviz but except embree)
To generate packages, make sure that a v5.9.1 release exists in the topology-tool-kit/ttk-paraview repository, then push a tag matching the regular expressions `v*`, `*-*-*` or `*.*.*`.

A new workflow to generate ccache archives to speed up the ttk-data CI is also included. This workflow bypasses the GitHub Action cache by using Release Assets as a permanent object storage. Only Ubuntu and macOS jobs will benefit from it since object caching is not working (yet?) on Windows. This should further reduce the CI from about 1h to 40min (the bottleneck being the Windows compilation of TTK) when used with the headless packages generated by https://github.com/topology-tool-kit/ttk-paraview/pull/3 (see https://github.com/topology-tool-kit/ttk/pull/661 for an example usage). The ccache archive is generated by pushing a tag named "ccache". It should be manually refreshed every time the Ubuntu job takes more time than the Windows one. Once generated on this repository, I will modify https://github.com/topology-tool-kit/ttk/pull/661 to use it.

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/ccache.yml
+++ b/.github/workflows/ccache.yml
@@ -1,0 +1,211 @@
+name: store_ccache
+
+on:
+  push:
+    branch-ignore:
+      - '*'
+    tags:
+      - 'ccache'
+
+env:
+  PV_TAG: v5.9.1-headless
+  PV_REPO: topology-tool-kit/ttk-paraview
+
+jobs:
+
+  # -------#
+  # Ubuntu #
+  # -------#
+  ccache-ubuntu:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, ubuntu-20.04]
+    steps:
+
+    - uses: actions/checkout@v2
+      name: Checkout TTK source code
+
+    - name: Install Ubuntu dependencies
+      run: |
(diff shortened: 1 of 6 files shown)


julien-tierny:
thanks!

PR #672, opened 2021-10-21
Periodic boundary bug fix #671

### LINKED ISSUE #671: Periodic conditions are not properly handled by the Morse-Smale complex in 3D
**Describe the bug**
When periodic conditions are enabled in 3D, an excessive number of critical points is identified by the Morse-Smale complex, as reported by an email on the ttk-users@googlegroups.com mailing list on October 20, 2021.

**To Reproduce**
1. Load `BuiltInExample2.vti` from `ttk-data`
2. Apply the PeriodicGrid filter
3. Apply the Morse-Smale complex filter (on the field `log(Rho)`)
4. Too many critical points are reported on the boundary.

**Expected behavior**
Fewer critical points are expected, typically, nearly none on the boundary

**Screenshots**
issueBoundaryPeriodic

 - OS: [e.g. iOS] gentoo linux 
 - Compiler [e.g. 22] gcc
 - Installed TTK's optional dependencies 

**Additional context**
Note that:
- PersistenceDiagram produces the right behavior when periodic conditions are enabled
- ManifoldCheck produces the right behavior when periodic conditions are enabled
  - then it is unlikely that the issue is at the `periodicImplicitTriangulation` level (ManifoldCheck spans a good set of traversal features)
  - plus, the periodic conditions were initially developed with a use case dealing with the Morse-Smale complex in 3D
  - then, I suspect this is a regression introduced with the homotopicExpansion type discrete gradient (which was introduced after periodic conditions, if I'm not mistaken)
- The issue does not appear in 2D, so this may be an issue with the edge-triangle traversal in the homotopicExpansion.

tbmasood:
This fixes a bug in the periodic implicit triangulation which was causing faulty Morse-Smale complex computation reported recently in #671.


julien-tierny:
awesome @tbmasood 
thanks for the immediate fix!!!

PR #673, opened 2021-10-25
Make IdentifierRandomizer platform-independant

pierre-guillou:
This PR replaces the usage of `std::shuffle` inside IdentifierRandomizer with a custom implementation of a well-known algorithm, the [Fisher-Yates shuffle](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle).

This ensures that the IdentifierRandomizer output is the same on all supported platforms (Linux, macOS and Windows). As it happens, `std::shuffle` results are platform-dependant.

As a consequence, the dependency to GCC11 in the Ubuntu VM of the test workflow is lifted.

Enjoy,
Pierre

### DIFF
--- a/core/vtk/ttkIdentifierRandomizer/ttkIdentifierRandomizer.cpp
+++ b/core/vtk/ttkIdentifierRandomizer/ttkIdentifierRandomizer.cpp
@@ -1,4 +1,4 @@
-#include "BaseClass.h"
+#include <Shuffle.h>
 #include <ttkIdentifierRandomizer.h>
 
 #include <vtkCellData.h>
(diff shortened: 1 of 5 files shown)


core/vtk/ttkIdentifierRandomizer/ttkIdentifierRandomizer.cpp
```diff
@@ -42,6 +42,14 @@ int ttkIdentifierRandomizer::FillOutputPortInformation(int port,
   return 0;
 }
 
+template <typename T, typename U>
```
julien-tierny: I guess it would make sense to move that function further down the base code.
for example, in OS specifics `core/base/common/Os.h`

pierre-guillou: This function has been moved into the `common` module, in its own header `Shuffle.h` (to be explicitly imported by modules needing it).

julien-tierny:
ok, thanks.

PR #674, opened 2021-10-25
Compact Triangulation Integration

guoxiliu:

This PR adds a compact triangulation data structure and also a preconditioning plugin for it.


julien-tierny:
hi @guoxiliu 

thanks a lot for your PR.

it seems that the CI identified a couple of issues with this PR. 
these need to be solved before we move on with the code review. 

could you please make sure:
- to clang-format your code properly (see https://github.com/topology-tool-kit/ttk/blob/dev/CONTRIBUTING.md)
- that your code also compiles fine with the following configuration: KAMIKAZE=FALSE, OPENMP=FALSE, 64BIT_IDS=FALSE, Release (the CI failed to build there, see the error messages there: https://github.com/topology-tool-kit/ttk/runs/4017913908?check_suite_focus=true#logs)

Thanks!

guoxiliu:
Thanks, @julien-tierny. I will continue to work on these issues. 

> hi @guoxiliu
> 
> thanks a lot for your PR.
> 
> it seems that the CI identified a couple of issues with this PR. these need to be solved before we move on with the code review.
> 
> could you please make sure:
> 
> * to clang-format your code properly (see https://github.com/topology-tool-kit/ttk/blob/dev/CONTRIBUTING.md)
> * that your code also compiles fine with the following configuration: KAMIKAZE=FALSE, OPENMP=FALSE, 64BIT_IDS=FALSE, Release (the CI failed to build there, see the error messages there: https://github.com/topology-tool-kit/ttk/runs/4017913908?check_suite_focus=true#logs)
> 
> Thanks!

core/base/compactTriangulation/CompactTriangulation.cpp
```diff
@@ -0,0 +1,30 @@
+#include <CompactTriangulation.h>
+
+using namespace std;
+using namespace ttk;
+
+CompactTriangulation::CompactTriangulation() {
+  setDebugMsgPrefix("CompactTriangulation");
+  clear();
+
+#ifdef TTK_ENABLE_OPENMP
```
julien-tierny: this distinction is not necessary.
`threadNumber_` is initialized to 1 when `TTK_ENABLE_OPENMP` is not defined.
(see https://github.com/topology-tool-kit/ttk/blob/dev/core/base/common/BaseClass.cpp)

julien-tierny: please update.

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4214 @@
...
+/// simplicial meshes.
+/// The key idea of TopoCluster is to subdivide the simplicial mesh into
+/// clusters. Then, the connectivity information is computed locally for each
+/// cluster and discarded when it is no longer needed. The simplicial mesh needs
+/// to be subdivided before using TopoCluster, i.e., there has to be a scalar
+/// field named "ttkCompactTriangulationIndex" to denote the cluster index of
+/// each vertex. Note Topocluster will reindex the simplices based on the
+/// clustering input array.
+/// \b Related \b publications \n
+/// "TopoCluster: A Localized Data Structure for Topology-based Visualization"
```
julien-tierny, line 17: don't forget to also add that reference:
- in the corresponding vtk header
- in the paraview xml file

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4214 @@
...
+      triangleLinks_ = nullptr;
+      triangleEdges_ = nullptr;
+      triangleStars_ = nullptr;
+      /* cell relationships */
+      tetraEdges_ = nullptr;
+      cellNeighbors_ = nullptr;
+      tetraTriangles_ = nullptr;
+    }
+    ~ImplicitCluster() {
+      delete internalEdgeList_;
```
julien-tierny: we strongly discourage the explicit use of `new`/`delete` (which are now usually regarded as non-recommendable programming practices), unless there's a very good reason.
in this example, it seems that `internalEdgeList_` could simply be declared as `vector<array<SimplexId, 2>>` instead of `vector<array<SimplexId, 2>> *` (pointer).
is there any motivation for using pointers here?

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4214 @@
...
+      for(SimplexId nid = 0; nid < nodeNumber_; nid++) {
+        for(SimplexId vid : nodeVertices[nid]) {
+          vertexMap[vid] = vertexCount++;
+        }
+        vertexIntervals_[nid + 1] = vertexCount - 1;
+      }
+
+      // rearange the vertex coordinate values
+      if(doublePrecision_) {
+        double *newPointSet = new double[3 * vertexNumber_];
```
julien-tierny: again, the usage of `new` and `delete` is strongly discouraged.
in this instance `vector<double> newPointSet(3*vertexNumber_);` seems to do the trick.

julien-tierny (changes requested):
I have started a review. It is not finished at the moment.
I will resume it asap.

core/base/compactTriangulation/CompactTriangulation.h
guoxiliu: Previously I thought it might be helpful to use dynamic allocation for relation arrays (like removing the unused ones to release the allocated memory), but the implementation does not work like that. I will update the code accordingly.

core/base/compactTriangulation/CompactTriangulation.cpp
julien-tierny: solved.

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved

julien-tierny (approved):

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4144 @@
...
+        // missCount_++;
+        if(caches_[threadId].size() >= cacheSize_) {
+          if(caches_[threadId].back()->nid == reservedId) {
+            return nullptr;
+          }
+          cacheMaps_[threadId].erase(caches_[threadId].back()->nid);
+          delete caches_[threadId].back();
+          caches_[threadId].pop_back();
+        }
+        caches_[threadId].push_front(new ImplicitCluster(nodeId));
```
julien-tierny: please remove the dynamic allocation (i.e, please do not use `new` and `delete`)

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4144 @@
...
+      }
+
+      boost::unordered_map<
+        SimplexId, std::vector<std::array<SimplexId, 2>>>::iterator iter;
+      for(iter = edgeNodes.begin(); iter != edgeNodes.end(); iter++) {
+        ImplicitCluster *exnode = searchCache(iter->first, nodePtr->nid);
+        bool remove = false;
+        if(!exnode) {
+          remove = true;
+          exnode = new ImplicitCluster(iter->first);
```
julien-tierny: please do not use dynamic allocation here.

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4144 @@
...
+
+      ImplicitCluster *exnode;
+      boost::unordered_map<
+        SimplexId, std::vector<std::array<SimplexId, 3>>>::iterator iter;
+      for(iter = nodeTriangles.begin(); iter != nodeTriangles.end(); iter++) {
+        exnode = searchCache(iter->first, nodePtr->nid);
+        bool remove = false;
+        if(!exnode) {
+          remove = true;
+          exnode = new ImplicitCluster(iter->first);
```
julien-tierny: please do not use dynamic allocation here.

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4144 @@
...
+        }
+      }
+
+      ImplicitCluster *exnode;
+      for(auto iter = edgeNodes.begin(); iter != edgeNodes.end(); iter++) {
+        exnode = searchCache(iter->first, nodePtr->nid);
+        bool remove = false;
+        if(!exnode) {
+          remove = true;
+          exnode = new ImplicitCluster(iter->first);
```
julien-tierny: please do not use dynamic allocation here.

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4144 @@
...
+        nodeMaps;
+      for(SimplexId cid = cellIntervals_[nodePtr->nid - 1] + 1;
+          cid <= cellIntervals_[nodePtr->nid]; cid++) {
+        for(SimplexId j = 1; j < verticesPerCell; j++) {
+          if(cellArray_->getCellVertex(cid, j)
+             > vertexIntervals_[nodePtr->nid]) {
+            SimplexId nodeId
+              = vertexIndices_[cellArray_->getCellVertex(cid, j)];
+            if(nodeMaps.find(nodeId) == nodeMaps.end()) {
+              ImplicitCluster *newNode = new ImplicitCluster(nodeId);
```
julien-tierny: please do not use dynamic allocation here.

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4144 @@
...
+              getClusterVertexStars(newNode);
+              // sort the vertex star list
+              std::vector<std::vector<SimplexId>> tmpVec;
+              newNode->vertexStars_.copyTo(tmpVec);
+              for(size_t i = 0; i < tmpVec.size(); i++) {
+                sort(tmpVec[i].begin(), tmpVec[i].end());
+              }
+              nodeMaps[nodeId] = tmpVec;
+              // clear memory
+              delete newNode;
```
julien-tierny: please do not use dynamic allocation here.

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4144 @@
...
+      }
+
+      ImplicitCluster *exnode;
+      for(auto iter = nodeTriangles.begin(); iter != nodeTriangles.end();
+          iter++) {
+        exnode = searchCache(iter->first, nodePtr->nid);
+        bool remove = false;
+        if(!exnode) {
+          remove = true;
+          exnode = new ImplicitCluster(iter->first);
```
julien-tierny: please do not use dynamic allocation here.

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4144 @@
...
+                    } else {
+                      printErr("[CompactTriangulation] More than two other "
+                               "vertices are "
+                               "found in the edge!\n");
+                    }
+                  }
+                }
+                SimplexId nodeId = vertexIndices_[otherEdge[0]];
+                if(nodeMaps.find(nodeId) == nodeMaps.end()) {
+                  ImplicitCluster *newNode = new ImplicitCluster(nodeId);
```
julien-tierny: please do not use dynamic allocation here.

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4144 @@
...
+        }
+      }
+
+      ImplicitCluster *exnode;
+      for(auto iter = edgeNodes.begin(); iter != edgeNodes.end(); iter++) {
+        exnode = searchCache(iter->first, nodePtr->nid);
+        bool remove = false;
+        if(!exnode) {
+          remove = true;
+          exnode = new ImplicitCluster(iter->first);
```
julien-tierny: please do not use dynamic allocation here.

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4144 @@
...
+            cid <= cellIntervals_[nodePtr->nid]; cid++) {
+          std::array<SimplexId, 3> vertexIds
+            = {(SimplexId)cellArray_->getCellVertex(cid, 0),
+               (SimplexId)cellArray_->getCellVertex(cid, 1),
+               (SimplexId)cellArray_->getCellVertex(cid, 2)};
+          // the first vertex of the cell must be in the cluster
+          std::array<SimplexId, 2> edgePair = {vertexIds[1], vertexIds[2]};
+          SimplexId nodeId = vertexIndices_[vertexIds[1]];
+          if(nodeMaps.find(nodeId) == nodeMaps.end()) {
+            ImplicitCluster *newNode = new ImplicitCluster(nodeId);
```
julien-tierny: please do not use dynamic allocation here.

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4144 @@
...
+        }
+        for(SimplexId cid : externalCells_[nodePtr->nid]) {
+          std::array<SimplexId, 3> vertexIds
+            = {(SimplexId)cellArray_->getCellVertex(cid, 0),
+               (SimplexId)cellArray_->getCellVertex(cid, 1),
+               (SimplexId)cellArray_->getCellVertex(cid, 2)};
+          std::array<SimplexId, 2> edgePair = {vertexIds[0], vertexIds[2]};
+          SimplexId nodeId = vertexIndices_[edgePair[0]];
+          if(nodeMaps.find(nodeId) == nodeMaps.end()) {
+            ImplicitCluster *newNode = new ImplicitCluster(nodeId);
```
julien-tierny: please do not use dynamic allocation here.

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4144 @@
...
+               (SimplexId)cellArray_->getCellVertex(cid, 1),
+               (SimplexId)cellArray_->getCellVertex(cid, 2),
+               (SimplexId)cellArray_->getCellVertex(cid, 3)};
+
+          // v1: (v2, v3, v4)
+          std::array<SimplexId, 3> triangleVec
+            = {vertexIds[1], vertexIds[2], vertexIds[3]};
+          SimplexId nodeId = vertexIndices_[vertexIds[1]];
+          if(nodeMaps.find(nodeId) == nodeMaps.end()) {
+            ImplicitCluster *newNode = new ImplicitCluster(nodeId);
```
julien-tierny: please do not use dynamic allocation here.

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4144 @@
...
+            = {(SimplexId)cellArray_->getCellVertex(cid, 0),
+               (SimplexId)cellArray_->getCellVertex(cid, 1),
+               (SimplexId)cellArray_->getCellVertex(cid, 2),
+               (SimplexId)cellArray_->getCellVertex(cid, 3)};
+          // start from v2
+          std::array<SimplexId, 3> triangleVec
+            = {vertexIds[0], vertexIds[2], vertexIds[3]};
+          SimplexId nodeId = vertexIndices_[vertexIds[0]];
+          if(nodeMaps.find(nodeId) == nodeMaps.end()) {
+            ImplicitCluster *newNode = new ImplicitCluster(nodeId);
```
julien-tierny: please do not use dynamic allocation here.

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4144 @@
...
+              }
+            }
+          }
+          // external edges
+          boost::unordered_map<SimplexId, ImplicitCluster *> nodeMaps;
+          for(auto iter = nodePtr->externalEdgeMap_.begin();
+              iter != nodePtr->externalEdgeMap_.end(); iter++) {
+            SimplexId nodeId = vertexIndices_[iter->first[0]];
+            if(nodeMaps.find(nodeId) == nodeMaps.end()) {
+              nodeMaps[nodeId] = new ImplicitCluster(nodeId);
```
julien-tierny: please do not use dynamic allocation here.

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4144 @@
...
+              }
+            }
+          }
+          // external triangles
+          boost::unordered_map<SimplexId, ImplicitCluster *> nodeMaps;
+          for(auto iter = nodePtr->externalTriangleMap_.begin();
+              iter != nodePtr->externalTriangleMap_.end(); iter++) {
+            SimplexId nodeId = vertexIndices_[iter->first[0]];
+            if(nodeMaps.find(nodeId) == nodeMaps.end()) {
+              nodeMaps[nodeId] = new ImplicitCluster(nodeId);
```
julien-tierny: please do not use dynamic allocation here.

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4144 @@
...
```
julien-tierny: please do not use dynamic allocation here.

core/vtk/ttkCompactTriangulationPreconditioning/ttkCompactTriangulationPreconditioning.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+  ttk::Triangulation *triangulation
+    = ttkAlgorithm::GetTriangulation(inputDataSet);
+  if(!triangulation)
+    return 0;
+
+  // Precondition the triangulation (e.g., enable fetching of vertex neighbors)
+  // this->preconditionTriangulation(triangulation); // implemented in base
+  // class
+  this->vertices = new vector<SimplexId>();
```
julien-tierny: please do not use dynamic allocation here.

core/vtk/ttkCompactTriangulationPreconditioning/ttkCompactTriangulationPreconditioning.cpp
```diff
@@ -0,0 +1,207 @@
...
+  this->nodes = new vector<SimplexId>();
```
julien-tierny: please do not use dynamic allocation here.

core/vtk/ttkCompactTriangulationPreconditioning/ttkCompactTriangulationPreconditioning.cpp
```diff
@@ -0,0 +1,207 @@
...
+  this->cells = new vector<SimplexId>();
```
julien-tierny: please do not use dynamic allocation here.

core/vtk/ttkCompactTriangulationPreconditioning/ttkCompactTriangulationPreconditioning.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+    pointData->AddArray(newField);
+  }
+
+  // insert the cells in the output mesh
+  outputMesh->Allocate(this->cells->size());
+  int dimension = triangulation->getCellVertexNumber(0);
+
+  for(unsigned int i = 0; i < this->cells->size(); i++) {
+    vtkIdType *cell = new vtkIdType[dimension];
```
julien-tierny: please do not use dynamic allocation here.

core/base/compactTriangulationPreconditioning/Octree.h
```diff
@@ -0,0 +1,432 @@
+/// \ingroup base
+/// \class Octree
```
julien-tierny: It seems that this octree implementation is fairly generic (it is not specific to the CompactTriangulation) and that it could be useful for others as well. 
If so, please move that class to a new directory `core/base/octree/`

core/vtk/ttkCompactTriangulationPreconditioning/ttkCompactTriangulationPreconditioning.cpp
```diff
@@ -0,0 +1,207 @@
...
+#include <ttkMacros.h>
+#include <ttkUtils.h>
+
+// A VTK macro that enables the instantiation of this class via ::New()
+// You do not have to modify this
+vtkStandardNewMacro(ttkCompactTriangulationPreconditioning);
+
+ttkCompactTriangulationPreconditioning::
+  ttkCompactTriangulationPreconditioning() {
+  this->setDebugMsgPrefix("ttkCompactTriangulationPreconditioning");
```
julien-tierny: please remove this line.
differentiating the base and vtk layers in the messages has been deprecated since the second version of the internal log API.

julien-tierny (changes requested):
hi @guoxiliu 

sorry for the late reply but here's my review for this PR.
if you could fix the few remarks I added in there, that would be wonderful.
maybe @pierre-guillou, you could also review the PR if you have time.

once the above remarks are addressed, I'll go ahead and run experiments with the code.

thanks!

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved.

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved.

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved.

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved.

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved.

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved.

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved.

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved.

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved.

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved.

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved.

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved.

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved.

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved.

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved.

core/base/compactTriangulation/CompactTriangulation.h
julien-tierny: solved.

core/vtk/ttkCompactTriangulationPreconditioning/ttkCompactTriangulationPreconditioning.cpp
julien-tierny: solved.

core/vtk/ttkCompactTriangulationPreconditioning/ttkCompactTriangulationPreconditioning.cpp
julien-tierny: solved.

core/vtk/ttkCompactTriangulationPreconditioning/ttkCompactTriangulationPreconditioning.cpp
julien-tierny: solved.

core/vtk/ttkCompactTriangulationPreconditioning/ttkCompactTriangulationPreconditioning.cpp
julien-tierny: solved.

core/base/compactTriangulationPreconditioning/Octree.h
julien-tierny: solved.

core/vtk/ttkCompactTriangulationPreconditioning/ttkCompactTriangulationPreconditioning.cpp
julien-tierny: solved.

julien-tierny (approved):

core/base/octree/Octree.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+  while(current != nullptr && current->childExists != 0) {
+    computeCenterSize(location, ncenter, nsize);
+    location = getChildLocation(location, vertexId, ncenter);
+    current = lookupNode(location);
+  }
+
+  if(current == nullptr) {
+    OctreeNode newnode(location);
+    newnode.vertexIds = new vector<SimplexId>{vertexId};
```
julien-tierny: please remove this dynamic allocation.

core/base/octree/Octree.cpp
```diff
@@ -0,0 +1,248 @@
...
+      current = lookupNode(location);
+    }
+
+    if(current == nullptr) {
+      this->printErr("[Octree] insertCell(): Cannot find the vertex id ("
+                     + to_string(vertexId) + ") in the tree!");
+      return -1;
+    } else {
+      if(current->cellIds == nullptr) {
+        current->cellIds = new vector<SimplexId>{cellId};
```
julien-tierny: please remove this dynamic allocation.

core/base/octree/Octree.h
```diff
@@ -0,0 +1,218 @@
...
+
+class OctreeNode {
+public:
+  OctreeNode() {
+    OctreeNode(0);
+  }
+  OctreeNode(uint32_t location) {
+    locCode = location;
+    childExists = 0;
+    vertexIds = new vector<SimplexId>();
```
julien-tierny: please remove this dynamic allocation.

julien-tierny (changes requested):
hi @guoxiliu 
thanks a lot for your quick fix.
I have found yet a few occurrences of dynamic allocation.
could you also address these and let me know when you're done?
thanks!

julien-tierny (approved):

julien-tierny:
ok, thanks @guoxiliu 
that looks all good to me.
I'll start to play around and experiment with the code a little bit.
@pierre-guillou could you have a look at this PR too?
thanks

julien-tierny:
as a side note, I switched the `CompactTriangulationPreconditioning` to operate on `vtkPointSet` inputs (and outputs) instead of the more general `vtkDataSet`.
specifically, there is no need to call it on a regular grid for instance (and it actually crashed there).

paraview/xmls/CompactTriangulationPreconditioning.xml
```diff
@@ -0,0 +1,69 @@
...
+          <DataTypeDomain name="input_type">
+            <DataType value="vtkPointSet" />
+          </DataTypeDomain>
+          <!-- <InputArrayDomain name="input_array" attribute_type="point"
+                number_of_components="1" />
+          <Documentation>A vtkDataSet that has at least one point data scalar array that can be used to compute averages.</Documentation> -->
+        </InputProperty>
+
+      <!-- INPUT PARAMETER WIDGETS -->
+        <StringVectorProperty name="SelectedFields" command="SetScalarField" element_types="2 0" information_property="CellArrayInfo" label="Fields" number_of_elements="2" number_of_elements_per_command="1" repeat_command="1">
```
julien-tierny: what is the purpose of this field selection?
as far as I understand, this filter only processes a triangulation, irrespective of the data defined on it.

thanks for your feedback.

julien-tierny: btw, in paraview's gui, turning them off (by unchecking the corresponding checkbox), applying, then turning them on again, then applying, results in a crash.

guoxiliu: > what is the purpose of this field selection? as far as I understand, this filter only processes a triangulation, irrespective of the data defined on it.
> 
> thanks for your feedback.

The purpose here is to get the original fields in the data set and modify the values since the vertex indices could get changed by the clustering method. 


> btw, in paraview's gui, turning them off (by unchecking the corresponding checkbox), applying, then turning them on again, then applying, results in a crash.

I will look into the issue and solve it.

julien-tierny: I see, thanks for the feedback.
if you reproduce (and fix) this issue, that'd be fantastic.

core/base/octree/Octree.h
```diff
@@ -0,0 +1,215 @@
...
+
+#include <Triangulation.h>
+#include <assert.h>
+#include <iostream>
+#include <stack>
+#include <unordered_map>
+#include <vector>
+
+using namespace ttk;
+using namespace std;
```
pierre-guillou: I'm a bit concerned with these `using namespace` directives inside a header file since, through preprocessor inclusion, this can have non-local effects on other parts of the TTK codebase. Would it be possible to remove those and use instead fully qualified types (with `ttk::` and `std::` prefixes)? (`using namespace` in .cpp source files is fine though)

core/base/octree/Octree.h
```diff
@@ -0,0 +1,215 @@
...
+
+class OctreeNode {
+public:
+  OctreeNode() {
+    OctreeNode(0);
```
pierre-guillou: This creates a temporary object that is never used. Did you intend `OctreeNode() : OctreeNode(0) {}` instead? It is flagged by clang-tidy in the `check_code` CI workflow, first `lint_code` job, `Use clang-tidy to lint code` step as `bugprone-undelegated-constructor`

core/base/octree/Octree.cpp
```diff
@@ -0,0 +1,248 @@
+#include <Octree.h>
+
+using namespace ttk;
+
+Octree::Octree(const AbstractTriangulation *t) {
+  Octree(t, 1000);
```
pierre-guillou: This creates a temporary object. Did you intend to call `Octree::Octree(const AbstractTriangulation *t):Octree(t, 1000) {}` instead?

core/base/octree/Octree.cpp
```diff
@@ -0,0 +1,248 @@
...
+int Octree::insertCell(SimplexId &cellId) {
+  if(cellId < 0 || cellId >= triangulation_->getNumberOfCells()) {
+    return -1;
+  }
+
+  int dim = triangulation_->getCellVertexNumber(cellId);
+  uint32_t location;
+  float ncenter[3] = {0.0}, nsize[3] = {0.0};
+  for(int i = 0; i < dim; i++) {
+    SimplexId vertexId;
```
pierre-guillou: Due to the internals of `getCellVertex`, this variable needs to be initialized (otherwise, a garbage value might be propagated). A simple `SimplexId vertexId{};` with two braces should suffice to zero-initialize the variable. This is flagged by the clang static analyzer (`check_code` CI workflow, second `lint_code` job, `Use Clang static analyzer` step).

core/base/octree/Octree.h
```diff
@@ -0,0 +1,215 @@
...
+      }
+    }
+  }
+
+  /**
+   * Get the location code of the child.
+   */
+  uint32_t
+    getChildLocation(uint32_t parLoc, SimplexId vertexId, float centerArr[3]) {
+    float xval, yval, zval;
```
pierre-guillou: Those values should also be zero-initialized: `float xval{}, yval{}, zval{};`

core/vtk/ttkCompactTriangulationPreconditioning/ttkCompactTriangulationPreconditioning.cpp
```diff
@@ -0,0 +1,203 @@
...
+        break;
+
+      case VTK_INT:
+        newField = vtkIntArray::New();
+        break;
+
+      case VTK_ID_TYPE:
+        newField = vtkIdTypeArray::New();
+        break;
+    }
```
pierre-guillou: We might need a `default` label here (all possible VTK types are not covered). Or, in a simpler way, replace the whole switch/case with `vtkSmartPointer<vtkDataArray> newField{inputArray->NewInstance()};`?

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4133 @@
...
+#include <CellArray.h>
+#include <FlatJaggedArray.h>
+#include <algorithm>
+#include <boost/functional/hash.hpp>
+#include <boost/unordered_map.hpp>
+#include <boost/unordered_set.hpp>
+#include <list>
+
+#define EDGE_ID 1
+#define TRIANGLE_ID 2
```
pierre-guillou: I would favor a private (if possible) `enum class` inside `ImplicitCluster` to reduce the scope of these `define`.

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4133 @@
...
+
+  public:
+    CompactTriangulation();
+
+    ~CompactTriangulation();
+
+    /**
+     * Set up vertices from the input.
+     */
+    int setInputPoints(const SimplexId &pointNumber,
```
pierre-guillou: This method should either be marked `inline` or moved to the .cpp source file to enforce the One-Definition Rule (ODR). Moving to the .cpp might also have a positive impact on the build times.

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4133 @@
...
+      }
+
+      return 0;
+    }
+#endif
+
+    /**
+     * Reorder the input vertices.
+     */
+    int reorderVertices(std::vector<SimplexId> &vertexMap) {
```
pierre-guillou: `inline` or move to .cpp?

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4133 @@
...
+        }
+      }
+
+      return 0;
+    }
+
+    /**
+     * Reorder the input cells.
+     */
+    int reorderCells(const std::vector<SimplexId> &vertexMap,
```
pierre-guillou, line 213: `inline` or move to .cpp?

core/base/compactTriangulation/CompactTriangulation.h
```diff
@@ -0,0 +1,4133 @@
...
+        ((LongSimplexId *)connectivity)[i] = newCellArray[i];
+      }
+
+      return 0;
+    }
+
+    /**
+     * Reorder the input cells.
+     */
+    int reorderCells(const std::vector<SimplexId> &vertexMap,
```
pierre-guillou, line 221: `inline` or move to .cpp?

core/vtk/ttkCompactTriangulationPreconditioning/ttkCompactTriangulationPreconditioning.h
```diff
@@ -0,0 +1,97 @@
...
+#include <vtkDataSetAlgorithm.h>
+#include <vtkDoubleArray.h>
+#include <vtkFiltersCoreModule.h>
+#include <vtkFloatArray.h>
+#include <vtkInformation.h>
+#include <vtkIntArray.h>
+#include <vtkObjectFactory.h>
+#include <vtkPointData.h>
+#include <vtkSmartPointer.h>
+#include <vtkUnstructuredGrid.h>
```
pierre-guillou: A lot of those `includes` can be moved to the .cpp

core/base/octree/Octree.h
```diff
@@ -0,0 +1,215 @@
...
+  void reindex(vector<SimplexId> &vertices,
+               vector<SimplexId> &nodes,
+               vector<SimplexId> &cells);
+
+private:
+  const AbstractTriangulation *triangulation_;
+  unordered_map<uint32_t, OctreeNode> allNodes;
+  int capacity;
+  float center[3];
+  float size[3];
```
pierre-guillou: Maybe use `std::array<float, 3>` instead of C-array? `std::array` are zero-initialized and have additional methods (`begin()`, `end()`, ...) that make them compatible with `std::vector`. To get a C-array like pointer, use the `data()` method.

core/vtk/ttkCompactTriangulationPreconditioning/ttkCompactTriangulationPreconditioning.cpp
```diff
@@ -0,0 +1,203 @@
...
+    vertexMap[this->vertices.at(i)] = i;
+    indices->InsertNextTuple1(this->nodes.at(i));
+  }
+  outputMesh->SetPoints(points);
+
+  vtkPointData *pointData = outputMesh->GetPointData();
+  pointData->AddArray(indices);
+
+  std::vector<std::string>::iterator iter = scalarFields.begin();
+  while(iter != scalarFields.end()) {
```
pierre-guillou: A range-based for loop (`for(const auto &s : scalarFields)`) should be more readable here

core/vtk/ttkCompactTriangulationPreconditioning/ttkCompactTriangulationPreconditioning.cpp
```diff
@@ -0,0 +1,203 @@
...
+      outputMesh->InsertNextCell(VTK_TETRA, 4, cell);
+    } else {
+      this->printErr("Should not get here!");
+    }
+  }
+
+  // the cell is the same order as the input mesh
+  vtkCellData *cellData = outputMesh->GetCellData();
+  iter = scalarFields.begin();
+  while(iter != scalarFields.end()) {
```
pierre-guillou: A range-based for loop might be preferred here too

pierre-guillou (commented):
Hi @guoxiliu!
I've looked at this PR and written a few comments on minor points. I hope I'm not too picky...

paraview/xmls/CompactTriangulationPreconditioning.xml
julien-tierny: one quick remark @guoxiliu:
wouldn't it make more sense, instead of shuffling all data arrays, to generate only one new data array storing the correspondence of indices between new and old identifiers?
(for instance called `ttkCompactTriangulationIdentifiers`).

if the input contains a large number of data arrays, only one new data array would need to be produced without shuffling the entire set of input data arrays.

I am also concerned about possible inconsistencies, with regard to the VTK pipeline further down that filter, resulting from swapping data values.

finally, at a conceptual level, again, this filter should not modify the data attached to the input geometry as its purpose is only to provide an alternative geometry representation. in principle, that filter could be applied without any data defined on the geometry. that would argue overall for storing identifier correspondences in an extra array instead of shuffling data arrays.

what do you think?

guoxiliu: Hi @julien-tierny,

I see your reason here for not modifying the original data. The only concern for me with the additional array is how the data array is going to be accessed later (as you mentioned). I don't see an easy way to do that. 

Right now the user can select which data arrays for preconditioning so that the plugin doesn't shuffle all the input data arrays. Though I was still fixing the issue, maybe it's a workaround?

julien-tierny: I experimented a bit with the filter and I agree that there is no easy workaround this unfortunately.
thanks for the info!

julien-tierny:
@pierre-guillou 
could you please double-check @guoxiliu 's latest commits and let me know if you approve all edits?
if so, I guess we'd be good to go for the merge :)

pierre-guillou:
Hi @guoxiliu,

Thanks for taking my remarks into account.
I looked at your last changes and played a bit with the code. I have some last suggestions in the form of the following diffs (you can apply them by copying these files in the root of your TTK repo then using `git am` on them - or reproducing the changes by hand):

[0001-Reduce-variables-scope.txt](https://github.com/topology-tool-kit/ttk/files/7648879/0001-Reduce-variables-scope.txt)
[0002-Clear-state-before-computation.txt](https://github.com/topology-tool-kit/ttk/files/7648881/0002-Clear-state-before-computation.txt)
[0003-Call-parent-class-copy-constructor.txt](https://github.com/topology-tool-kit/ttk/files/7648882/0003-Call-parent-class-copy-constructor.txt)

Also, I noticed several times some weird memory issue that filled the 64GB of RAM of my computer and makes it swap. To trigger them, I load the [noisyTerrain.vtu](https://github.com/topology-tool-kit/ttk-data/raw/dev/noisyTerrain.vtu) dataset in ParaView, apply `CompactTriangulationPreconditionning` then launching `PersistenceDiagram`. Sometimes, this is enough to trigger the phenomenon, otherwise, fiddling with the input scalar fields in the `CompactTriangulationPreconditionning` filter causes it (my 2nd patch might also have an effect there). Using [heaptrack](https://github.com/KDE/heaptrack), I found that the faulty memory allocations occured in `CompactTriangulation::reorderCells`. Have you already noticed this behavior?

Thanks,
Pierre

julien-tierny:
@guoxiliu:
could you have a look at merging @pierre-guillou 's patch?

also, it seems that the memory issue @pierre-guillou mentioned seems to be the last obstacle to the merge of this PR.
this would be great if you could have a look into it.

thanks for your feedback.
best,

guoxiliu:
Thank you, @julien-tierny and @pierre-guillou. I am working on merging the patch and the memory issue. I will update the repo when I fix it. 

Best,
Guoxi

pierre-guillou:
Thanks @guoxiliu! The memory issue seems to be gone with your last commit!
Everything looks good to me.
Looking forward to the merge!

julien-tierny:
Alright, looks good to me.
Let's go then!

julien-tierny:
Now @guoxiliu 
I will ask you one more thing:
can you checkout our [contributor guide to ttk-data](https://github.com/topology-tool-kit/ttk-data/blob/dev/CONTRIBUTING.md#2-adding-a-new-example) and PR ttk-data to provide an example of usage of the compact triangulation?

guoxiliu:
> Alright, looks good to me. Let's go then!

Thank you, @julien-tierny!

> Now @guoxiliu I will ask you one more thing: can you checkout our [contributor guide to ttk-data](https://github.com/topology-tool-kit/ttk-data/blob/dev/CONTRIBUTING.md#2-adding-a-new-example) and PR ttk-data to provide an example of usage of the compact triangulation?

I will follow the guideline in the link you mentioned and prepare the example.

PR #677, opened 2021-10-29
Generate a header for the base layer

pierre-guillou:
This PR generates a unique header, `TopologyToolKit.h`, that includes the main header file of every base layer module.

This header is generated with CMake, using a custom list of edge cases.

This header serves two purposes:
* it makes it easier for (new) TTK users to build simple applications or libraries on top of the TTK base layer and
* it checks if the different base modules can be used together without conflicting with each other.

Using this new header exposes some conflicts between base module. Such conflicts have been fixed in the current PR:
* conflicting header guards between FTMTree and ContourForestsTree, replaced with `#pragma once` in FTMTree
* missing installed headers in PersistenceDiagramClustering and KDTree
* collisions in header file names in case-insensitive filesystems between ContourTree and ContourTreeAlignment (`contourtree.h` has been renamed `CTA_contourtree.h`)
* several structs named `Node`, `Edges` or `ContourTree` have been placed in their own namespace.

The c++ example has been updated to use this new header. This also serves as a non-regression check in the CI.

Enjoy,
Pierre


julien-tierny:
ok

PR #679, opened 2021-11-04
Fix static analysis warnings

pierre-guillou:
This PR fixes some warnings detected either by GCC with the `-Wnull-dereference` flag enabled or by Clang's Static Analyzer. Those fixes includes:
* replacing C arrays with zero-initialized `std::array`s in the Triangulation modules to avoid garbage values,
* avoid returning `nullptr`s in ContourTree and FTMTree, with (helpful?) message errors in non KAMIKAZE mode,
*  adding more checks to avoid segfaults or divisions by zero.

This also reduces the number of warnings in the `check_code / lint-code` CI job.

No change observed in the ttk-data states files.

Enjoy,
Pierre

### DIFF
--- a/core/base/localizedTopologicalSimplification/LocalizedTopologicalSimplification.h
+++ b/core/base/localizedTopologicalSimplification/LocalizedTopologicalSimplification.h
@@ -624,7 +624,7 @@ namespace ttk {
 
         segment.resize(propagation->segmentSize);
         IT segmentIndex = 0;
-        {
+        if(propagation->segmentSize > 0) {
           std::vector<IT> queue(propagation->segmentSize);
           IT queueIndex = 0;
 
@@ -701,7 +701,7 @@ namespace ttk {
           return 0;
 
         // print status
-        if(this->debugLevel_ < 4) {
+        if(this->debugLevel_ < 4 || nPropagations == 0) {
           this->printMsg(msg, 1, timer.getElapsedTime(), this->threadNumber_);
         } else {
 
@@ -944,7 +944,7 @@ namespace ttk {
             = std::numeric_limits<IT>::max();
         }
 
-        if(this->debugLevel_ < 4) {
+        if(this->debugLevel_ < 4 || nPropagations == 0) {
           this->printMsg("Computing Local Order of Segments ("
                            + std::to_string(nPropagations) + ")",
                          1, timer.getElapsedTime(), this->threadNumber_);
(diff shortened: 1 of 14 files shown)


core/base/localizedTopologicalSimplification/LocalizedTopologicalSimplification.h
```diff
@@ -624,7 +624,7 @@ namespace ttk {
 
         segment.resize(propagation->segmentSize);
         IT segmentIndex = 0;
-        {
+        if(propagation->segmentSize > 0) {
```
julien-tierny, line 627: can `segmentSize` be really smaller than 0?

julien-tierny: (note the above resize)

julien-tierny (changes requested):

core/base/localizedTopologicalSimplification/LocalizedTopologicalSimplification.h
pierre-guillou: Well, GCC seemed to think that `propagation->SegmentSize == 0` might be a possible value. This would segfault when dereferencing the queue at line 633.

julien-tierny: this variable is intialized at zero in `core/base/localizedTopologicalSimplification/Propagation.h` and it is only incremented in both files.
I would vote against the insertion of that if statement, which looks unnecessary and may have an impact on performances.
@JonasLukasczyk what's your opinion?

julien-tierny (changes requested):

core/base/localizedTopologicalSimplification/LocalizedTopologicalSimplification.h
JonasLukasczyk: Segmentation size is initialized as 0, but theoretically after all propagation computations any segment must have a size of least 1. So 0 can only happen if the propagation procedure is buggy.

JonasLukasczyk: A much more likely error scenario is when LTS is launched on a non simply connected manifold. This will crash for sure. But checking connectivity before LTS is a huge task.

JonasLukasczyk: Maybe this issue can be detected during the propagation procedure, but this is not happening right now. I would say I can look into this when I have time, but I'm pretty busy right now.

julien-tierny: thanks @JonasLukasczyk for your quick input :)
let's keep the if statement then.

julien-tierny (approved):

PR #680, opened 2021-11-18
Add new Persistence Diagram backend

pierre-guillou:
This PR adds a new Persistence Diagram backend named "Persistent Simplex Pairs". This algorithm is described in "Algorithm and Theory of Computation Handbook (Second Edition) - Special Topics and Techniques" by Atallah and Blanton on page 97.

Since I used the `VisitedMask` data structure from the `DiscreteGradient` module, this structure has been moved in the `common` module.

This algorithm is very slow - 4 to 5 orders of magnitudes slower than FTM. However, it computes saddle-saddle and infinite pairs.

Enjoy,
Pierre

### DIFF
--- a/core/base/common/VisitedMask.h
+++ b/core/base/common/VisitedMask.h
@@ -1,0 +1,64 @@
+/// \ingroup base
+/// \class ttk::VisitedMask
+/// \author Pierre Guillou <pierre.guillou@lip6.fr>
+/// \date November 2021.
+
+#pragma once
+
+#include <DataTypes.h>
+
+#include <algorithm>
+#include <vector>
+
+namespace ttk {
+
+  /**
+   * @brief Auto-cleaning re-usable graph propagations data structure
+   *
+   * This structure stores two vector references. The first vector
+   * maps the nodes of a graph to a boolean (true if the node is
+   * visited, false if not). The second reference points to a vector
+   * that contains the visited nodes indices. Using RAII, the boolean
+   * vector is cleaned using the indices vector when this structure
+   * goes out of scope. See @ref
+   * ttk::MorseSmaleComplex3D::getSaddleConnectors for an example use.
+   *
+   */
+  struct VisitedMask {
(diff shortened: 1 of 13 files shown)


core/base/common/VisitedMask.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
```
julien-tierny, line 6: thanks for adding your name, email address and date.

paraview/xmls/PersistenceDiagram.xml
```diff
@@ -1,67 +1,72 @@
...
+        Persistence diagrams are useful and stable concise representations of the
+        topological features of a data-set. It is useful to fine-tune persistence
+        thresholds for topological simplification or for fast similarity
+        estimations for instance.
 
-Persistence diagrams are useful and stable concise representations of the
-topological features of a data-set. It is useful to fine-tune persistence
-thresholds for topological simplification or for fast similarity
-estimations for instance.
+        Two backends are available for the computation of the persistence diagram:
```
julien-tierny: Two --› Three

julien-tierny (changes requested):
Thanks for this PR
Could you address my couple of remarks?
Thanks

core/base/common/VisitedMask.h
pierre-guillou: Fixed

paraview/xmls/PersistenceDiagram.xml
pierre-guillou: Fixed

julien-tierny (approved):

julien-tierny:
For some reason, the Azure Windows CI is now exceeding its time limit and fails all the time.
As it seems to be redundant with the `test-build-windows` github action, I'd suggest to remove the Azure Windows CI.
What do you think?

julien-tierny:
awesome, thanks!

PR #681, opened 2021-11-19
PersistenceDiagram: Fix PersistentSimplex 2D maxima critical type

pierre-guillou:
This PR fixes the CriticalType of 2D saddle-max pairs for the persistent simplex pairs backend.

Enjoy,
Pierre


julien-tierny:
thanks!

PR #682, opened 2021-11-22
TriangulationRequests: Support multiple indices

pierre-guillou:
With this PR, users can pass comma and/or space separated list of indices to TriangulationRequests.
A pre-processing step removes the duplicates, negative & out-of-bounds values. After this step, if the list is empty, the filter aborts.

Additionally, new CellData arrays were added to display the indices and the dimensions (similarly to the DiscreteGradient/MorseSmaleComplex critical points). The filter also makes use of Jonas' vtkGetEnumMacro/ttkSetEnumMacro.

Enjoy,
Pierre


julien-tierny:
looks cool, thanks!

PR #688, opened 2021-11-26
Doxygen: expand VTK macros

lhofmann:
This PR grabs the include directory from `VTK::CommonCore` if available, and lets doxygen expand the `vtkGetMacro`/`vtkSetMacro` in the docs.

I think it is a bit nicer to have `GetXYZ`, `SetXYZ` instead of `vtkGetMacro(XYZ)` etc.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -2,6 +2,10 @@ add_subdirectory(base)
 
 if(TTK_BUILD_DOCUMENTATION)
   if(DOXYGEN_FOUND)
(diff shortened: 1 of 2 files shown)


lhofmann:
It looks like this only works if the VTK include is in the header file itself, but not if VTK is included transitively (e.g. by including ttkAlgorithm.h but no other VTK header).

Update: This seems to be a limitation of doxygen. The macro expansions are now hard-coded in the doxyfile (copied from ParaView source).

julien-tierny:
awesome, looking forward to giving it a shot :)

PR #689, opened 2021-11-26
Cleanup some cmake files a bit

CharlesGueunet:
Dear Julien,

This MR improves a bit the CMake structure of TTK, trying to keep the main CMakeList lighter.

There is still a point that could be improved; if a user set a
TTK_ENABLE_... to true and the package is not found, CMake still configure
and then the compilation fails. A simple workaround is to add
find_package(... REQUIRED) in the variable is ON for each package.
It should not be costly as it will reuse existing variable for the find
and it will handle unfound packages properly.
Do you want me to handle this in this PR ?

Best,
Charles

### DIFF
--- a/CMake/config.cmake
+++ b/CMake/config.cmake
@@ -1,6 +1,23 @@
+# --- Prerequisites
+
 set(CMAKE_CXX_STANDARD 11)
 
-# --- Options
+# --- Global Options
+
+option(TTK_BUILD_VTK_WRAPPERS "Build the TTK VTK Wrappers" ON)
+cmake_dependent_option(TTK_BUILD_PARAVIEW_PLUGINS "Build the TTK ParaView Plugins" ON "TTK_BUILD_VTK_WRAPPERS" OFF)
+option(TTK_BUILD_STANDALONE_APPS "Build the TTK Standalone Applications" ON)
+option(TTK_WHITELIST_MODE "Explicitely enable each filter" OFF)
+mark_as_advanced(TTK_WHITELIST_MODE BUILD_SHARED_LIBS)
+
+# This option allows library to be built dynamic
+# like the TopologyToolKit.so file for paraview
+option(BUILD_SHARED_LIBS "Build TTK as shared lib" ON)
+
+if(TTK_BUILD_STANDALONE_APPS AND NOT TTK_BUILD_VTK_WRAPPERS)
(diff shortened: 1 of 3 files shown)


CharlesGueunet:
The failed pipeline seems to be a timeout or an other unrelated problem

julien-tierny:
> There is still a point that could be improved; if a user set a
> TTK_ENABLE_... to true and the package is not found, CMake still configure
> and then the compilation fails.

In principle, this should not happen.
If a user sets TTK_ENABLE_XXX=ON and the dependency is not found, then the flag should be turned back to OFF automatically. that is the expected behavior (e.g. scikit-learn, websocketio, etc.). 
thanks for letting me know for which dep you observed this behavior.

CharlesGueunet:
I had this at least with ZFP and WebsocketPP. Looking at the CMakeFile, the option is only first set depending on the `find_package` result, but nothing prevent the user to switch the value to ON after event if the find_package did not find anything, as not REQUIRED have been provided :)

julien-tierny:
> A simple workaround is to add find_package(... REQUIRED) in the variable is ON for each package. 

Well that's precisely the point, we don't want these dependencies to be mandatory. We want them to be optional.

julien-tierny:
I'll go ahead and merge this PR as is, but the problem with the optional dep remains and marking them as `REQUIRED` does not look like the right solution.

CharlesGueunet:
My point is not to replace the existing `find_package` with a **REQUIRED** one, but to add a `find_package(... REQUIRED) when the **TTK_ENABLE...** is ON so we avoid to configure if the package is not found.

This would avoid compil time error like:
```
In file included from /home/charles/Software/sources/ttk/dev/core/vtk/ttkWebSocketIO/ttkWebSocketIO.h:36,
                 from /home/charles/Software/sources/ttk/dev/build/core/vtk/CMakeFiles/ttkWebSocketIOPython/ttkWebSocketIOPython.cxx:10:
/home/charles/Software/sources/ttk/dev/core/base/webSocketIO/WebSocketIO.h:38:10: fatal error: websocketpp/config/asio_no_tls.hpp: No such file or directory
   38 | #include <websocketpp/config/asio_no_tls.hpp>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
ninja: build stopped: subcommand failed.
```

PR #701, opened 2021-12-08
PlanarGraphLayout Template Improvements

JonasLukasczyk:
Hi Julien,
this PR adds some template macros to `ttkMacros.h` which make it easier to templatize over function arguments. They can be called like this:
```
int status = 0;
ttkTypeMacroARIT(
  arrayThatDeterminesT0->GetDataType(), // A: array can be of any data type
  arrayThatDeterminesT1->GetDataType(), // R: array type can only be float or double 
  arrayThatDeterminesT2->GetDataType(), // I: array type can only be an integer type
  triangulationThatDeterminesT3->GetType(), // T: Triangulation type
  (
   status = someFunction<T0,T1,T2,T3>(
    ttkUtils::GetPointer<T0>(arrayThatDeterminesT0),
    ttkUtils::GetPointer<T1>(arrayThatDeterminesT1),
    ttkUtils::GetPointer<T2>(arrayThatDeterminesT2),
    ttkUtils::GetPointer<T3>(triangulation->getData()),
   )
  )
);
if(!status)
  return 0; // some error occurred
```
I provided several common use cases such as `ttkTypeMacroT` (only templatize over triangulation type), `ttkTypeMacroRT` (real and triangulation), and so on. If an array is not of a supported type the macro prints an error msg and the call arguments is not executed. In the example above `status` would therefore not be modified and an error would be detected.

I also updated the PlanarGraphLayout module to use the new macro. This macro refit also fixes a bug that resulted from the fact that the module until now not templatized over the connectivity list type.

NOTE: I tested this with the state files and everything is working as expected. However, immediately feeding the arcs of FTMTree to PlanarGraphLayout and enabling `isMergeTree` causes a segfault, but this already happened before my updates! So this seg fault is a separate issue independent of this PR.

### DIFF
--- a/core/base/planarGraphLayout/PlanarGraphLayout.h
+++ b/core/base/planarGraphLayout/PlanarGraphLayout.h
@@ -38,59 +38,59 @@ namespace ttk {
     PlanarGraphLayout();
     ~PlanarGraphLayout();
 
-    template <class idType, class dataType>
-    int execute(
+    template <typename ST, typename IT, typename CT>
+    int computeLayout(
       // Output
       float *layout,
 
       // Input
-      const LongSimplexId *connectivityList,
+      const CT *connectivityList,
       const size_t &nPoints,
       const size_t &nEdges,
-      const dataType *pointSequences,
+      const ST *pointSequences,
       const float *sizes,
-      const idType *branches,
-      const idType *levels) const;
+      const IT *branches,
+      const IT *levels) const;
 
-    template <class idType>
+    template <typename IT, typename CT>
     int extractLevel(
       // Output
       std::vector<size_t> &nodeIndicies,
       std::vector<size_t> &edgeIndicies,
 
       // Input
-      const LongSimplexId *connectivityList,
+      const CT *connectivityList,
       const size_t &nPoints,
       const size_t &nEdges,
-      const idType &level,
-      const idType *levels) const;
+      const IT &level,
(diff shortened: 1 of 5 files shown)


MatPont:
Hi Jonas, 
Thank you for this PR.

Regarding the note at the end of your message, to use the ```isMergeTree``` option, both the nodes and the arcs of FTMTree should be provided (by maintaining CTRL), for the moment it is only documented in the xml file (but soon in the .h etc.). 
If only the arc is provided it is expected to not work. Still, it would be better to not cause a segfault if the user does not provide correctly the input, I add this to my TODO list.

However, if you found a case where the filter causes a segfault by providing both the nodes and the arcs, could you please open an issue? (or giving me the details to reproduce the problem in order to fix it)
Thank you very much.

JonasLukasczyk:
Thank you @MatPont,
we did not provide both input objects. What information do you need from the points that are not available in the arcs?

The `isMergeTree` issue is not related to the PR so I think it is ready.

MatPont:
> Thank you @MatPont, we did not provide both input objects. What information do you need from the points that are not available in the arcs?
> 
> The `isMergeTree` issue is not related to the PR so I think it is ready.

I need the CriticalType field (that is not present in the arcs output).

julien-tierny:
ok, looks good to me, thanks!

PR #707, opened 2022-01-04
example links

emmwel:
Added links to geometry approximation example for all included modules, (clang-formatted hopefully).

Pull request from ttk-data:

- https://github.com/topology-tool-kit/ttk-data/pull/111


julien-tierny:
wonderful, thanks!

PR #702, opened 2021-12-10
Fix Doxygen warnings

pierre-guillou:
This PR fixes all Doxygen warnings produced by `make doc` or `ninja doc` when `TTK_BUILD_DOCUMENTATION=ON`. It also adds a new job in the `check_code` workflow to ensure that no new Doxygen warning is introduced by new PRs.

Most of these warnings consist in either missing function documentation, misuses of Doxygen syntax or Doxygen shortcomings to fully understand the C++ code. The latter is generally fixed by using fully qualified names (with the `std` or `ttk` prefix) in function signatures.

Lastly, Doxygen's assets handling has been overhauled using the `IMAGE_PATH` and `INCLUDE_DIRS` configuration keys. With `IMAGE_PATH`, Doxygen now copies the `doc/img/splash.png` directly into the build directory, thus making unnecessary to install the top-level `doc` folder.

This should help the documentation to be more in sync with the code.

No change observed in the ttk-data states.

Enjoy,
Pierre

PR #694, opened 2021-11-26
ContourTreeAlignment Doxygen Documentation

floWetzels:
Added doxygen documentation for the ContourTreeAlignment base module. The vtk layer documentatoin was already there. I also cleaned up namespaces for all internal classes and structures that are involved in the alignment computation, since this cluttered the doxygen Class List a lot. They are now all bundled in the ttk::cta:: namespace.

PR #708, opened 2022-01-11
Fix CI Doxygen check

pierre-guillou:
The Doxygen check introduced in #702 was not properly enabled, since the Doxygen process in the virtual machine was outputting warnings on `stderr` instead of `stdout`.

This PR fixes behavior. Now any Doxygen warning will be correctly detected by the check.
Incidentally, some Doxygen typos introduced in #694 have been fixed.

Enjoy,
Pierre


julien-tierny:
looks good.

PR #709, opened 2022-01-11
Ensure no "using namespace" in headers

pierre-guillou:
This PR introduces a new check in the CI that detects if a `using namespace` directive has been put in a header file.

Confer https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice for further reasoning about this check.

(`using namespace` directives in source files should not be affected by this check).

The last module featuring `using namespace` in headers was `PersistenceDiagramClustering`. This PR also removes those declarations and replaces them with fully qualified names. Also, some unused variables (written but never read) were removed.

Enjoy,
Pierre


julien-tierny:
ok!

PR #710, opened 2022-01-11
Enforce Clang static analyzer CI check

pierre-guillou:
This PR fixes the last warning of the Clang static analyzer CI check concerning the `ProjectionFromField` module. To do so, the branches that select the operating mode (`UseTextureCoordinates`, `Use3DCoordinatesArray` or neither)  were hoisted out of the main (parallel) loop to be closer to the corresponding nullptr checks.

As a consequence, the CI check is now enforced.

Enjoy,
Pierre


julien-tierny:
ok

PR #713, opened 2022-01-17
Refactor MorseSmaleComplex base modules

pierre-guillou:
This PR merges all 4 MorseSmaleComplex base modules into one. Nearly 1.1kLoC have been removed from TTK with this operation, with no loss in functionality.

The base layer API has been overhauled: class member pointers were replaced with ad-hoc parameter structs passed by reference to the `execute` method. 1D datasets are now supported alongside 2D and 3D ones.

No performance loss has been observed. No changes observed on ttk-data state files.

Enjoy,
Pierre


julien-tierny:
alright, looks good. 
fingers crossed!

PR #715, opened 2022-01-27
[DimensionReduction] t-SNE and Isomap distance matrix support

MatPont:
Since the last update of the DimensionReduction filter, sklearn has updated t-SNE and Isomap to allow to consider the input as a distance matrix.
This PR modifies the DimensionReduction filter to take these changes into account by adding the distance matrix support for t-SNE and Isomap.


julien-tierny:
looks good, let's go.

Issue #705, opened 2021-12-25
Round macro causes miscompilation on GCC 11.2 (from winlibs)


ghost:
**Describe the bug**
Compilation of localized simplification with WinLibs's GCC (https://github.com/brechtsanders/winlibs_mingw/releases/download/11.2.0-13.0.0-9.0.0-ucrt-r2/winlibs-x86_64-posix-seh-gcc-11.2.0-llvm-13.0.0-mingw-w64ucrt-9.0.0-r2.zip) causes error because of the round macro in Os.h.

I searched the code and the macro does not seem to be used anywhere. However, I do not know why it is there, and removing it may break other builds.

**System (please complete the following information):**
 - TTK version 1.0
 - OS: Windows 10
 - Compiler GCC 11.2
 - No TTK optional dependencies, I compiled the code by hand, see below

Batch file content to compile the code
```
C:\Users\klacansky\tmp\winlibs-x86_64-posix-seh-gcc-11.2.0-llvm-13.0.0-mingw-w64ucrt-9.0.0-r2\mingw64\bin\g++.exe ^
	-O3 -march=native -fopenmp -Wall -DNDEBUG -DTTK_ENABLE_OPENMP -static -static-libgcc -static-libstdc++ ^
	-o localized_topological_simplification_gcc.exe ^
	localized_topological_simplification.cpp ^
	ttk-1.0.0/core/base/abstractTriangulation/abstractTriangulation.cpp ^
	ttk-1.0.0/core/base/common/BaseClass.cpp ^
	ttk-1.0.0/core/base/common/Debug.cpp ^
	ttk-1.0.0/core/base/implicitTriangulation/implicitTriangulation.cpp ^
	-Iboost_1_78_0 -Ittk-1.0.0/core/base/common ^
	-Ittk-1.0.0/core/base/triangulation ^
	-Ittk-1.0.0/core/base/abstractTriangulation ^
	-Ittk-1.0.0/core/base/geometry ^
	-Ittk-1.0.0/core/base/explicitTriangulation ^
	-Ittk-1.0.0/core/base/skeleton ^
	-Ittk-1.0.0/core/base/implicitTriangulation ^
	-Ittk-1.0.0/core/base/periodicImplicitTriangulation ^
	-Ittk-1.0.0/core/base/localizedTopologicalSimplification
```

Error output
```
[...]

PR #716, opened 2022-01-27
Remove unused round macro

pierre-guillou:
This PR removes an unused macro definition that exports the `ttk::OsCall::roundToNearestInt` function with the `round` name.
I believe this macro name is the root cause to issue #705. Since this macro is unused, removing it should solve this issue. The `ttk::OsCall::roundToNearestInt` function is still available for those who need it. 

As an alternative to this removal, we can scope properly the `round` function with 
```cpp
namespace ttk {
  constexpr auto round = OsCall::roundToNearestInt;
}
```

Enjoy,
Pierre

### DIFF
--- a/core/base/common/Os.h
+++ b/core/base/common/Os.h
@@ -15,9 +15,6 @@
 
 #define drand48() (double(rand()) / RAND_MAX)
 //  #define               isnan(x)      _isnan(x)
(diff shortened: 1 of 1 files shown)


julien-tierny:
ok, thanks.

PR #717, opened 2022-01-27
PersistenceDiagramDistanceMatrix: Fix distance computation

pierre-guillou:
The persistenceDiagramDistanceMatrix module was computing the square (L2-) distance between diagrams instead of the distance itself. This PR addresses this bug.

Additionally, the base layer has been refactored with templates to decrease the scope of the `PersistenceDiagramAuction.h` include. This include has been moved from the header file to the cpp file, in order not to expose the included header to the VTK layer.

Enjoy,
Pierre


julien-tierny:
ok, looks good. thanks!

PR #718, opened 2022-01-28
Fixing boundary saddles on volumes

julien-tierny:
The generic backend for scalar field critical points classifies vertices based on the number of connected components of lower and upper links. For manifold domains, this is sufficient in practice up to dimension 3, where the components additionally need to be checked for simply-connectedness (specifically on the boundary). This is what this PR adds.
The previously default backend (progressive computation) hasn't included this fix yet so the generic backend came back as the default for now.

### DIFF
--- a/core/base/contourForestsTree/MergeTreeTemplate.h
+++ b/core/base/contourForestsTree/MergeTreeTemplate.h
@@ -793,7 +793,9 @@ namespace ttk {
       if(params_->simplifyMethod == SimplifMethod::Persist) {
         pairs.emplace_back(
           orig, term,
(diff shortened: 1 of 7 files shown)


julien-tierny:
Some failures occur when checking against certain example pipelines.
For instance, interactionSites (which makes sense).
To be investigated closely.

julien-tierny:
Reported differences only occur for the interactionSites pipeline, which makes sense.
It has been modified in ttk-data.
Re-running checks.

julien-tierny:
Re-runinng checks with pv5.10.0 (used to edit ttk-data)

julien-tierny:
Now validated.

PR #719, opened 2022-01-28
LDistanceMatrix: Improve performance

pierre-guillou:
This PR improves the performance of the `LDistanceMatrix` filter by modifying the parallel execution. Previously, parallelism was managed by the `LDistance` backend module, thus creating a lot of parallel sections (one per LDistance computation).

In this PR, the parallel section has been moved up in the `LDistanceMatrix` base module. Print statements inside the `LDistance` module have also been disabled: the `std::to_string` calls may reduce performance.

The module API has been tweaked to receive typed and const pointers instead of void pointers. A templated `ttkLDistanceMatrix::dispatch` method has been introduced in the VTK layer to access these typed pointers.

Enjoy,
Pierre

### DIFF
--- a/core/base/lDistance/LDistance.h
+++ b/core/base/lDistance/LDistance.h
@@ -51,6 +51,10 @@ namespace ttk {
       return result;
     }
 
+    inline void setPrintRes(const bool data) {
+      this->printRes = data;
+    }
+
     template <typename type>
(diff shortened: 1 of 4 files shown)


julien-tierny:
looks good, thanks!

PR #720, opened 2022-01-28
Add RipsComplex module

pierre-guillou:
On top of #719, this PR adds a new module that compute Rips complexes from either coordinates in a high dimension space or a distance matrix plus some low dimension coordinates (typically a `DimensionReduction` filter output).

Enjoy,
Pierre

### DIFF
--- a/core/base/ripsComplex/RipsComplex.h
+++ b/core/base/ripsComplex/RipsComplex.h
@@ -30,10 +30,16 @@ namespace ttk {
                 const std::vector<std::vector<double>> &inputMatrix) const;
 
   protected:
+    int computeDistanceMatrix(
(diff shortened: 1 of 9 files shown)


core/base/ripsComplex/RipsComplex.h
```diff
@@ -0,0 +1,89 @@
...
+#include <Debug.h>
+
+namespace ttk {
+
+  class RipsComplex : virtual public Debug {
+  public:
+    RipsComplex();
+
+    /**
+     * @brief Main entry point
```
julien-tierny, line 29: nice :)

julien-tierny (commented):
Alright, looks good!

PR #722, opened 2022-01-29
Add TableDistanceMatrix filter

pierre-guillou:
This PR adds a new VTK module, `ttkTableDistanceMatrix`, that calls the `lDistanceMatrix` base module on vtkTables.
The corresponding feature in the `RipsComplex` module (computing a distance matrix from coordinates in a high-dimension space) has been removed. Now, `RipsComplex` only acts on distance matrices (+ a lower dimension embedding).

Enjoy,
Pierre

### DIFF
--- a/core/vtk/ttkTableDistanceMatrix/ttkTableDistanceMatrix.cpp
+++ b/core/vtk/ttkTableDistanceMatrix/ttkTableDistanceMatrix.cpp
@@ -1,0 +1,122 @@
+#include <ttkTableDistanceMatrix.h>
+#include <ttkUtils.h>
+
+#include <vtkDataSet.h>
+#include <vtkDoubleArray.h>
+#include <vtkInformation.h>
+#include <vtkInformationVector.h>
+#include <vtkNew.h>
+#include <vtkObjectFactory.h>
+#include <vtkTable.h>
+
+#include <regex>
+
+vtkStandardNewMacro(ttkTableDistanceMatrix);
+
+ttkTableDistanceMatrix::ttkTableDistanceMatrix() {
+  SetNumberOfInputPorts(1);
+  SetNumberOfOutputPorts(1);
+}
+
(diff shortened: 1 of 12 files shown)


core/vtk/ttkTableDistanceMatrix/ttkTableDistanceMatrix.cpp
```diff
@@ -0,0 +1,122 @@
...
+    arrays.push_back(input->GetColumnByName(s.data()));
+  }
+
+  std::vector<std::vector<double>> inputMatrix(numberOfRows);
+#ifdef TTK_ENABLE_OPENMP
+#pragma omp parallel for num_threads(this->threadNumber_)
+#endif // TTK_ENABLE_OPENMP
+  for(int i = 0; i < numberOfRows; ++i) {
+    for(size_t j = 0; j < numberOfColumns; ++j) {
+      inputMatrix[i].emplace_back(arrays[j]->GetVariantValue(i).ToDouble());
```
julien-tierny, line 81: Note for later @pierre-guillou 
anyway to pass pointers to the raw data down to the base code (like we do for the triangulation?)
(i.e. without input data copy)

pierre-guillou: The `lDistanceMatrix`/`lDistance` modules work on coordinates pointers: one pointer for each point, but the vtkTable arrays are accessed by column. This copy of the input data also serves to transpose the coordinates array to feed it to base modules (while taking care of typing by converting everything to double).
The same thing happens in `ttkDimensionReduction` (from which this snippet is taken).

julien-tierny: Makes sense, thanks!

julien-tierny:
ok, let's go.

PR #723, opened 2022-01-31
DiscreteGradient: Simplify setManifoldSize

pierre-guillou:
This PR simplifies the `DiscreteGradient::setManifoldSize` method by computing directly the manifold sizes from the ascending and the descending manifolds without using maps. This is due to the fact that manifolds are identified following the critical points order.

As a consequence, some segfaults that could be happening when computing a Morse-Smale Complex on Rips complexes should not occur anymore.

Enjoy,
Pierre


julien-tierny (changes requested):
thanks for resolving conflicts with PR #713

julien-tierny (approved):

julien-tierny:
alright, thanks!

PR #725, opened 2022-02-08
GitHub Actions: Add CI for ttk-data Python scripts

pierre-guillou:
This PR adds the GitHub Actions CI bits to run the new ttk-data Python scripts for each CI platform (using the hashes introduced in https://github.com/topology-tool-kit/ttk-data/pull/112). Additionally, it fixes several compatibility bugs in TTK:

* in TopologicalCompression, `size_t` writes were replaced with `uint64_t` (on Windows, `sizeof(size_t) == 4` instead of 8 in the other platforms). Every integer read/write now uses fixed-size integers.
* in MorseSmaleQuadrangulation, when executed in parallel, the separatrices that form the basis of the coarse-grain quadrangulation are now sorted to ensure a deterministic output.
* Python scripts in the `example` directory have been formatted using Black.
* on the Windows GitHub Actions CI, GraphViz is now correctly found by CMake.
* a bug in the MorseSmaleComplex introduced in #713 has been fixed.
* the `fetch-gh-release-asset` action has been updated to `master` since it seems to work better now.

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/ccache.yml
+++ b/.github/workflows/ccache.yml
@@ -46,16 +46,18 @@ jobs:
     - name: Install optional dependencies
       uses: ./.github/actions/install-deps-unix
 
-    - uses: dsaltares/fetch-gh-release-asset@0.0.5
+    - uses: dsaltares/fetch-gh-release-asset@master
       name: Fetch TTK-ParaView headless Debian package
       with:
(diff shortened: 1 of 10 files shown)


julien-tierny:
alright, thanks!

PR #726, opened 2022-02-11
IdentifierRandomizer: Optionally wrap values in compact range

pierre-guillou:
This PR adds an optional parameter (disabled by default) to the IdentifierRandomizer filter in order to wrap the shuffled values into the `[0, #values[` integer range.

If the input values are already in this range, the output should be the same with and without this option.

Enjoy,
Pierre


julien-tierny:
thanks!

PR #727, opened 2022-02-14
Integration: Persistence diagram approximation (LDAV 21)

julesvidal:
Dear Julien, 

This PR integrates the work  pubished in LDAV 2021 on the computation of a persistence diagram approximation, with guarantees on the related Bottleneck error. 

Changes are the following:

- introduction of a new base class **approximateTopology**
- introduction of a new base class **multiresTopology**, which contains utilities shared by approximateTopology and progressiveTopology
- introduction of a new back-end approach (the fourth!) for the computation of the persistence diagram, both in PersistenceDiagram and the paraview filter TTKPersistenceDiagram.
- introduction of a new ParaView filter **TTKPersistenceDiagramApproximation** that only uses the approximation approach to compute the persistence diagram, but also produces the related scalar field approximation and visual indications about the related uncertainty in the diagram. (3 outputs in total in ParaView).
- some documentation about this new approach.

Cheers,

### DIFF
--- a/core/base/approximateTopology/ApproximateTopology.cpp
+++ b/core/base/approximateTopology/ApproximateTopology.cpp
@@ -1,0 +1,1 @@
+#include <ApproximateTopology.h>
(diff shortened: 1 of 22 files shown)


julien-tierny:
ok, looks very good. 
thanks!

PR #728, opened 2022-02-16
Another batch of small fixes

pierre-guillou:
This PR is a collection of small fixes. Among them:
* The Windows CI was referring to my ttk-data fork (typo)
* Removed the WebSocketIO module from the generated TopologyToolKit.h base header, as it would draw the websocket++ library as a package dependency and it's not (yet) packaged on Ubuntu...
* Replaced all instances of `gnu::parallel_sort` with the `PSORT` macro, which does the same thing but factored.
* `CinemaQuery` now filters the input columns to exclude non-scalar columns & columns with a space in their name, in order to prevent SQLite errors.
* `ImplicitTriangulation` preconditions with `ManifoldCheck` in non-kamikaze mode should not return errors any more
* Removed a print in `skeleton` that was dumping the `vertexStars` content in the terminal in the last debug level (the only data structure printed in this module)
* Replaced a Variable-lenght array in `ttkCompactTriangulationPreconditioning` with a `std::array` of fixed size
* Added guards around potential nullptr dereferences in `Octree`
* Other minor things & typos corrections.

Enjoy,
Pierre


core/base/ftmTree/FTMTree_MT.cpp
```diff
@@ -805,20 +805,8 @@ void FTMTree_MT::sortLeaves(const bool para) {
...
-    __gnu_parallel::sort(
-      mt_data_.leaves->begin(), mt_data_.leaves->end(), indirect_sort);
-#else
-    std::sort(mt_data_.leaves->begin(), mt_data_.leaves->end(), indirect_sort);
-#endif
-#else
-    std::sort(mt_data_.leaves->begin(), mt_data_.leaves->end(), indirect_sort);
-#endif
-#endif
+    PSORT(this->threadNumber_)
```
julien-tierny: I'd recommend to rename in the future this macro into `TTK_PSORT`, to make it clear to developers that they can find the definition of the macro within TTK.

core/base/implicitTriangulation/ImplicitTriangulation.cpp
```diff
@@ -493,7 +493,7 @@ const vector<vector<SimplexId>> *
 
 SimplexId ImplicitTriangulation::getVertexEdgeNumberInternal(
   const SimplexId &vertexId) const {
-  return getVertexNeighborNumber(vertexId);
+  return TTK_TRIANGULATION_INTERNAL(getVertexNeighborNumber)(vertexId);
```
julien-tierny, line 496: please report the same fix to `PeriodicImplicitTriangulation.cpp`

core/base/ftmTree/FTMTree_MT.cpp
julien-tierny: (see my other comments on slack)

julien-tierny:
ok, thanks for considering my two suggestions about `PSORT` and `PeriodicImplicitTriangulation`.

julien-tierny (changes requested):
Thanks for considering my 2 remarks.

core/base/implicitTriangulation/ImplicitTriangulation.cpp
pierre-guillou: No need: PeriodicImplicitTriangulation uses fixed values for all these functions.

julien-tierny (approved):

julien-tierny:
alright, looks good!

PR #730, opened 2022-02-22
Connected Component Filter

JonasLukasczyk:
Hi Julien,
this PR adds a new connected components filter which has advantages over the version that already exists in PV:
1. Vertex neighbors are determined with the TTK::Triangulation data structure.
2. The filter can compute connected components of a feature mask on an implicit grid. Lets say we have some volume data with a scalar field where 0 represents the background and 1 the features (e.g., a mask that identifies superlevel set components can be generated with the calculator and the expression: Scalar>SomeIsoValue). The filter will then compute the connectivity inside the mask without having to explicitly derive the geometry of the components. 
3. The filter has two outputs: 1. the segmentation of the input data where each vertex is associated with its component id (or -1 if it belongs to the background), and 2. a point set where each point represents a single connected component (the location of the point is at the center of mass of the associated component and stores its size, id, and seed vertex).

Best
Jonas

### DIFF
--- a/core/base/connectedComponents/ConnectedComponents.h
+++ b/core/base/connectedComponents/ConnectedComponents.h
@@ -1,0 +1,162 @@
+/// \ingroup base
+/// \class ttk::ConnectedComponents
+/// \author Jonas Lukasczyk <jl@jluk.de>
+/// \date 01.02.2019
+///
+/// \brief TTK %connectedComponents processing package.
+///
+/// This filter consumes a scalar field with a feature mask and computes for
+/// each edge connected group of vertices with a non-background mask value a
+/// so-called connected component via flood-filling, where the backgroud is
+/// masked with values smaller-equal zero. The computed components store the
+/// size, seed, and center of mass of each component. The flag
+/// UseSeedIdAsComponentId controls if the resulting segmentation is either
+/// labeled by the index of the component, or by its seed location (which can be
+/// used as a deterministic component label).
+
+#pragma once
+
+// base code includes
+#include <Debug.h>
+#include <Triangulation.h>
+
+#include <stack>
+
+namespace ttk {
+  class ConnectedComponents : virtual public Debug {
+
+    using TID = ttk::SimplexId;
+
+  private:
+    static const int UNLABELED{-2};
+    static const int IGNORE{-1};
+
+  public:
+    struct Component {
+      int seed = -1;
+      float center[3]{0, 0, 0};
+      float size{0};
+    };
+
+    ConnectedComponents() {
+      this->setDebugMsgPrefix("ConnectedComponents");
+    };
+    virtual ~ConnectedComponents() = default;
+
+    int preconditionTriangulation(
+      ttk::AbstractTriangulation *triangulation) const {
+      return triangulation->preconditionVertexNeighbors();
+    };
+
+    template <typename TT = ttk::AbstractTriangulation>
+    int computeFloodFill(int *labels,
+                         std::vector<Component> &components,
+
+                         const TT *triangulation,
+                         const TID seed) const {
+      // get component id
+      const TID componentId = components.size();
+
+      std::stack<TID> stack;
+      stack.push(seed);
+      labels[seed] = componentId;
+
+      float size = 0;
+      float x, y, z;
+      float center[3] = {0, 0, 0};
+
+      while(!stack.empty()) {
+        const auto cIndex = stack.top();
+        stack.pop();
+
+        // update node data
+        triangulation->getVertexPoint(cIndex, x, y, z);
+        center[0] += x;
+        center[1] += y;
+        center[2] += z;
+        size++;
+
+        // add neihbors
+        size_t nNeighbors = triangulation->getVertexNeighborNumber(cIndex);
+        for(size_t i = 0; i < nNeighbors; i++) {
+          TID nIndex;
+          triangulation->getVertexNeighbor(cIndex, i, nIndex);
+          if(labels[nIndex] == this->UNLABELED) {
+            labels[nIndex] = componentId;
+            stack.push(nIndex);
+          }
+        }
+      }
+      center[0] /= size;
+      center[1] /= size;
+      center[2] /= size;
+
+      // create component
+      components.resize(componentId + 1);
+      auto &c = components[componentId];
+      std::copy(center, center + 3, c.center);
+      c.size = size;
+      c.seed = seed;
+
+      return 1;
+    }
+
+    template <typename DT>
+    int initializeOutputLabels(int *labels,
+                               const TID nVertices,
+                               const DT *featureMask = nullptr) const {
+      Timer timer;
+      std::string msg
+        = "Initializing IDs" + std::string(featureMask ? " via Mask" : "");
+      this->printMsg(msg, 0, 0, 1, ttk::debug::LineMode::REPLACE);
+      if(featureMask) {
+        for(TID i = 0; i < nVertices; i++)
+          labels[i] = featureMask[i] > 0 ? this->UNLABELED : this->IGNORE;
+      } else {
+        std::fill(labels, labels + nVertices, this->UNLABELED);
+      }
+      this->printMsg(msg, 1, timer.getElapsedTime(), 1);
+
+      return 1;
+    }
+
+    template <typename TT = ttk::AbstractTriangulation>
+    int computeConnectedComponents(std::vector<Component> &components,
+                                   int *outputLabels,
+                                   const TT *triangulation,
+                                   const bool useSeedAsComponentId
+                                   = false) const {
+
+      TID nVertices = triangulation->getNumberOfVertices();
+
+      {
+        Timer timer;
+        const std::string msg = "Computing Connected Components";
+        this->printMsg(msg, 0, 0, 1, ttk::debug::LineMode::REPLACE);
+
+        for(TID i = 0; i < nVertices; i++)
+          if(outputLabels[i] == this->UNLABELED)
+            this->computeFloodFill<TT>(
+              outputLabels, components, triangulation, i);
+
+        this->printMsg(msg, 1, timer.getElapsedTime(), 1);
+      }
+
+      if(useSeedAsComponentId) {
+        Timer timer;
+        const std::string msg = "Labeling Components by Seed Id";
+        this->printMsg(msg, 0, 0, 1, ttk::debug::LineMode::REPLACE);
+
+        for(TID i = 0; i < nVertices; i++) {
+          auto &cid = outputLabels[i];
+          if(cid >= 0)
+            cid = components[cid].seed;
+        }
+
+        this->printMsg(msg, 1, timer.getElapsedTime(), 1);
+      }
+
+      return 1;
+    }
+  };
+} // namespace ttk
--- a/core/vtk/ttkConnectedComponents/ttkConnectedComponents.cpp
+++ b/core/vtk/ttkConnectedComponents/ttkConnectedComponents.cpp
@@ -1,0 +1,167 @@
+#include <ttkConnectedComponents.h>
+
+#include <vtkDataArray.h>
+#include <vtkInformation.h>
+#include <vtkObjectFactory.h>
+#include <vtkPointData.h>
+#include <vtkPolyData.h>
+#include <vtkSmartPointer.h>
+
+#include <vtkFloatArray.h>
+#include <vtkIntArray.h>
+
+#include <ttkMacros.h>
+#include <ttkUtils.h>
+
+vtkStandardNewMacro(ttkConnectedComponents);
+
+ttkConnectedComponents::ttkConnectedComponents() {
+  this->SetNumberOfInputPorts(1);
+  this->SetNumberOfOutputPorts(2);
+
+  // Suppress warning if one does not set the optional input array
+  this->SetInputArrayToProcess(0, 0, 0, 0, "%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%");
+}
+
+int ttkConnectedComponents::FillInputPortInformation(int port,
+                                                     vtkInformation *info) {
+  if(port == 0) {
+    info->Set(vtkAlgorithm::INPUT_REQUIRED_DATA_TYPE(), "vtkDataSet");
+    return 1;
+  }
+  return 0;
+}
+
+int ttkConnectedComponents::FillOutputPortInformation(int port,
+                                                      vtkInformation *info) {
+  if(port == 0) {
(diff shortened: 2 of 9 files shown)


JonasLukasczyk:
@julien-tierny I don't understand the lint issue but the rest looks good.

pierre-guillou:
The lint issue might be related to the `@ConnectedComponents` command in `ttkConnectedComponents.h` line 9 (in the `\brief` paragraph starting line 6), which is an invalid Doxygen command (Doxygen commands usually are prefixed with `@` or `\`). I suggest to try replacing it with `ttk::ConnectedComponents` (see for instance `ttkHelloWorld.h`). This way, Doxygen should be able to find the base module and generate an HTML link.

Pierre

JonasLukasczyk:
Thank you Pierre! I fixed it.

core/base/connectedComponents/ConnectedComponents.h
```diff
@@ -0,0 +1,159 @@
...
+    struct Component {
+      int seed = -1;
+      float center[3]{0, 0, 0};
+      float size{0};
+    };
+
+    ConnectedComponents() {
+      this->setDebugMsgPrefix("ConnectedComponents");
+    };
+    ~ConnectedComponents(){};
```
vollia: should be virtual if class is not marked as final

core/base/connectedComponents/ConnectedComponents.h
```diff
@@ -0,0 +1,159 @@
...
+/// labeled by the index of the component, or by its seed location (which can be
+/// used as a deterministic component label).
+
+#pragma once
+
+// base code includes
+#include <Debug.h>
+#include <Triangulation.h>
+
+typedef ttk::SimplexId TID;
```
vollia: prefere `using` over `typedef`

core/base/connectedComponents/ConnectedComponents.h
```diff
@@ -0,0 +1,159 @@
...
+                         std::vector<Component> &components,
+                         std::vector<TID> &stack,
+
+                         const TT *triangulation,
+                         const TID seed) const {
+      // get component id
+      const TID componentId = components.size();
+
+      int stackSize = 1;
+      stack[0] = seed;
```
vollia: why not using a proper std::stack ?

core/base/connectedComponents/ConnectedComponents.h
```diff
@@ -0,0 +1,159 @@
...
+
+    template <typename DT>
+    int initializeOutputLabels(int *labels,
+                               const TID nVertices,
+                               const DT *featureMask = nullptr) const {
+      if(featureMask) {
+        for(TID i = 0; i < nVertices; i++)
+          labels[i] = featureMask[i] > 0 ? this->UNLABELED : this->IGNORE;
+      } else {
+        for(TID i = 0; i < nVertices; i++)
```
vollia: `std::fill` ?

core/base/connectedComponents/ConnectedComponents.h
```diff
@@ -0,0 +1,159 @@
...
+            outputLabels, components, stack, triangulation, i);
+
+      this->printMsg(msg, 1, timer.getElapsedTime(), 1);
+
+      if(useSeedAsComponentId) {
+        timer.reStart();
+        msg = "Labeling Components by Seed Id";
+        this->printMsg(msg, 0, 0, 1, ttk::debug::LineMode::REPLACE);
+
+        for(TID i = 0; i < nVertices; i++) {
```
vollia, line 150: `for(auto cid : outputLables)` ?

core/vtk/ttkConnectedComponents/ttkConnectedComponents.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+ttkConnectedComponents::ttkConnectedComponents() {
+  this->SetNumberOfInputPorts(1);
+  this->SetNumberOfOutputPorts(2);
+
+  // Suppress warning if one does not set the optional input array
+  this->SetInputArrayToProcess(0, 0, 0, 0, "%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%");
+}
+
+ttkConnectedComponents::~ttkConnectedComponents() {
```
vollia: `= default`

core/vtk/ttkConnectedComponents/ttkConnectedComponents.cpp
```diff
@@ -0,0 +1,170 @@
...
+}
+
+int ttkConnectedComponents::FillInputPortInformation(int port,
+                                                     vtkInformation *info) {
+  if(port == 0) {
+    info->Set(vtkAlgorithm::INPUT_REQUIRED_DATA_TYPE(), "vtkDataSet");
```
vollia, line 29: not sur about this, only dataset supported by the ttkTriangulation are supported no ?

core/vtk/ttkConnectedComponents/ttkConnectedComponents.h
```diff
@@ -0,0 +1,64 @@
...
+public:
+  vtkSetMacro(UseSeedIdAsComponentId, bool);
+  vtkGetMacro(UseSeedIdAsComponentId, bool);
+
+  static ttkConnectedComponents *New();
+  vtkTypeMacro(ttkConnectedComponents, ttkAlgorithm);
+
+protected:
+  ttkConnectedComponents();
+  ~ttkConnectedComponents();
```
vollia: virtual

CharlesGueunet (commented):
Some minor remarks :)

core/vtk/ttkConnectedComponents/ttkConnectedComponents.cpp
julien-tierny: `vtkDataSet` includes `vtkImageData` and `vtkUnstructuredGrid` which are both supported by the triangulation.

vollia: The problem is that it also includes types that are not (`vtkStructuredGrid` for example). When using the VTK module, this may lead to a TTK filter being called on a bad dataset and a crash. We may call the `ttkTriangulation::FillInputPortInformation` for the case where `port == 0` to keep the type checking clean.

JonasLukasczyk: Yes, but this is a problem all over the place. Theoretically we would have to check if the input is either a `vtkPolyData`, `vtkUnstructuredGrid`, or `vtkImageData` object. But in all filters so far we just check for `vtkDataSet` and return an error if it is not possible to derive a triangulation. If we decide to fix this I would suggest doing it in an own PR that fixes this across the board, but I would just keep it like this.

vollia: all right :)

julien-tierny: `core/vtk/ttkAlgorithm/ttkTriangulationFactory.cpp` has various checks implemented already and report warning messages on the standard output if the input is not supported (for instance: unsupported class [NOT: vtkImageData, vtkPolytData, vtkUnstructuredGrid], cells are not simplices, cells have non-uniform dimensions, etc.).
I believe this is a sufficient safeguard for the moment, isn't it?

vollia: I think this is a sufficient safeguard, as it should not lead to a crash. If my understanding is correct, these checks are called on the `RequesetData` pass. Filling the `InputPortInformation` with the valid datasets only may allow the pipeline for an early exit with a clear message.

julien-tierny: you can try to the following to see the safeguard in action:
```
from paraview.simple import *
cleantoGrid1 = CleantoGrid(registrationName='CleantoGrid1', Input=wavelet1)
tTKScalarFieldCriticalPoints1 = TTKScalarFieldCriticalPoints(registrationName='TTKScalarFieldCriticalPoints1', Input=cleantoGrid1)
UpdatePipeline(time=0.0, proxy=tTKScalarFieldCriticalPoints1)
```

JonasLukasczyk: I also think it is safe, but I can see Charle's point that it would be nice that the ttk filters would only accept triangulable input data objects. Like I said this can be achieved by replacing `info->Set(vtkAlgorithm::INPUT_REQUIRED_DATA_TYPE(), "vtkDataSet");` with the checks for the supported three data types:
```
info->Remove(vtkAlgorithm::INPUT_REQUIRED_DATA_TYPE());
info->Append(vtkAlgorithm::INPUT_REQUIRED_DATA_TYPE(), "vtkImageData");
info->Append(vtkAlgorithm::INPUT_REQUIRED_DATA_TYPE(), "vtkPolyData");
info->Append(vtkAlgorithm::INPUT_REQUIRED_DATA_TYPE(), "vtkUnstructuredGrid");
```

julien-tierny: you should get the following output:
```
[ScalarFieldCriticalPoints] %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
[TriangulationFactory] [WARNING] Cells are not simplices.
[TriangulationFactory] [WARNING] Consider using `vtkTetrahedralize` in pre-processing.
[ScalarFieldCriticalPoints] [ERROR] Unable to retrieve/initialize triangulation for 'vtkUnstructuredGrid'
(  17.889s) [paraview        ]       vtkExecutive.cxx:752    ERR| vtkPVCompositeDataPipeline (0x5571da9ea770): Algorithm ttkScalarFieldCriticalPoints(0x5571d877d4b0) returned failure for request: vtkInformation (0x5571d8d6e530)
  Debug: Off
  Modified Time: 1319979
  Reference Count: 1
  Registered Events: (none)
  Request: REQUEST_DATA
  FROM_OUTPUT_PORT: 0
  ALGORITHM_AFTER_FORWARD: 1
  FORWARD_DIRECTION: 0
```

julien-tierny: > I also think it is safe, but I can see Charle's point that it would be nice that the ttk filters would only accept triangulable input data objects. Like I said this can be achieved by replacing `info->Set(vtkAlgorithm::INPUT_REQUIRED_DATA_TYPE(), "vtkDataSet");` with the checks for the supported three data types:
> 
> ```
> info->Remove(vtkAlgorithm::INPUT_REQUIRED_DATA_TYPE());
> info->Append(vtkAlgorithm::INPUT_REQUIRED_DATA_TYPE(), "vtkImageData");
> info->Append(vtkAlgorithm::INPUT_REQUIRED_DATA_TYPE(), "vtkPolyData");
> info->Append(vtkAlgorithm::INPUT_REQUIRED_DATA_TYPE(), "vtkUnstructuredGrid");
> ```

that only partially solves the issue as vtkPolyData and vtkUnstructuredGrid can still represent some invalid inputs for TTK, for instance cells which are not simplices (e.g. quads) or with non-uniform dimensions (e.g. mixture of edges and triangles).

vollia: @julien-tierny this check is nice, and for the manifold / simplices check we have no choice but to do them in Request Data as we need the data set to be available :)

@jonas this is indeed what we want, but this would make sense at a TTK wide level. So I would argue to keep this filter TTK compliant and let this change as a futur work

core/base/connectedComponents/ConnectedComponents.h
JonasLukasczyk: The issue with missing virtual destructors is also present in a lot of ttk filters (including hello world). I will fix this here but this should also be addressed in an own PR.

core/base/connectedComponents/ConnectedComponents.h
JonasLukasczyk: OutputLabels is a pointer and therefore automatic iteration is not possible.

CharlesGueunet (approved):

julien-tierny:
alright, looks good, thanks!

@charles: could you add a comment to this PR with a summary of the low-level recommendations you gave (virtual destructor, default destructor, etc.) thanks!

julien-tierny:
@JonasLukasczyk 
wouldn't it be useful to add a numerical parameter to the filter to allow the user to select the isovalue corresponding to background? (for instance, given a black & white image, e.g. `naturalImage.png`, consider everything below 128 as background)

julien-tierny:
right now, I have the impression that users need to use a calculator in a pre-process to shift the data such that the desired background boundary corresponds to the 0-level set.

julien-tierny:
also, I believe it would be very useful to add the "Size" information on the "Segmentation" output as well, such that users could use the "Threshold" to remove tiny components very easily

CharlesGueunet:
Some recommendations that may be useful for all TTK Filters:
* [ ] Use a virtual destructor for non final class (may avoid memory leaks)
* [ ] Initialize prameters in the header (when possible) so we can have a default constructor / destructor (`=default`)
* [ ] Create a `ttkAlgorithm::FillInputPortInformation` that set the accepted input dataset to those supported by the triangulation so all `ttkAlgorithm` can check the input type early.  (cf. https://github.com/topology-tool-kit/ttk/pull/730#discussion_r813668590)
* [ ] (optional) prefer `using` over `typedef`

If you have any question, I may help :)

julien-tierny:
thanks a lot @CharlesGueunet 

@JonasLukasczyk 
could you PR these edits to HelloWorld? (to make sure future instances get off on the right foot)

JonasLukasczyk:
That all sounds good. I will move those comments to the issues.

julien-tierny:
> That all sounds good. I will move those comments to the issues.

thanks!
that'd be even better if you could PR HelloWorld (to make sure from now on, new modules have that in)

pierre-guillou:
Hi @CharlesGueunet,

> Use a virtual destructor for non final class (may avoid memory leaks)

Since I am not very familiar with this particular aspect of C++, do you have any pointer on that? Should we use `override` instead of `virtual` (since we override `virtual BaseClass::~BaseClass()`)? What happens if we just omit to declare the destructor in the derived classes when they don't own resources (Rule of Zero)?

Also, is there a way to automatically verify that? For instance with a clang-tidy lint? Or some automated testing?

CharlesGueunet:
Both are equivalent here. I prefer `virtual` in the case of the destructor because under the hood destructors are chained and not overriden. Some people prefer `override` for consistency (this is the case in VTK). The initial issue was that none of them was given.

You are right about the zero rules, for class with no field or classes where all fields are trivially destructible (no pointer), no need to specify the destructor.

For the clang tidy, I do not know, I will ask.

PR #734, opened 2022-02-25
[MergeTreeClustering] cap alpha

MatPont:
This PR caps the alpha parameter of the MergeTreeClustering filter to avoid 0 and 1 value.


julien-tierny:
thanks for the quick fix!

PR #735, opened 2022-02-25
ttkConnectedComponents Upgrades

JonasLukasczyk:
Hi,
as suggested by @julien-tierny this PR upgrades the ttkConnectedComponents filter which brings the following features:
- Connected components will always have a deterministic id
- The segmentation can optionally record the size of each component
- The background threshold can now be chosen dynamically and therefore simple threshold based feature masks (which are a very common use case) no longer require the prior execution of a calculator filter.


julien-tierny:
alright, looks good. thanks!

PR #736, opened 2022-02-25
[PDCLustering] Fixes: global pair duplicate in centroids and matching…

julesvidal:
This PR adds some fixes to PersistenceDiagramClustering:

1) The global minimum-maximum pair is no longer duplicated in the centroids when the number of clusters is larger than 1. It also has the correct pairType (-1) and criticalType for the extremities.
This issue resulted in a bad localization of the centroids when performing a dimension reduction on the wasserstein matrix between diagrams and centroids (see image below).
 
old_mds_isabel

This fix results in a better visualization on the isabel ensemble: 
new_mds_isabel

2) It's now possible to explore all the mathings involved in a Wasserstein distance computation between two diagrams. Previously, some diagonal matchings were absent from the result displayed in ParaView. Additionnally, the cost of the matchings was 0 for half of them, this has been also fixed.
This fix will change the images produced by persistenceDiagramDistance.pvsm in ttk-data, and I expect it to trigger an error in the automatic testings.
new_matchings

Best,

### DIFF
--- a/core/base/persistenceDiagramClustering/PDClustering.h
+++ b/core/base/persistenceDiagramClustering/PDClustering.h
@@ -119,8 +119,15 @@ namespace ttk {
     void updateClusters();
     void invertClusters();
     void invertInverseClusters();
-    void computeBarycenterForTwo(
+    void computeBarycenterForTwoGlobal(
       std::vector<std::vector<std::vector<std::vector<matchingTuple>>>> &);
+
+    void computeBarycenterForTwo(
+      std::vector<std::vector<matchingTuple>> &matchings,
+      std::vector<std::vector<int>> &bidders_ids,
+      std::vector<BidderDiagram<dataType>> &current_bidder_diagrams,
+      std::vector<BidderDiagram<dataType>> &bidder_diagrams,
+      GoodDiagram<dataType> &barycenter);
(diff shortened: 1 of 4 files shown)


julien-tierny:
Looks great, thanks for the quick fix.

Quick question though:

> The global minimum-maximum pair is no longer duplicated in the centroids when the number of clusters is larger than 1.

What happens if the number of cluster is 1? Will the barycenter be correctly located in the 2D projection?

julesvidal:
If the number of clusters is 1, the global pair is correctly set in the output barycenter, as it was the case before.
In the 2D projection, the barycenter is not really close from the members of the ensemble:
barycenter_isabel_mds
 However in my opinion, that's to be expected. There are 3 trends in the ensemble, and as an average diagram, the barycenter is not really close from any of them, as it can be seen in the matrix.

julesvidal:
Hi Julien, 
I indeed needed to fix the type and index of the global pair in the output barycenter for it to be correctly picked up in PersistenceDiagramDistanceMatrix. After this last fix, this is the result of the 2d projection:
mds_barycenter
Cheers,

julien-tierny:
looks much better, thanks!

PR #737, opened 2022-02-28
Removed unnecessary include in ttkMergeTreeClustering

JonasLukasczyk:
This PR removes `vtkGeometryFilter.h` from `ttkMergeTreeClustering.cpp`. This header does not seem to be necessary and the include causes a build error when using a newer VTK version.


julien-tierny:
alright, let's go

Issue #738, opened 2022-03-01
MergeTreeTemporalReduction segfault


julien-tierny:
@MatPont thanks!

PR #731, opened 2022-02-24
[MergeTreeVisu] add vertexId

MatPont:
This PR copies the vertexId field of FTMTree filter to the output of MergeTreeClustering or PlanarGraphLayout (in merge tree mode).

PR #739, opened 2022-03-01
[MergeTreeVisu] fix vertexID

MatPont:
This PR fixes the PR #731 and solves the issue #738 .

The source of the problem was that the interpolated trees, made with the temporal reduction algorithm, don't have nodes and arcs (output of FTM) like the other trees.

Enjoy.


julien-tierny:
awesome, thanks!

julien-tierny:
fix #738

PR #740, opened 2022-03-07
OpenMP: Fix stored number of threads

pierre-guillou:
This PR fixes an OpenMP shortcoming introduced in #728. As it happens, `omp_get_num_threads` is not the complementary to `omp_set_num_threads` but merely indicates the *current* number of threads in use (1 in a sequential context). This can lead to setting an incorrect number of thread and making the execution of following filters sequential. To reproduce, compute twice a DiscreteGradient over any dataset in ParaView, the first order field computation will trigger the bug.

According to the [OpenMP documentation](https://www.openmp.org/spec-html/5.0/openmpsu112.html), `omp_set_num_threads`'s complement is in fact `omp_get_max_threads` since both access the *nthreads-var* ICV which
> controls the number of threads requested for encountered parallel regions.

This PR modifies TTK's code accordingly.

Enjoy,
Pierre

### DIFF
--- a/core/base/common/OpenMP.h
+++ b/core/base/common/OpenMP.h
@@ -18,7 +18,7 @@ namespace ttk {
 
   public:
     ParallelGuard(const int nThreads)
-      : oldThreadNumber_{omp_get_num_threads()} {
+      : oldThreadNumber_{omp_get_max_threads()} {
       omp_set_num_threads(nThreads);
     }
     ~ParallelGuard() {


core/base/common/OpenMP.h
```diff
@@ -18,7 +18,7 @@ namespace ttk {
 
   public:
     ParallelGuard(const int nThreads)
-      : oldThreadNumber_{omp_get_num_threads()} {
+      : oldThreadNumber_{omp_get_max_threads()} {
```
julien-tierny, line 21: I believe this will restore the default number of threads (cf `OMP_NUM_THREADS`) regardless of what the variable `threadNumber_` might have been set by the user. to discuss.

julien-tierny (changes requested):
let's discuss that in person this week.
thanks!

PR #741, opened 2022-03-07
Developer experience improvements

pierre-guillou:
This PR introduce 3 new CMake cache keys that might be useful for TTK developers:
1. `TTK_TIME_TARGETS` (OFF by default) prints the build time of every C++ compilation unit. It helps to diagnose where the compilation takes time;
2. `TTK_SCRIPTS_PATH` (defaulting to `$prefix/scripts/ttk`) allows to set the path of TTK scripts (`dimensionReduction.py`) at compile time. It can be useful for install-less uses of TTK where it is points directly to the `core/base/dimensionReduction` source folder;
3. `TTK_REDUCE_TEMPLATE_INSTANTIATIONS` (OFF by default) redefines the `vtkTemplateMacro` to reduce the number of types used in TTK (only `int`, `long long`, `float` & `double`). This greatly reduces the number of template instantiations and thus the edit-compile-run cycle (scalar fields of unsupported types can still be converted to `float` or `double` with a `Calculator`).

These three variables are marked as advanced and should only be used by TTK developers, especially the 3rd one, which has been only tested so far on my Linux computer(s).

What do you think?
Pierre

### DIFF
--- a/CMake/VTKModule.cmake
+++ b/CMake/VTKModule.cmake
@@ -52,6 +52,10 @@ macro(ttk_add_vtk_module)
       vtk_module_definitions(${TTK_NAME} PRIVATE TTK_ENABLE_DOUBLE_TEMPLATING)
     endif()
 
+    if(TTK_REDUCE_TEMPLATE_INSTANTIATIONS)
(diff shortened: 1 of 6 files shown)


julien-tierny:
looks good, thanks!

PR #742, opened 2022-03-09
ttkConnectedComponent: removed static keyword from member variables.

floWetzels:
I removed the static keyword from the member variables of ttkConnectedComponents as some compilers seem to have problems with static const member variables in header files. According to @JonasLukasczyk, this does not change the behaviour of the filter anyway, so it seems a reasonable fix.


julien-tierny:
okidoki!

PR #744, opened 2022-03-14
Cache DiscreteGradient internal structure

pierre-guillou:
This PR adds a caching system to the DiscreteGradient internal data structure.
The AbstractTriangulation is now used to store this data structure. The DiscreteGradient::gradient_ member variable is now a pointer to the cached data structure.
The cache key is a pair (scalar field buffer pointer, scalar field last modified time). The VTK layer modules have been adapted to pass the modified time (`vtkObject::GetMTime()`) to the DiscreteGradient class.

Enjoy,
Pierre


julien-tierny:
ok, thanks!

PR #745, opened 2022-03-15
PersistenceDiagram: Support infinite pairs, improve data structures, VTU representations

pierre-guillou:
This PR overhauls the `PersistenceDiagram` data structures to easily convert between a VTU file and an internal representation using a more expressive `struct` whose goal is to be used in `PersistenceDiagram` client modules (`Clustering`, `DistanceMatrix`, `BottleneckDistance`...) instead of a non-documented and templated tuple.

In this data-structure, a new member variable has been introduce to characterize the finitude of the pair. The global min-max pair is marked as infinite, as the other infinite pairs found by the `PersistentSimplex` back-end.

To factor conversions from and to the VTU format, new functions are introduced into `ttkPersistenceDiagramUtils.cpp`. These functions, currently only used in `ttkPersistenceDiagram` and `ttkPersistenceDiagramApproximation`, help switching between the VTU representation and the newly-introduced struct. Conversions functions between the canonical and the embedded VTU representations have been moved from `ttkProjectionFromFields` to this new file.

The VTU representation has also been updated with these changes:
* a new CellData array named `IsFinite` marks infinite pairs.
* the `Birth` PointData array, previously only created for the embedded VTU representation, is now present in every VTU diagram in order to make canonical representations translation-independent (for instance for `Clustering`'s "Clusters as Stars" option). The array has been moved to CellData.
* the `Death` PointData array has been removed. It can be computed on CellData by adding the `Birth` and `Persistence` arrays.
* consequently, the `Coordinates` PointData array is the only change (modulo the diagonal) between the embedded and the canonical representation.
* `Birth` and `Persistence` CellData arrays are now of the same vtkType than the input scalar field.
* the global min-max pair has now a `PairType` of 0 (previously it was -1, same as the diagonal).

Further PRs will add the support of this new internal data-structure and these conversion methods to `BottleneckDistance`/`TrackingFromFields`/`TrackingFromPersistenceDiagrams` on one hand and `PersistenceDiagramClustering`/`PersistenceDiagramDistanceMatrix` on the other hand.

The ttk-data Python scripts should expose the changes in the Persistence Diagrams VTU representations. A ttk-data PR should follow to update the hashes database. Otherwise, no change should be observed in the ttk-data state file screenshots.

Enjoy,
Pierre


julien-tierny:
ok, thanks!

PR #747, opened 2022-03-24
Display FTM, FTR skeleton nodes in ParaView

pierre-guillou:
This PR refactors #588 (and to a lesser extent #592) by refactoring the code used for generating vtkCells for a point cloud. A new function has been created in `ttkUtils` and is used in the relevant modules: `DiscreteGradient`, `GaussianPointCloud`, `MorseSmaleComplex` and `ScalarFieldCriticalPoints`.

Two other modules have been added to this list: `FTMTree` and `FTRGraph` with their `Skeleton Nodes` outputs.

Enjoy,
Pierre

### DIFF
--- a/core/vtk/ttkAlgorithm/ttkUtils.cpp
+++ b/core/vtk/ttkAlgorithm/ttkUtils.cpp
@@ -361,7 +361,7 @@ int ttkUtils::CellVertexFromPoints(vtkDataSet *const dataSet,
     const auto vtu{vtkUnstructuredGrid::SafeDownCast(dataSet)};
     if(vtu != nullptr) {
       vtu->SetPoints(points);
-      vtu->SetCells(VTK_VERTEX, cells);
+      vtu->SetCells(VTK_POLY_VERTEX, cells);
     }
   } else if(dataSet->IsA("vtkPolyData")) {
     const auto vtp{vtkPolyData::SafeDownCast(dataSet)};
--- a/core/vtk/ttkAlgorithm/ttkUtils.cpp
+++ b/core/vtk/ttkAlgorithm/ttkUtils.cpp
@@ -1,14 +1,19 @@
+#include <BaseClass.h>
 #include <ttkUtils.h>
 
 #include <limits>
-#include <vtkStringArray.h>
 
 #include <vtkAbstractArray.h>
 #include <vtkCellArray.h>
 #include <vtkDoubleArray.h>
 #include <vtkFieldData.h>
+#include <vtkIdTypeArray.h>
+#include <vtkNew.h>
 #include <vtkPoints.h>
+#include <vtkPolyData.h>
 #include <vtkSmartPointer.h>
+#include <vtkStringArray.h>
+#include <vtkUnstructuredGrid.h>
 
 int ttkUtils::replaceVariable(const std::string &iString,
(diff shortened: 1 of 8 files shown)


core/vtk/ttkAlgorithm/ttkUtils.cpp
```diff
@@ -318,3 +323,54 @@ void ttkUtils::FillCellArrayFromDual(vtkIdType const *cells_co,
...
+  }
+  offsets->SetTuple1(nPoints, nPoints);
+  vtkNew<vtkCellArray> cells{};
+  cells->SetData(offsets, connectivity);
+
+  if(dataSet->IsA("vtkUnstructuredGrid")) {
+    const auto vtu{vtkUnstructuredGrid::SafeDownCast(dataSet)};
+    if(vtu != nullptr) {
+      vtu->SetPoints(points);
+      vtu->SetCells(VTK_VERTEX, cells);
```
vollia: using `VTK_POLY_VERTEX` may use less memory for the same result (https://raw.githubusercontent.com/Kitware/vtk-examples/gh-pages/src/Testing/Baseline/Cxx/GeometricObjects/TestLinearCellDemo.png)
( A single cell for all points, thus shorter offset array)

CharlesGueunet (approved):
+1 after memory improvement

core/vtk/ttkAlgorithm/ttkUtils.cpp
pierre-guillou: Thanks for the suggestion!

CharlesGueunet:
in fact, the offset array should never be `nbPoints +1`, even for PolyData. Maybe using the `vtkCellArray::InsertNextCell` may remove the need of handling the connectivity and offests manually.

pierre-guillou:
> I think in this case, we would need to have a size of 1+1 for the offset array instead of nbPoints + 1

Do you mean doing 
```cpp
  offsets->SetNumberOfTuples(2);
  offsets->SetTuple1(0, 0);
  offsets->SetTuple1(1, 1);
```
for UnstructuredGrids? I tried with FTMTree (SkeletonNodes is a VTU) but only one point can be seen. Same for PolyData.

> in fact, the offset array should never be nbPoints +1, even for PolyData. Maybe using the vtkCellArray::InsertNextCell may remove the need of handling the connectivity and offests manually.

The `offset`/`connectivity` scheme is used in a variety of other TTK filters (DiscreteGradient, MorseSmaleComplex, RipsComplex, PersistenceDiagram), mostly in order to allocate once before the main loop. The loop can thus be made parallel.

CharlesGueunet:
I would suggest to let the `vtkCellArray` handle the allocation / insertion. For example:
```cxx
  vtkNew<vtkCellArray> cells;
  cells->InsertNextCell(ntPts);
  // insert each ids (loop)
    cells->InsertCellPoint(0);
    cells->InsertCellPoint(1);
    // ...
```

The offset / connectivity will be handled internally by the `vtkCellArray`, there is no need to explicitly create them here as local variables.

CharlesGueunet (approved):

pierre-guillou:
Thanks for your comments!

julien-tierny:
go!

PR #750, opened 2022-04-11
Optional preconditions for implicit/periodic triangulations

### LINKED ISSUE #749: Memory Footprint of Implicit Triangulations
Hi everyone,
the memory footprint of the implicit triangulation has increased significantly over the last year. Consider the following test:
1. Create a Wavelet source with a resolution of 256^3. This creates a vtkImageData object holding roughly 64 MB float data.
2. Switch to a SpreadSheet view in order to omit memory consumption caused by rendering.
3. Running now the ttkHelloWorld filter (which only requires vertex neighbors) now increases PV's memory consumption by 3GB! Note, turning the vtkImageData object into a triangle mesh with vtkTetrahedralize increases the memory consumption by 4GB. So the implicit triangulation uses almost as much memory as explicitly representing the input object in VTK.
4. An additional problem is that memory is no longer freed after the corresponding vtkDataObject was destroyed. I do not know at which point this feature broke. Maybe this is related to the newer versions of VTK/ParaView.

I'm not able to run this experiment with a 512^3 volume on my machine with 18 GB ram, which should be more than enough for processing 0.5 GB of data.

In short, the implicit triangulation data structure performs too aggressive caching. This negates the primary advantage of implicit representations: the low memory footprint. I do not even know if this caching really brings a noticeable speedup that justifies such a significant memory hit.

I suggest to solve the issue by reverting the implicit triangulation to what the name suggests and either provide the cached versions as alternative triangulation data structures (e.g., ttkImplicitTriangulationCached) or toggle caching at compile time. What are your thoughts @julien-tierny, @CharlesGueunet, @chgarth, @pierre-guillou?

Best
Jonas

pierre-guillou:
Should fix #749.

This PR introduce optional preconditions for the implicit and periodic triangulations. This is done by subclassing the `ImplicitTriangulation` (resp. `PeriodicImplicitTriangulation`) classes with two new classes: `ImplicitWithPreconditions` and `ImplicitNoPreconditions` (resp. `PeriodicWithPreconditions` and `PeriodicNoPreconditions`). The `WithPreconditions` classes retain the current behavior of the triangulations by filling some internal buffers at preconditioning time. On the contrary, the `NoPreconditions` classes computes these values on the fly (involving a -25/-30% performance penalty). The parent classes have been made abstract.

To switch between the two modes, a new filter, `TriangulationManager`, is introduced by also merging `PeriodicGrid` and `CompactTriangulationPreconditioning`.

To avoid the automatic precondition of large datasets, a strategy system has been designed. Three strategies can be selected by the user using `TriangulationManager`:
* Default: use preconditioning under a grid size of TTK_IMPLICIT_PRECONDITIONS_THRESHOLD^3,
* Preconditioning: force preconditioning for faster computation,
* No Preconditioning: disable preconditioning to reduce memory pressure.

The TTK_IMPLICIT_PRECONDITIONS_THRESHOLD is both a pre-processor define and a CMake (advanced) option with a default value of 256, meaning that by default grids with dimensions larger than 256x256x256 won't be preconditioned by default but smaller grids will.

The `PeriodicTriangulation` is also not preconditioned by default alongside the `ImplicitTriangulation` but only when switching the triangulation type.

To reduce the memory footprint of the CI (in particular for the `ctBones` data-set), the TTK_IMPLICIT_PRECONDITIONS_THRESHOLD has been lowered to 64 in the relevant GitHub Actions workflow.

Enjoy,
Pierre


julien-tierny:
great stuff!

julien-tierny:
rock'n'roll

JonasLukasczyk:
Thank you @pierre-guillou! Nice work. Just one question: for the threshold do you test the xyz-dimension OR the total number of vertices? The latter would make more sense to me.

Best

pierre-guillou:
The threshold value is the number of vertices on a cubic grid edge. This threshold is cubed then compared to the product of the `dataset->GetDimensions()` result / `AbstractTriangulation::gridDimensions_` member variable (Triangulation.cpp:159). The default threshold value is 256, meaning that a grid larger than 256x256x256 vertices is not preconditioned (by default). 

I don't know what you mean with the xyz-dimension though. What's differentiate it from the number of vertices?

JonasLukasczyk:
I was just saying that I would expect that the test simply compares `dimX x dimY x dimZ < threshold` instead of `dimX<threshold` or something. I just wanted to make sure that a grid of size 100x1024x1024 does not fall though the cracks.

JonasLukasczyk:
We should be aware that if someone performs benchmarks and does not know about this threshold then this will yield a discontinuity in a scaling study, i.e., when suddenly the performance increases although the size increases (since the hybrid kicks in).

julien-tierny:
@JonasLukasczyk 
thanks for your feedback.
there's a fairly explicit warning message in the console when a large grid is detected.

PR #751, opened 2022-04-11
TriangulationManager: Fix printed messages

pierre-guillou:
Following #750, this PR fixes a few logic errors leading to incorrect messages displayed to the user:
* in TriangulationManager, the "Switched regular grid preconditions" message is now displayed iff the preconditions state has changed,
* in Triangulation, the default strategy warning on large grids should appear on triangulation creation (by manipulating the `debugLevel_` member variable)
* the condition surrounding this message has been fixed
* the message is slightly different in 2D and 1D: the number of vertices at the threshold is displayed, instead of the 3D grid dimensions.

Sorry for this,
Pierre

### DIFF
--- a/core/base/triangulation/Triangulation.cpp
+++ b/core/base/triangulation/Triangulation.cpp
@@ -158,11 +158,29 @@ bool Triangulation::processImplicitStrategy(const STRATEGY strategy) const {
     // disable preconditioning above TTK_IMPLICIT_PRECONDITIONS_THRESHOLD
     const auto doPreconditioning = nVerts <= threshold * threshold * threshold;
 
-    if(!this->hasImplicitPreconditions()) {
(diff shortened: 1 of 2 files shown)


julien-tierny:
thanks!

JonasLukasczyk:
I think the issue is still not fixed. Especially the error msg is misleading because you always assume that the input grid is cubical. Imagine the following case: input grid has resolution of `100 x 5k x 5k` then TTK says: `large input grid detected 256x256x256` (with default threshold). I think this needs to be reformulated in terms of total number of vertices.

pierre-guillou:
@JonasLukasczyk This should be fixed with #759. We now use the global number of vertices in the grid for the threshold. You might want to reset the `TTK_IMPLICIT_PRECONDITIONS_THRESHOLD` CMake key to use the new value (`256*256*256` instead of `256`).

PR #752, opened 2022-04-11
New SurfaceGeometrySmoother filter

pierre-guillou:
This PR adds a new filter, `SurfaceGeometrySmoother` that works a bit like `GeometrySmoother` but takes a triangulated surface as a second input and projects its first input onto it. This is a bit similar to what `QuadrangulationSubdivision` does internally after the subdivision step (in fact, `SurfaceGeometrySmoother` borrows a lot from it - and a potential future `QuadrangulationSubdivision` PR should take advantage of this new filter).

The first input is expected to be a triangulation of dimensions 1 or 2. Similarly to `GeometrySmoother`, a mask field is supported. The filter also supports an optional `ttkVertexScalarField` point data array on the first input. If absent, for each point to smooth, the nearest vertex on the surface is computed using the euclidean distance. This helps supporting a broader range of input datasets (Morse-Smale 1-separatrices for instance).

To improve performance, a majority of heap allocations inside `Geometry` have been replaced with stack allocations (`std::vector` to `std::array`). This should have an impact also on `FiberSurface` and `QuadrangulationSubdivision`.

Additionally, `GeometrySmoother` has been updated with a new GUI option that disable the use of the mask field. Previously, a dataset with a `ttkMaskScalarField` was always using it.

Enjoy,
Pierre

### DIFF
--- a/core/base/fiberSurface/FiberSurface.cpp
+++ b/core/base/fiberSurface/FiberSurface.cpp
@@ -143,7 +143,7 @@ int FiberSurface::computeTriangleFiber(
   }
 
   // compute the interpolations
-  vector<double> baryCentrics0, baryCentrics1;
+  std::array<double, 2> baryCentrics0{}, baryCentrics1{};
   vector<double> p(2), p0(2), p1(2), p2(2);
 
   p[0] = intersection.first;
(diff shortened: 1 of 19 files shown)


julien-tierny:
looks good, thanks!

PR #753, opened 2022-04-14
Point set to surface

MatPont:
This PR adds the PointSetToSurface filter to TTK.
It takes a point set as its only input and produces a vtkUnstructuredGrid as output corresponding to a surface made by "linking" points together given two arrays (corresponding to the X and Y values of a point in the grid of the resulting surface).
It can be seen as an extension of the PointSetToCurve filter to two dimensions (PointSetToSurface is very inspired by PointSetToCurve).

### DIFF
--- a/core/vtk/ttkPointSetToSurface/ttkPointSetToSurface.cpp
+++ b/core/vtk/ttkPointSetToSurface/ttkPointSetToSurface.cpp
@@ -1,19 +1,10 @@
 #include <ttkPointSetToSurface.h>
 
 #include <vtkCellData.h>
-#include <vtkPointData.h>
-#include <vtkPointSet.h>
 #include <vtkUnstructuredGrid.h>
-
 #include <vtkInformation.h>
-#include <vtkInformationVector.h>
 
-#include <ttkMacros.h>
 #include <ttkUtils.h>
-
-#include <array>
-#include <map>
-#include <set>
 
 vtkStandardNewMacro(ttkPointSetToSurface);
 
--- a/core/vtk/ttkPointSetToSurface/ttkPointSetToSurface.cpp
+++ b/core/vtk/ttkPointSetToSurface/ttkPointSetToSurface.cpp
@@ -2,7 +2,7 @@
 
 #include <vtkCellData.h>
 #include <vtkInformation.h>
-#include <vtkUnstructuredGrid.h>
+#include <vtkPolyData.h>
 
 #include <ttkUtils.h>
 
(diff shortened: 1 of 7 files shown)


core/vtk/ttkPointSetToSurface/ttkPointSetToSurface.cpp
```diff
@@ -0,0 +1,171 @@
...
+    info->Set(vtkDataObject::DATA_TYPE_NAME(), "vtkPointSet");
+    return 1;
+  }
+  return 0;
+}
+
+int ttkPointSetToSurface::FillOutputPortInformation(int port,
+                                                    vtkInformation *info) {
+  if(port == 0) {
+    info->Set(vtkDataObject::DATA_TYPE_NAME(), "vtkUnstructuredGrid");
```
julien-tierny: I believe a `vtkPolyData` output would be more appropriate here (better to model 0, 1 and 2D).

core/vtk/ttkPointSetToSurface/ttkPointSetToSurface.h
```diff
@@ -0,0 +1,45 @@
+/// \class ttkPointSetToSurface
+/// \ingroup vtk
+/// \author Mathieu Pont <mathieu.pont@lip6.fr>
+/// \date April 2022
```
julien-tierny, line 4: April fools!

paraview/xmls/PointSetToSurface.xml
```diff
@@ -0,0 +1,86 @@
...
+        </Documentation>
+      </StringVectorProperty>
+
+      <PropertyGroup panel_widget="Line" label="Input options">
+        <Property name="InputOrderingXArray" />
+        <Property name="InputOrderingYArray" />
+      </PropertyGroup>
+
+      <Hints>
+        <ShowInMenu category="TTK - Misc" />
```
julien-tierny: "TTK - Domain"
(thanks for moving PointSetToCurve as well)

julien-tierny (changes requested):
that looks great, could you take care of the two minor things I reported?
thanks!

core/vtk/ttkPointSetToSurface/ttkPointSetToSurface.cpp
MatPont: Done in the next commit!

paraview/xmls/PointSetToSurface.xml
MatPont: Done in the next commit!

julien-tierny (approved):

julien-tierny:
alright, thanks!

PR #754, opened 2022-04-14
IntegralLines: Don't convert point data to double

pierre-guillou:
This PR fixes & cleans the VTK layer of the `IntegralLines` filter. The output scalar point data arrays are no longer forcibly converted to `vtkDoubleArray` but retain the types of the corresponding input scalar arrays.

Enjoy,
Pierre


julien-tierny:
alright, looks good. thanks!

PR #755, opened 2022-04-14
Changed compile flags for Apple Silicon

m-s-will:
This PR adds a check if we are on Apple Silicon on which the '-march=native' flag is currently not available. Alternatively, '-mcpu=apple-m1' is available, which allows for some optimization.


julien-tierny:
looks good, thanks!

PR #756, opened 2022-04-14
MergeTreeDistanceMatrix better memory usage

MatPont:
This PR modifies the MergeTreeDistanceMatrix filter in order to have a better memory usage. 

Here after, a table showing the maximum memory usage (RES column of htop) for different datasets before and after the PR:
```
                   |  N |  BEFORE |  AFTER |
Dark Sky           | 40 | 12.5 Gb | 2.3 Gb |
Sea Surface Height | 48 | 12.1 Gb | 2.5 Gb |
Isabel             | 12 |  2.2 Gb | 1.3 Gb |
```

The problem was that before this PR the trees were copied among all threads instead of being shared! Moreover there are slight changes in MergeTreeDistance to better manage the option "SaveTree" (i.e. not modifying the original trees, important for the MergeTreeDistanceMatrix filter, especially if the trees are shared).

Some messages have also been added, such as the value of the parameters BranchDecomposition, NormalizedWasserstein and KeepSubtree at the beginning of the execution to avoid any ambiguities about what is executed and MergeTreeDistanceMatrix shows less messages.

### DIFF
--- a/core/base/mergeTreeClustering/MergeTreeDistance.h
+++ b/core/base/mergeTreeClustering/MergeTreeDistance.h
@@ -334,9 +334,6 @@ namespace ttk {
       Memory m;
       Timer t_total;
 
-      ftm::FTMTree_MT *tree1 = &(mTree1.tree);
-      ftm::FTMTree_MT *tree2 = &(mTree2.tree);
-
       // ---------------------
       // ----- Testing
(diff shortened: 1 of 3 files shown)


julien-tierny:
wonderful, thanks!

PR #758, opened 2022-04-15
Fix/Enforce some clang-tidy modernize checks

### LINKED ISSUE #732: Virtual Destructors
Many base classes have a non-virtual destructor, but the vtk layer inherits from the base classes. Since almost always we use the default destructor this did not raise any issues yet, but in the future we should make this more consistent and add the `virtual` qualifier to all base and even vtk class destructors.

pierre-guillou:
This PR enables, enforces & fixes several `modernize` clang-tidy checks. Among them:
* replace `NULL` with `nullptr`,
* replace `typedef` with `using`,
* replace `push_back` with `emplace_back` (when a copy can be avoided),
* replace C standard library includes with their C++ counterparts (they reside in the `std::` namespace),
* remove void arguments in functions (this is useless in C++),
* use bool literals instead of integers,
* use `= default` for relevant constructors & destructors
* use `override` on methods instead of `virtual` when there is a virtual parent class method with the same signature. Also adds `override` when missing. I believe this fixes #732.

All of this has been automatically fixed using `clang-tidy -fix` and should also be enforced by the GitHub Actions CI.

Additionally, header guards were replaced with `#pragma once` directives. This is shorter, might have dedicated code paths inside compilers and don't shadow other `#define`s. I did not find a way to enforce that (yet).

Enjoy,
Pierre

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -11,6 +11,7 @@ Checks: >-
   misc-*,
   -misc-non-private-member-variables-in-classes,
   -misc-no-recursion,
+  modernize-redundant-void-arg,
   modernize-use-bool-literals,
   modernize-use-emplace,
   modernize-use-equals-default,
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -11,6 +11,7 @@ Checks: >-
   misc-*,
   -misc-non-private-member-variables-in-classes,
   -misc-no-recursion,
+  modernize-use-bool-literals,
   modernize-use-emplace,
   modernize-use-equals-default,
   modernize-use-nullptr,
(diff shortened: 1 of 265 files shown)


julien-tierny:
looks great! (didn't check completely though)

the CI reports something odd for the python scripts (see the end of the 20.04 test):
4c4
<     "1manifoldLearning_OutputArc": "75cc5d095c26cf3d81db977317919e3f9477c2d8",
---
>     "1manifoldLearning_OutputArc": "6819ee58cb00ef281b327da8d48d6b41d2d27416",
Error: Process completed with exit code 1.
does it mean the outputs differed here?

pierre-guillou:
Actually, the `1manifoldLearning_OutputArc` is the only instance in the Python CI when an output can randomly switch between two (?) different states, leading to a discrepancy in the hashes. This happens around 10% of the time and I haven't yet been able to track this. I will focus on it in the near future.

julien-tierny:
alright, thanks for the info!

PR #759, opened 2022-04-19
Triangulation: Switch implicit preconditions threshold to number of vertices

pierre-guillou:
This PR switches from an implicit preconditions threshold defined with a grid edge dimension to a threshold defined by the total number of vertices in the implicit grid.

The default threshold value has been changed from `256` to `256*256*256`.

Enjoy,
Pierre


julien-tierny:
awesome, thanks!

PR #760, opened 2022-04-19
SurfaceGeometrySmoother: Support PolyData surfaces, hide error message

pierre-guillou:
This PR updates the SurfaceGeometrySmoother filter by allowing the Surface input to be a `vtkPointSet` and not only a `vtkUnstructuredGrid`.

A new boolean parameter in the `ttkAlgorithm::GetIdentifierArrayPtr` now controls the display of the error log message. Since SurfaceGeometrySmoother can work even when `ttkVertexScalarField` is absent, a warning message is displayed in the filter instead.

Enjoy,
Pierre


julien-tierny:
awesome, thanks!

PR #761, opened 2022-04-19
added GhostCellpreprocessing and updated ArrayPreconditioning

m-s-will:
This PR adds a GhostCellPreprocessing Filter which uses vtkGhostType generated by a GhostCellsGenerator to calculate which rank is the "primary" owner of each vertex and denotes this in the new RankArray. It does so by first simply writing their own rank into the RankArray for vertices where the vtkGhostType is zero (this rank is the primary owner) and afterwards using MPI to request the owner ranks for their ghost cells. Currently broadcasts are used, this can be optimized by calculating a rank neighborhood graph beforehand and only requesting from their neighbors.

Also updates ArrayPreconditioning to correctly calculate an order field over multiple ranks. It does so by using distributed sorting over all non-ghost cells, rank 0 merging the sorted arrays from each rank and sending the correct order back to the ranks. Afterwards, the order information for ghost cells can be requested from the owning rank by using the RankArray.

The following trace shows the complete workflow, on a pvserver run with mpirun -n 4:

```
# trace generated using paraview version 5.10.0
#import paraview
#paraview.compatibility.major = 5
#paraview.compatibility.minor = 10

from paraview.simple import *
#### disable automatic camera reset on 'Show'
paraview.simple._DisableFirstRenderCameraReset()

# create a new 'Wavelet'

UpdatePipeline(time=0.0, proxy=wavelet1)

# create a new 'Generate Global Ids'
generateGlobalIds1 = GenerateGlobalIds(registrationName='GenerateGlobalIds1', Input=wavelet1)

UpdatePipeline(time=0.0, proxy=generateGlobalIds1)

# create a new 'Ghost Cells Generator'
ghostCellsGenerator1 = GhostCellsGenerator(registrationName='GhostCellsGenerator1', Input=generateGlobalIds1)

UpdatePipeline(time=0.0, proxy=ghostCellsGenerator1)

# create a new 'TTK GhostCellPreprocessing'
tTKGhostCellPreprocessing1 = TTKGhostCellPreprocessing(registrationName='TTKGhostCellPreprocessing1', Input=ghostCellsGenerator1)

UpdatePipeline(time=0.0, proxy=tTKGhostCellPreprocessing1)

# create a new 'TTK ArrayPreconditioning'
tTKArrayPreconditioning1 = TTKArrayPreconditioning(registrationName='TTKArrayPreconditioning1', Input=tTKGhostCellPreprocessing1)

UpdatePipeline(time=0.0, proxy=tTKArrayPreconditioning1)
```


core/vtk/ttkGhostCellPreprocessing/ttkGhostCellPreprocessing.h
```diff
@@ -0,0 +1,54 @@
+/// \ingroup vtk
+/// \class ttkGhostCellPreprocessing
+/// \author Pierre Guillou <pierre.guillou@lip6.fr>
```
julien-tierny: please use your name and email address.

julien-tierny (changes requested):
hi michael,
thanks a lot for the PR.
can you please take into account the above (minor) remark about the author info in the header?
also, thanks for fixing the CI errors. these are tiny things, we'd better take care of them now than later.
thanks!

core/vtk/ttkGhostCellPreprocessing/ttkGhostCellPreprocessing.h
m-s-will: done, for GhostCellPreprocessing. I left ArrayPreconditioning at the original author

m-s-will:
Some of the checks are failing because apparently mpi is not found in the test environments

julien-tierny:
@pierre-guillou could you have a look into this please?
thanks!

pierre-guillou:
Well, MPI is an an optional TTK dependency and this PR is the first one that makes something of it. As a consequence, the CI does not have MPI installed (yet) and any unguarded `#include <mpi.h>` in TTK will raise an error (every source file is tested with `clang-check`). 

I would suggest to wrap any MPI-dependent code under `TTK_ENABLE_MPI` macro guards, similarly to what is done with OpenMP. For an example of handling an optional dependency in TTK code, see core/base/dimensionReduction/DimensionReduction.cpp with `TTK_ENABLE_SCIKIT_LEARN` or core/base/harmonicField/HarmonicField.cpp with `TTK_ENABLE_EIGEN`.

julien-tierny:
Agreed 100% for `TTK_ENABLE_MPI`.
That being said, I believe the CI needs to be upgraded as well to include openmpi as a dependency.
Could you take care of that please?

(Maybe a new test action could be created with its support ON)

pierre-guillou:
I can extend the matrix in the `check_code`/`check-warnings` job to test with and without `TTK_ENABLE_MPI` but I think a full rebuild of `ttk-paraview` with MPI will be needed to fully take advantage of MPI.

julien-tierny:
could you PR ttk-paraview to include actions supporting the build of a package of ttk-paraview with the MPI support builtin?
(also could you PR ttk to upgrade its patches?)
thanks!

JonasLukasczyk:
In contrast to OpenMP modules that use MPI will not work without it since there is no fallback. So instead of guarding every expression, I would put one guard around the source files. What do you think?

JonasLukasczyk:
Modules that only have a MPI extension (such as critical points and so on) of course should use guards similar to OpenMP.

m-s-will:
Yes for some modules such as computing the rankarray, using it by itself without MPI does not make much sense, while some modules such as the ArrayPreconditioning for the order field simply have an alternate mode when running it with MPI there the guards around codeblocks makes sense.

core/base/common/OrderDisambiguation.h
```diff
@@ -1,12 +1,204 @@
...
+using IT = long long int;
 namespace ttk {
 
+  struct value {
+    float scalar;
+    IT globalId;
+    IT localId;
+    IT ordering = 0;
+
+    value(float _scalar, IT _globalId, IT _localId)
```
julien-tierny: I am a bit confused by the presence of this `float` here.
What is that supposed to encode?
Some input scalar data? If so, the data type should be a template parameter (like it is done throughout TTK)

core/base/common/OrderDisambiguation.h
```diff
@@ -1,12 +1,204 @@
...
 
 #include <BaseClass.h>
+#include <iostream>
 
 #include <algorithm>
+#include <unordered_map>
+#include <unordered_set>
 #include <vector>
 
+using IT = long long int;
```
julien-tierny: is IT used to represent indices?
If so I believe `idType` would be more appropriate (cf non-distributed implementation)

core/base/common/OrderDisambiguation.h
```diff
@@ -1,12 +1,204 @@
...
+  // ( so only vertices which are no ghost cells)
+  inline std::
+    tuple<std::vector<value>, std::vector<IT>, std::unordered_map<IT, IT>>
+    populateVector(const size_t nVerts,
+                   const float *const scalars,
+                   const IT *const globalIds,
+                   const char *const ghostCells) {
+    std::vector<value> valuesToSortVector;
+    std::vector<IT> gidsToGetVector;
+    std::unordered_map<IT, IT> gidToLidMap;
```
julien-tierny: this map is used to store the correspondence from global Ids to local Ids, right?
if so, that information would typically be needed by nearly all algorithms.
maybe it would be relevant to move its construction to `ttkMpiUtils`

core/base/common/OrderDisambiguation.h
```diff
@@ -1,12 +1,204 @@
...
+      outVector.assign(values.end() - burstSize, values.end());
+      values.erase(values.end() - burstSize, values.end());
+    }
+
+    return outVector;
+  }
+
+  // takes in an ordered (as defined above) vector of values and creates an
+  // ordermap for each scalar value
+  inline std::unordered_map<float, IT> buildOrderMap(std::vector<value> &values,
```
julien-tierny: I don't understand why such a function would be needed (I haven't read the other classes yet).
Also, as discussed above, the usage of `float` seems improper.

core/base/common/OrderDisambiguation.h
```diff
@@ -1,12 +1,204 @@
...
+   *
+   * @param[in] nVerts number of vertices
+   * @param[in] scalars array of size nVerts, the scalar values which we want to
+   * order
+   * @param[in] orderMap map which maps scalar values to a defined order
+   * @param[out] order array of size nVerts, computed order of vertices
+   * @param[in] nThreads number of parallel threads
+   */
+  inline void buildOrderArray(const size_t nVerts,
+                              const float *const scalars,
```
julien-tierny: `float` seems wrong here.

core/base/common/OrderDisambiguation.h
```diff
@@ -1,12 +1,204 @@
...
+                              std::unordered_map<float, IT> &orderMap,
```
julien-tierny: same as above

core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
```diff
@@ -43,6 +44,27 @@ int ttkArrayPreconditioning::FillOutputPortInformation(int port,
...
+  int intTag,
+  std::vector<std::vector<ttk::value>> &unsortedReceivedValues) {
+  std::vector<ttk::value> receivedValues;
+  // be prepared to receive burstsize of elements, resize after receiving to the
+  // correct size
+  int amount;
+  MPI_Recv(
+    &amount, 1, MPI_INT, rankFrom, intTag, MPI_COMM_WORLD, MPI_STATUS_IGNORE);
+  receivedValues.resize(amount, {0, 0, 0});
+  // this->printMsg("Receiving " + std::to_string(amount) + " values from rank "
```
julien-tierny: please unnecessary comments.

core/base/common/OrderDisambiguation.h
julien-tierny: Also, the general philosophy in TTK is to avoid the copy of data arrays. Here I am not sure the storage of the actual scalar value is necessary.

core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
```diff
@@ -84,6 +106,365 @@ int ttkArrayPreconditioning::RequestData(vtkInformation *ttkNotUsed(request),
...
+            }
+
+            // move the struct from the unsortedReceivedValues subvector to the
+            // finalValues vector to get an ordering
+            ttk::value currentValue
+              = unsortedReceivedValues[rankIdOfMaxScalar].back();
+            // we send the globalId and the the order via one send command,
+            // therefore we need to check two concurrent values at the same time
+            // later on
+            // this->printMsg("Current GlobalId is " +
```
julien-tierny: please remove commented code (in other places as well). thanks

paraview/xmls/GhostCellPreprocessing.xml
```diff
@@ -0,0 +1,42 @@
...
+      </InputProperty>
+
+      
+
+      ${DEBUG_WIDGETS}
+
+
+
+      <Hints>
+        <ShowInMenu category="TTK - Scalar Data" />
```
julien-tierny: "TTK - Domain"

core/vtk/ttkGhostCellPreprocessing/ttkGhostCellPreprocessing.cpp
```diff
@@ -0,0 +1,184 @@
...
+    while(i < allUnknownIds[rank].size()) {
+      std::vector<IT> receivedGlobals;
+      receivedGlobals.resize(allUnknownIds[rank].size());
+      MPI_Status status;
+      int amount;
+      MPI_Recv(receivedGlobals.data(), allUnknownIds[rank].size(), MIT,
+               MPI_ANY_SOURCE, MPI_ANY_TAG, MPI_COMM_WORLD, &status);
+      int sourceRank = status.MPI_SOURCE;
+      MPI_Get_count(&status, MIT, &amount);
+      // this->printMsg("Rank " + std::to_string(rank) + " got " +
```
julien-tierny: thanks for removing all commented code.

julien-tierny (changes requested):
I made a first round of code review.
I have some concerns about the data structures in `core/base/common/OrderDisambiguation.h`, which I believe should be revised. There's a number of ttk coding-usages we should be included in this PR (data type templating, avoiding data copy etc.)

@JonasLukasczyk can you please review the code too please?
@pierre-guillou can you please enter a review as well?
thanks!

core/base/common/OrderDisambiguation.h
m-s-will: You are right, `float` is used for the input data. Initially ttkMacros where used for templatizing, however we ran into problems. I'm sure we can fix them for the main repo and use templatized typing.
The scalar values need to get sent from all ranks to rank 0 which merges their sorted subsets of the vertices, therefore I don't think we can get rid of the storage, however ideas would be appreciated!

julien-tierny:
hi @JonasLukasczyk & @pierre-guillou ,

this is a friendly reminder: could you review this PR really quick too?

thanks

core/base/arrayPreconditioning/ArrayPreconditioning.h
```diff
@@ -0,0 +1,602 @@
...
+  /**
+   * The ArrayPreconditioning class provides methods to generate order arrays
+   * from a selection of scalar field arrays.
+   */
+  class ArrayPreconditioning : virtual public Debug {
+
+  public:
+    ArrayPreconditioning();
+#ifdef TTK_ENABLE_MPI
+    MPI_Datatype getMPIType(float val) const {
```
pierre-guillou: Here you can use the `ttkNotUsed` macro instead of `TTK_FORCE_USE` to compact more the code:
```cpp
MPI_Datatype getMPIType(const float ttkNotUsed(val)) const {
  return MPI_FLOAT;
}
```
(`TTK_FORCE_USE` is rather used for parameters that are used within `#ifdef`s).

core/base/common/OrderDisambiguation.h
```diff
@@ -1,12 +1,13 @@
 #pragma once
 
 #include <BaseClass.h>
+#include <iostream>
```
pierre-guillou: It seems that the changes in this file can be safely reverted

core/vtk/ttkGhostCellPreprocessing/ttkGhostCellPreprocessing.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include <vtkCommand.h>
+#include <vtkDataArray.h>
+#include <vtkDataSet.h>
+#include <vtkIdTypeArray.h>
+#include <vtkInformation.h>
+#include <vtkIntArray.h>
+#include <vtkPointData.h>
+
+vtkStandardNewMacro(ttkGhostCellPreprocessing);
+using IT = long long int;
```
pierre-guillou: Use `LongSimplexId` from base/common/DataTypes.h` instead?

pierre-guillou (commented):
I did a quick pass on the code, everything seems good (apart from a few comments about code style).

core/base/arrayPreconditioning/ArrayPreconditioning.h
```diff
@@ -0,0 +1,593 @@
...
+                        const size_t nVerts,
+                        const DT *const scalars,
+                        const IT *const globalIds,
+                        const char *const ghostCells) const {
+      for(size_t i = 0; i < nVerts; i++) {
+        IT globalId = globalIds[i];
+        IT localId = i;
+        gidToLidMap[globalId] = localId;
+        if((int)ghostCells[i] == 0) {
+          float scalarValue = scalars[i];
```
julien-tierny: `float` --› DT

core/base/arrayPreconditioning/ArrayPreconditioning.h
```diff
@@ -0,0 +1,593 @@
...
+            } else {
+              this->ReceiveAndAddToVector<DT, IT>(
+                mpi_values, i, structTag, intTag, unsortedReceivedValues);
+            }
+          }
+
+          while(finalValues.size() < totalSize) {
+            // take the current maximum scalar over all ranks
+            int rankIdOfMaxScalar = -1;
+            float maxScalar = -FLT_MAX;
```
julien-tierny: `float` -› `DT`

core/base/arrayPreconditioning/ArrayPreconditioning.h
```diff
@@ -0,0 +1,593 @@
...
+            IT maxGId = -1;
+            for(int i = 0; i < numProcs; i++) {
+              if(unsortedReceivedValues[i].size() > 0) {
+                int thisId = i;
+                float thisScalar = unsortedReceivedValues[i].back().scalar;
```
julien-tierny: `float` -› `DT`

core/vtk/ttkGhostCellPreprocessing/ttkGhostCellPreprocessing.cpp
```diff
@@ -0,0 +1,176 @@
...
+        gIdSet.insert(globalId);
+      } else {
+        // otherwise the vertex belongs to another rank and we need to find out
+        // to which one this needs to be done by broadcasting the global id and
+        // hoping for some other rank to answer
+        currentRankUnknownIds.push_back(globalId);
+        gIdToLocalMap[globalId] = i;
+      }
+    }
+    // this->printMsg("Rank " + std::to_string(rank) + " done with local
```
julien-tierny: please remove unnecessary comments.

core/base/arrayPreconditioning/ArrayPreconditioning.h
```diff
@@ -0,0 +1,593 @@
...
+  /**
+   * The ArrayPreconditioning class provides methods to generate order arrays
+   * from a selection of scalar field arrays.
+   */
+  class ArrayPreconditioning : virtual public Debug {
+
+  public:
+    ArrayPreconditioning();
+#ifdef TTK_ENABLE_MPI
+    MPI_Datatype getMPIType(const float ttkNotUsed(val)) const {
```
julien-tierny: could you please move these functions to `core/base/common/MPI.h`?

core/base/arrayPreconditioning/ArrayPreconditioning.h
```diff
@@ -0,0 +1,593 @@
...
+      }
+    };
+
+    // creates a vector of value structs based on the given pointers
+    // only takes values of vertices which mainly belong to the current rank
+    // ( so only vertices which are no ghost cells)
+    template <typename DT, typename IT>
+    void populateVector(std::vector<value<DT, IT>> &valuesToSortVector,
+                        std::vector<IT> &gidsToGetVector,
+                        std::unordered_map<IT, IT> &gidToLidMap,
```
julien-tierny, line 50: NOTE for later: this variable (`gidToLidMap`) should be a member of all the classes deriving from the `BaseClass` (if MPI is detected and enabled).

core/base/arrayPreconditioning/ArrayPreconditioning.h
```diff
@@ -0,0 +1,593 @@
...
+#ifdef TTK_ENABLE_OPENMP
+#pragma omp parallel for num_threads(nThreads)
+#endif // TTK_ENABLE_OPENMP
+      for(size_t i = 0; i < nInts; i += 2) {
+        order[gidToLidMap[orderedValuesForRank[i]]]
+          = orderedValuesForRank[i + 1];
+      }
+    }
+
+    // gets all the neighbors of a rank based on the rankarray of the vertices
```
julien-tierny, line 128: NOTE for later: that function should be a member of all the classes deriving from the BaseClass (if MPI is detected and enabled).

core/base/arrayPreconditioning/ArrayPreconditioning.h
```diff
@@ -0,0 +1,593 @@
...
+        int r = rankArray[i];
+        if(r != ownRank)
+          neighbors.emplace(r);
+      }
+    }
+
+    // returns a vector of gid vectors, one vector for each neighbor
+    // this shows us where we need to get the order for these gids from
+    template <typename IT>
+    void getGIdForNeighbors(std::vector<std::vector<IT>> &outVector,
```
julien-tierny, line 144: NOTE for later: that function should be a member of all the classes deriving from the BaseClass (if MPI is detected and enabled).

core/base/arrayPreconditioning/ArrayPreconditioning.h
```diff
@@ -0,0 +1,593 @@
...
+        } else {
+          gidsToGetVector.push_back(globalId);
+        }
+      }
+    }
+
+    // orders an value vector first by their scalar value and then by global id
+    template <typename DT, typename IT>
+    void sortVerticesDistributed(std::vector<value<DT, IT>> &values) const {
+      std::sort(
```
julien-tierny: this is a sequential sort. 
please use OpenMp if available (see `core/base/common/OrderDisambiguation.h`:78, function `sortVertices()`)

julien-tierny: the number of threads per node should then be set by the user with the env var `OMP_NUM_THREADS`

core/vtk/ttkGhostCellPreprocessing/ttkGhostCellPreprocessing.h
```diff
@@ -0,0 +1,54 @@
...
+#pragma once
+
+// VTK Module
+#include <ttkGhostCellPreprocessingModule.h>
+
+// VTK Includes
+#include <ttkAlgorithm.h>
+
+#ifdef TTK_ENABLE_MPI
+#include <mpi.h>
```
julien-tierny: please move this include to `core/base/common/MPI.h`

core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
```diff
@@ -84,6 +83,50 @@ int ttkArrayPreconditioning::RequestData(vtkInformation *ttkNotUsed(request),
     }
   }
 
+  auto vtkGlobalPointIds = pointData->GetGlobalIds();
+  auto vtkGhostCells = pointData->GetArray("vtkGhostType");
+  auto rankArray = pointData->GetArray("RankArray");
+  if(vtkGlobalPointIds != nullptr && vtkGhostCells != nullptr
+     && rankArray != nullptr) {
+#ifdef TTK_ENABLE_MPI
```
julien-tierny, line 91: checking if ttk has been built with MPI is not sufficient to trigger some MPI-specific "behavior".
you need to check that the paraview server has also been executed with `mpirun`.
I believe this can be done by retrieving the MPI communicator from `vtkMPIController` (to check with @CharlesGueunet )

Generally speaking, we need to have to preprocessor check first (#ifdef TTK_ENABLE_MPI), but then a dynamic piece of code (e.g. "isRunningWithMPI()") which checks if the above communicator is `nullptr` or not.

In the present case, if the code is not run with `mpirun`, then simply default back to the original array preconditionning (based on `core/base/common/DisambiguateOrder.h`)

julien-tierny: the function `isRunningWithMPI()` could be added to `ttkMPIUtils`.
thanks!

core/vtk/ttkGhostCellPreprocessing/ttkGhostCellPreprocessing.cpp
```diff
@@ -0,0 +1,176 @@
...
+  output->ShallowCopy(input);
+
+  auto pointData = input->GetPointData();
+  ttk::SimplexId nVertices = input->GetNumberOfPoints();
+  this->printMsg("#Points: " + std::to_string(nVertices));
+
+  auto vtkGlobalPointIds = pointData->GetGlobalIds();
+  auto vtkGhostCells = pointData->GetArray("vtkGhostType");
+  if(vtkGlobalPointIds != nullptr && vtkGhostCells != nullptr) {
+#ifdef TTK_ENABLE_MPI
```
julien-tierny, line 67: same remark as before (`isRunningWithMPI()`)

julien-tierny (changes requested):
hi @m-s-will,
thanks a lot for the revision, it definitely improved.
I added a few notes for late (@eve-le-guillou)). 
there's a few general purpose MPI things which should start to be put together already at this stage (see my detailed comments). thanks!

core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
m-s-will: There is the `MPI_Initialized(int *flag)` method which allows to check whether MPI has been initialized yet (either by Paraview executed with `mpirun` or manually). 

Do you think about two MPI Utils classes then? One in `base/common` for actual MPI usage in the base code and one in `vtk/ttkAlgorithm` for usage in VTK classes? Or do you mean `base/common/MPI.h` with `ttkMPIUtils`?

core/base/arrayPreconditioning/ArrayPreconditioning.h
julien-tierny: NOTE for late (@eve-le-guillou): it looks like this should be embedded in the triangulation class.

core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
julien-tierny: I guess only one in `core/base/common` is sufficient for the moment. thanks!
(I'll probably rename it to `MPI.h` later as it is going to include more than just utils in the near future).

julien-tierny (approved):

julien-tierny:
alright michael, thanks a lot for this awesome PR. this is looking really good now.
I merge that in :)
@eve-le-guillou: can you give it a shot asap?

julien-tierny:
quick question: why did you create a class `core/base/arrayPreconditioning/ArrayPreconditioning` and did not add that code to `core/base/common/OrderDisambiguation`?
@JonasLukasczyk

m-s-will:
We started with adding it in `OrderDisambiguation`, but because of `ttkAlgorithm` using it, changing / experimenting with anything in it lead to very long recompiling times, when we explicitly only use the MPI stuff in the `ArrayPreconditioning` filter at the moment. 

For the future it might be worthwhile to move most of it back into `OrderDisambiguation` and to change `preconditionOrderArray`, such that the check if we are in MPI-mode or not only happens there. Then every filter using `GetOrderArray` would work directly without running the `ArrayPreconditioning` beforehand, no matter if we are using MPI or not.

core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
JonasLukasczyk: Be aware that the base code should not know anything about VTK.

JonasLukasczyk:
Having it in an own module seems cleaner to me.

core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
m-s-will: `MPI.h` itself also led to problems on my machine because of the MPI header itself being `mpi.h`. I think case sensitivity depends on the file system, therefore I would suggest a unique name.

julien-tierny: actually @eve-le-guillou is already renaming it for the next PR.

PR #762, opened 2022-04-21
[MergeTreeVisu] exlude pairs, custom arrays, copy point data and more

MatPont:
This PR adds some features to the MergeTreeVisualization class (used for the planar layout for instance).

Among the new features we found:
- Add the exclude important pairs intervals, that allow to forbid some pairs to be considered as important, useful to have a better layout and focus on some pairs in particular
- Add the custom arrays, useful when one want to add custom arrays to the point/cell data of the merge tree planar layout
- Allows to automatically copy point data of the input trees
- Add TrueNodeId array to ensure the same node ids when passing from FTMTree to vtkUnstructuredGrid to FTMTree again
- Fix some bugs (most are out of bounds access)

### DIFF
--- a/core/vtk/ttkPlanarGraphLayout/ttkMergeTreeVisualization.h
+++ b/core/vtk/ttkPlanarGraphLayout/ttkMergeTreeVisualization.h
@@ -17,9 +17,13 @@
 #include <vtkCellData.h>
 #include <vtkDoubleArray.h>
 #include <vtkFloatArray.h>
+#include <vtkImageData.h>
 #include <vtkIntArray.h>
 #include <vtkPointData.h>
 #include <vtkPoints.h>
+#include <vtkStringArray.h>
+#include <vtkTransform.h>
+#include <vtkTransformFilter.h>
 #include <vtkUnstructuredGrid.h>
 
 class ttkMergeTreeVisualization : public ttk::MergeTreeVisualization {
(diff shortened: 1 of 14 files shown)


core/vtk/ttkPlanarGraphLayout/ttkMergeTreeVisualization.h
```diff
@@ -247,13 +247,13 @@ class ttkMergeTreeVisualization : public ttk::MergeTreeVisualization {
 
   // Custom array
   void addCustomArray(std::string &name, std::vector<double> &vec) {
-    customArrays.push_back(std::make_tuple(name, vec));
+    customArrays.emplace_back(std::make_tuple(name, vec));
```
pierre-guillou: I think that with `emplace_back`, you don't need the call to `std::make_tuple`(it may create a temporary variable). At least that's what seems to be suggested in https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-emplace.html. This also applies to the two next methods.

MatPont: Done in the next commit!

julien-tierny:
ok, didn't check completely but it looks good.

julien-tierny:
ok, didn't check completely but it looks good.

PR #763, opened 2022-04-22
Projection from table

MatPont:
This PR adds the ProjectionFromTable filter to TTK.
It takes as input a surface (generated with PointSetToSurface for instance) and a table corresponding to points that will be projected on the surface. 
To do so, the user needs to select two ordering arrays (like PointSetToSurface) that will be used to find the 3D coordinates of the projected points given the values of these arrays for points in the surface.
Moreover, this PR fixes some things in PointSetToSurface like the sorting of the ordering arrays values.

### DIFF
--- a/core/base/projectionFromTable/ProjectionFromTable.h
+++ b/core/base/projectionFromTable/ProjectionFromTable.h
@@ -26,13 +26,6 @@ namespace ttk {
   public:
     ProjectionFromTable();
 
-    int preconditionTriangulation(AbstractTriangulation *triangulation) {
-      // Pre-condition functions.
-      if(triangulation) {
-      }
-      return 0;
-    }
-
     template <class triangulationType, class xDataType, class yDataType>
     void computeInputPoints(
       const triangulationType *triangulation,
@@ -50,7 +43,7 @@ namespace ttk {
       // 0 --- 2
       // |     |
       // 1 --- 3
-      std::vector<std::vector<int>> quadPoints(noPoints, std::vector<int>(4));
+      std::vector<std::array<int, 4>> quadPoints(noPoints);
       for(unsigned int i = 0; i < noPoints; ++i) {
         int i0 = 0, i1 = 0;
         for(unsigned int j = 0; j < surfaceDim[0]; ++j)
@@ -66,22 +59,21 @@ namespace ttk {
       }
 
       // Iterate through each point
-      std::vector<std::vector<double>> coef(
-        noPoints, std::vector<double>(4, 0.0));
+      std::vector<std::array<double, 4>> coef(noPoints);
       for(unsigned int i = 0; i < noPoints; ++i) {
-        std::vector<std::vector<double>> points(4, std::vector<double>(2));
+        std::vector<std::array<double, 2>> points(4);
         for(unsigned int j = 0; j < 4; ++j) {
           points[j][0] = std::get<1>(surfaceValues[quadPoints[i][j]]);
           points[j][1] = std::get<2>(surfaceValues[quadPoints[i][j]]);
         }
 
         // Find barycentric coordinates
-        std::vector<double> tableValues{static_cast<double>(tableXValues[i]),
-                                        static_cast<double>(tableYValues[i])};
-        std::vector<std::vector<double>> trianglePoints(3);
-        std::vector<int> indexes(3);
-        std::vector<double> mid{(points[3][0] - points[1][0]) / 2.0,
-                                (points[0][1] - points[1][1]) / 2.0};
+        std::array<double, 2> tableValues{static_cast<double>(tableXValues[i]),
+                                          static_cast<double>(tableYValues[i])};
+        std::vector<std::array<double, 3>> trianglePoints(3);
+        std::array<int, 3> indexes;
+        std::array<double, 2> mid{(points[3][0] - points[1][0]) / 2.0,
+                                  (points[0][1] - points[1][1]) / 2.0};
         if(tableXValues[i] > points[1][0] + mid[0]) {
           indexes[0] = 2;
           indexes[1] = 3;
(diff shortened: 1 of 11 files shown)


core/base/projectionFromTable/ProjectionFromTable.h
```diff
@@ -0,0 +1,132 @@
...
+      std::vector<std::vector<double>> &inputPoints) {
+      unsigned int noPoints = nTableValues;
+      inputPoints = std::vector<std::vector<double>>(
+        noPoints, std::vector<double>(3, 0.0));
+
+      // Find quadrant of each point
+      // 0 --- 2
+      // |     |
+      // 1 --- 3
+      std::vector<std::vector<int>> quadPoints(noPoints, std::vector<int>(4));
```
pierre-guillou: Use `std::array<int, 4>` for the internal dimension?

core/base/projectionFromTable/ProjectionFromTable.h
```diff
@@ -0,0 +1,132 @@
...
+            i1 = j;
+        quadPoints[i][0] = i0 * surfaceDim[1] + i1 + 1;
+        quadPoints[i][1] = i0 * surfaceDim[1] + i1;
+        quadPoints[i][2] = (i0 + 1) * surfaceDim[1] + i1 + 1;
+        quadPoints[i][3] = (i0 + 1) * surfaceDim[1] + i1;
+      }
+
+      // Iterate through each point
+      std::vector<std::vector<double>> coef(
+        noPoints, std::vector<double>(4, 0.0));
```
pierre-guillou: Replace with `std::vector<std::array<double, 4>> coef(nPoints)`? (`std::array`s are 0-initialized)

core/base/projectionFromTable/ProjectionFromTable.h
```diff
@@ -0,0 +1,132 @@
...
+   * The ProjectionFromTable class provides methods that
+   * projects on a surface points in a vtkTable.
+   */
+  class ProjectionFromTable : virtual public Debug {
+
+  protected:
+  public:
+    ProjectionFromTable();
+
+    int preconditionTriangulation(AbstractTriangulation *triangulation) {
```
pierre-guillou: This method seems to be quite useless...

core/base/projectionFromTable/ProjectionFromTable.h
```diff
@@ -0,0 +1,132 @@
...
+        noPoints, std::vector<double>(4, 0.0));
+      for(unsigned int i = 0; i < noPoints; ++i) {
+        std::vector<std::vector<double>> points(4, std::vector<double>(2));
+        for(unsigned int j = 0; j < 4; ++j) {
+          points[j][0] = std::get<1>(surfaceValues[quadPoints[i][j]]);
+          points[j][1] = std::get<2>(surfaceValues[quadPoints[i][j]]);
+        }
+
+        // Find barycentric coordinates
+        std::vector<double> tableValues{static_cast<double>(tableXValues[i]),
```
pierre-guillou: These 4 declarations may benefit from `std::array` too

core/base/projectionFromTable/ProjectionFromTable.h
MatPont: Done in the next commit!

core/base/projectionFromTable/ProjectionFromTable.h
MatPont: Done in the next commit!

core/base/projectionFromTable/ProjectionFromTable.h
MatPont: Removed in the next commit!

core/base/projectionFromTable/ProjectionFromTable.h
MatPont: Done in the next commit!

core/base/projectionFromTable/ProjectionFromTable.h
```diff
@@ -66,22 +59,21 @@ namespace ttk {
...
         // Find barycentric coordinates
-        std::vector<double> tableValues{static_cast<double>(tableXValues[i]),
-                                        static_cast<double>(tableYValues[i])};
-        std::vector<std::vector<double>> trianglePoints(3);
-        std::vector<int> indexes(3);
-        std::vector<double> mid{(points[3][0] - points[1][0]) / 2.0,
-                                (points[0][1] - points[1][1]) / 2.0};
+        std::array<double, 2> tableValues{static_cast<double>(tableXValues[i]),
+                                          static_cast<double>(tableYValues[i])};
+        std::vector<std::array<double, 3>> trianglePoints(3);
```
pierre-guillou: There's something off there... With `std::array` you should not need to resize the outer vector. Maybe you wanted `std::array<std::vector<double>, 3>` instead? Or even `std::array<std::array<double, 3>, 3>`?

MatPont: Indeed, it was needed to use an array of array here, I forgot one. 
Done in the next commit!

core/vtk/ttkProjectionFromTable/ttkProjectionFromTable.cpp
```diff
@@ -0,0 +1,318 @@
...
+    return (v - yRange[0]) / (yRange[1] - yRange[0]) * minXInterval * 0.99;
+  };
+  const auto cmp = [&](const std::tuple<int, double, double> &a,
+                       const std::tuple<int, double, double> &b) {
+    return std::get<1>(a) + normValue(std::get<2>(a))
+           < std::get<1>(b) + normValue(std::get<2>(b));
+  };
+
+  // sort the vector of indices/values in ascending order
+  std::sort(orderedValues.begin(), orderedValues.end(), cmp);
```
julien-tierny: this is a sequential sort.
please use OpenMp if available (see core/base/common/OrderDisambiguation.h:78, function sortVertices())

julien-tierny (changes requested):
thanks, I have just one minor remark (regarding sorts and OpenMP).

core/vtk/ttkProjectionFromTable/ttkProjectionFromTable.cpp
MatPont: Indeed, thanks. It is done in the next commit.
I've also modified PointSetToSurface and PointSetToCurve.

julien-tierny (approved):

julien-tierny:
wonderful, thanks!

PR #765, opened 2022-04-28
Added rank number in log prefix and documentation of MPI

eve-le-guillou:
In this PR:
- A new BaseMPIClass is added that inherits from BaseClass. If the compilation variable `TTK_ENABLE_MPI` is on, the class Debug will inherit from the BaseMPIClass.
- The utilities in the MPI.h file are moved to the file of BaseMPIClass and MPI.h is deleted.
- The rank number of a process is added to the prefix of TTK messages when executed in parallel.
- The TTK logo is only printed once at the beginning in parallel.
- Documentation for the use of TTK with MPI is added in the MPI.md file.

An example of a parallel TTK pipeline can be found in the PR #761 .

### DIFF
--- a/MPI.md
+++ b/MPI.md
@@ -1,0 +1,54 @@
+## Running TTK and Paraview with MPI
+
+### Compilation
+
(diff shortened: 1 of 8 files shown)


julien-tierny:
alright, looks good!

PR #766, opened 2022-04-28
[MergeTree*] bary. size limit, new mix distances, dist.mat. double input and more

MatPont:
This PR adds some features to the merge tree filters (especially MergeTreeClustering) and correct some bugs.

Here after an almost exhaustive list of the new features:
- **MergeTreeClustering** :
  - Limit size of the barycenter with a percentage of the sum of the nodes in input
  - Better clustering first tree initilization (farest input for the first tree then kmeans++ for the other trees)
  - New mix distances when double input
  - Move functions from MergeTreeBase to MergeTreeDistance
  - Remove some old code of MergeTreeDistance
  - Fix some very particular cases when passing from BDT to MT (due to multi-persistence pairs, when a saddle is involved in more than one pair)
  - Parallelize the planar layout of all trees

- **MergeTreeDistanceMatrix** :
  - Double input: allows to take in input join and split trees (like MergeTreeClustering can do)
  - Use Field Data parameters: this check-box can be used to automatically load the parameters (persistence threshold, epsilons etc.) from the field data of the multi-block in input if they are present. Useful in order to avoid to manually copy the parameters in the GUI of a MergeTreeClustering filter for instance, and to keep the same parameters along the pipeline by specifying them at a single point.

### DIFF
--- a/core/base/mergeTreeClustering/MergeTreeBarycenter.h
+++ b/core/base/mergeTreeClustering/MergeTreeBarycenter.h
@@ -39,6 +39,8 @@ namespace ttk {
     bool progressiveBarycenter_ = false;
     double progressiveSpeedDivisor_ = 4.0;
     double alpha_ = 0.5;
+    unsigned int barycenterMaximumNumberOfPairs_ = 0;
+    double barycenterSizeLimitPercent_ = 0.0;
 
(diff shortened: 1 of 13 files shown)


julien-tierny:
ok, looks good. thanks!

PR #767, opened 2022-04-28
Bugfix for negative scalar values for MPI

m-s-will:
`std::numeric_limits<T>::min()` leads to unexcpected behaviour when dealing with floating-point types, return the minimum positive normalized value. However, we want the value which has no values less than it (e.g. -FLT_MAX for floats), therefore we need `std::numeric_limits<T>::lowest()`.


julien-tierny:
awesome, thanks!

PR #769, opened 2022-05-09
Fix some memory-related issues

pierre-guillou:
This PR fixes some memory-related issues (mostly found thank to [heaptrack](https://apps.kde.org/fr/heaptrack/)).
Among them:
* a memory leak was fixed in the standalones by wrapping a naked pointer into a `vtkSmartPointer`,
* to reduce the memory taken by the `DiscreteGradient` cache in a iterative context (Catalyst), the cache has been reimplemented with a LRU (Least-Recently Used) approach,
* the last dynamic allocations were removed from `Geometry`,
* `std::vector`s were replaced with `std::array`s in `FiberSurface`, `RangeDrivenOctree`, `ReebSpace` & `ContinuousScatterPlot`,
* in `JacobiSet`, `reserve` has been called before filling a vector in a loop,
* in `RangeDrivenOctree`, guards were placed around `printMsg` calls to avoid string concatenations.

The last changes should slightly improve the performance of the mentioned filters (nothing revolutionary though).

Enjoy,
Pierre

### DIFF
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -12,11 +12,11 @@
 #pragma once
 
 // base code includes
+#include <Cache.h>
 #include <Geometry.h>
 #include <Wrapper.h>
 
 #include <array>
-#include <map>
 #include <ostream>
 
 #ifdef TTK_ENABLE_KAMIKAZE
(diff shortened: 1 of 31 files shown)


julien-tierny:
ok, thanks!
why not using boost's lru cache directly?

pierre-guillou:
I don't think that the Boost LRU cache is meant to be used directly since it resides in the `boost::compute::detail` namespace (is it part of the Boost headers or distributed as part as a Boost library?). It was easier to reimplement it using the C++ stdlib only and avoiding another big dependency (AbstractTriangulation header) to Boost.

julien-tierny:
ok, thanks for the clarification!

PR #770, opened 2022-05-09
Fix (most of) implicit triangulation performance loss

pierre-guillou:
Following #750, a drop in the implicit triangulation performance was detected. The performance loss seems to be coming from the 
use of virtual functions (preventing method inlining) and returning copies instead of references (due to inheritance).

This PR uses [CRTP](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern) to circumvent the use of virtual functions  and to allow to return either references or copies. The changes should restore the majority of the performance loss  (we should be at most 5% slower than before #750) and have been applied to `ImplicitTriangulation` and `PeriodicImplicitTriangulation`.

`ttkUtils`' macros, `PersistenceDiagram` and `ScalarFieldCriticalPoints` progressivity checks were updated to better take into account the new triangulations.

Enjoy,
Pierre

### DIFF
--- a/core/base/implicitTriangulation/ImplicitPreconditions.h
+++ b/core/base/implicitTriangulation/ImplicitPreconditions.h
@@ -6,37 +6,41 @@ namespace ttk {
   /**
    * @brief Implicit Triangulation class with preconditioning
    */
-  class ImplicitWithPreconditions final : public ImplicitTriangulation {
+  class ImplicitWithPreconditions final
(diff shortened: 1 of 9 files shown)


julien-tierny:
that looks good.
I'll merge it for now to compensate the performance loss but we'll need to discuss the CRTP design pattern (I have the impression something simpler could achieve the same outcome).

PR #771, opened 2022-05-13
Fix GCC 12 includes

pierre-guillou:
GCC 12 is now available in ArchLinux and, like every major GCC update, there are issues. This time, several includes were missing. Contrary to the GCC 10 -> 11 update (c.f. https://github.com/topology-tool-kit/ttk/pull/613), only TTK is affected this time (ParaView builds fine and ParaView headers are unaffected).

This PR adds the missing standard library includes to make it build correctly with GCC 12. Sadly, there's no way today to ensure that further PRs won't break the build since GCC 12 is not available yet in the GitHub Actions CI.

Enjoy,
Pierre

### DIFF
--- a/core/base/assignmentSolver/AssignmentMunkresImpl.h
+++ b/core/base/assignmentSolver/AssignmentMunkresImpl.h
@@ -2,6 +2,8 @@
 
 #include <AssignmentMunkres.h>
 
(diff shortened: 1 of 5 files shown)


julien-tierny:
great, thanks!

PR #772, opened 2022-05-13
[MPI] Fixes and Optimization for ArrayPreconditioning, more functionality for BaseMPIClass

m-s-will:
This fixes a bug which could lead to some ranks not sending messages correctly when not all ranks are direct neighbors e.g. a long and narrow wavelet where 4 ranks are in a 4x1 grid, not a 2x2 grid. Previously it could happen that rank 3 fails to correctly send the messages to rank 0. This should be fixed through extensive rework and unique send / recv tags per rank.

This PR also removes checks for the GhostCellArray when it is not needed, because we already have the RankArray.

Also moves some functionality from ArrayPreconditioning to the BaseMPIClass and adds the exchangeGhostCells method which allows us to exchange all the ghostcell information of a scalar array through a single call. Each rank will then request the needed scalars from their direct neighbors. Currently getNeighbors only works correctly if there is at least one layer of ghostcells, in the future we want to modify it to also work with only shared vertices (e.g. if one rank owns everything on the boundary, they are not ghostcells for him. They are only ghostcells for the neighbor).


julien-tierny:
thanks for reviewing PR #773 before resolving conflicts on this one.

m-s-will:
Integrated changes from PR #773 and moved functionality to `MPIUtils.h`

julien-tierny:
ok, looks good!

PR #773, opened 2022-05-13
MPI: Support edge/triangle local <-> global id maps in triangulation

pierre-guillou:
This PR extends the triangulation modules to generate mappings allowing correspondences between edges (resp. triangles) local ids (what is managed by the TTK triangulation in a given MPI process) and a global id that corresponds to the local edge (resp. triangle) id when using only one MPI process (identical with the `ExplicitTriangulation`, less so with the `ImplicitTriangulation`, due to numbering issues).

The PR introduces new triangulation methods, `getEdgeLocalId`, `getEdgeGlobalId`, `getTriangleLocalId` and `getTriangleGlobalId`. Precondition methods are automatically called to fill internal buffers. The ParaView pipeline should be as follows:

* VTI with "GenerateGlobalIds" data pre-computed on 1 process -> "GhostCellsGenerator" -> Any TTK filter that uses a triangulation.
* VTU with "GenerateGlobalIds" data pre-computed on 1 process -> "Redistribute Data-set" -> "GhostCellsGenerator" -> Any TTK filter that uses a triangulation.

Right now, the code only works for Explicit & Implicit triangulations (no support yet for Compact or Periodic triangulations). The code does roughly the same thing for both triangulations and fills internal buffers (the only change concerns the `GlobalCellIds` cell data array that identifies tetrahedron in Explicit and cubes in Implicit). A further PR should support the on-the-fly computation of global <-> local ids for Implicit triangulations and remove the need of the internal buffer.

In a similar way to the OpenMP support, `BaseMPIClass` has been merged into `BaseClass`. Another global variable, `ttk::MPIsize_`, stores the current number of MPI processes. MPI functions have been moved into `MPIUtils.h` (this should not collide with `mpi.h`), which is included in `BaseClass.h`.

Enjoy,
Pierre


julien-tierny:
alright, thanks a lot!

note for @pierre-guillou @eve-le-guillou @m-s-will 
I guess it would be nice to add the local2global members for vertices as well
`std::vector<SimplexId> vertexLidToGid_{};`
`std::unordered_map<SimplexId, SimplexId> vertexGidToLid_{};`

then @m-s-will you may need to update the array preconditioning accordingly.

PR #774, opened 2022-05-13
Docker Scripts Update and GitHub Action for Docker Package

floWetzels:
This PR contains an updated dockerfile using multi-stage builds, as well as up-to-date bash scripts for running paraview 5.10.1 connected to a pvserver with ttk in a docker container or pvpython in a docker container.
Furthermore, it adds a github workflow for building two docker packages, one for running ttk and one for developing ttk.

I should note that the github workflow is configured to use a self-hosted runner. I could change that back to the default one as an intermediate step, if this is prefered and we can then discuss how to integrate the self-hosted runner later.


julien-tierny:
looks great, thanks!

.github/workflows/docker.yml
```diff
@@ -0,0 +1,47 @@
+name: docker_build
+
+on:
+  push:
+    branches:
+      - 'dev'
```
pierre-guillou, line 6: Will this line trigger the workflow at every push to the `dev` branch for every user that has forked TTK?

floWetzels: Yes, that's right. Didn't think about the forks.
@julien-tierny What should I do about this?

pierre-guillou: What I do usually for the other workflows is do disable the branch push trigger and enable them only for pull requests (maybe not relevant in this case) but also enable the push trigger on specific tags (with at least one unique tag per workflow).

See for instance .github/workflows/package.yml (or every other workflow) for how I've done it.

The README file in the workflows directory also provides some user help on how to create and push tags to trigger the specific workflows.

floWetzels: Ah perfect, thanks!
So how should we proceed? Should I do another PR for that?

julien-tierny:
yes, please!
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

PR #775, opened 2022-05-16
Docker Workflow Fix

floWetzels:
As discussed, this PR changes which events trigger the docker package workflow. It is now only triggered by specific tags, not on every push on the dev branch. I chose those tags, that were used in `package.yml`, is that ok?

Another question: Should I also change the runner?

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -26,6 +26,12 @@ To create packages, delete any GitHub Release named "package", then:
 git tag -f package && git push -f github_remote package
 ```
 
+To create docker packages:
+
+```sh
+git tag -f docker && git push -f [remote] docker
+```
+
 Workflows
 ---------
 
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -2,8 +2,15 @@ name: docker_build
 
 on:
   push:
-    branches:
-      - 'dev'
+    # triggered on tag pushes with tags beginning with either "v" or "dev"
+    branch-ignore:
+      - '*'
+    tags:
+      - 'v*'
+      - 'dev*'
+      - '*-*-*'
+      - '*.*.*'
+      - 'pack*'
 
 jobs:
   docker:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -10,7 +10,7 @@ on:
       - 'dev*'
       - '*-*-*'
       - '*.*.*'
-      - 'pack*'
+      - 'docker*'
 
 jobs:
   docker:


pierre-guillou:
Thanks for your modifications!

However, I'm afraid that the tags that your require here are also used in the `package` workflow. If you look at the other workflows, you'll see that I'm using specific tags for each one of them, to allow to trigger them independently. I would rather suggest to use something like `docker*`, and not the other shared tag specifications, to make sure that launching other workflows don't launch this one by mistake.

(Also, the command I would use to trigger this workflow by push events: 
`$ git tag -f docker && git push -f [remote] docker`).

julien-tierny:
@floWetzels could you address @pierre-guillou 's remark about the `docker*` tag?

julien-tierny:
@floWetzels could you address @pierre-guillou 's remark about the `docker*` tag?

julien-tierny:
also, could you mention in `scripts/docker/README.md` how to trigger this action (cf @pierre-guillou 1-liner command)

floWetzels:
Yes sure, I can do those changes. Just that I understand it correctly, as I'm not very familiar with github actions: I should do the following two things, change the docker.yml to just use `docker*` tags and add Pierre's line to the readme for creating the tags?

Just for clarity: My idea behind using the same tags as in `package.yml` was that creating docker packages should basically happen at the same tags as creating binary packages. Did I misunderstand something or is this just about the possibility to trigger both workflows independently?

julien-tierny:
actually, your remark on the tags of `package.yml` is a very good point.
that'd be much more convenient to have only one tag for all package forms.

as a matter of fact, I think I would probably recommend to keep the existing tags as-is, and *add* the tag `docker*` on top, such that people can trigger the build of docker packages independently of the other package forms, if they want (which I suspect is one of the features you guys planned on using at Kaiserslautern, right?)

julien-tierny:
(sorry I sent the comment right after the latest commit)

I'd suggest to add the `pack*` tag as well (to make sure the docker packages are also built upon this tag)

pierre-guillou:
Well, I'd like to have at least one specific tag per workflow (to be able to trigger them independently), such as `pack*` for the packaging workflow (and `docker*` for the current one).

Otherwise, the `v*`, `*-*-*` or `*.*.*` specs are used to create binary releases tied to specific TTK versions (in semver format). You might want to keep them if you want users to trigger your workflow when creating release packages. 

I don't think the `dev*` spec has a purpose anymore (I used that at the beginning when testing GitHub Actions).

julien-tierny:
alright, let's merge it that way then.

julien-tierny:
(the `dev` tag was meant to generate custom binary packages in between releases when needed)

floWetzels:
Just a quick remark: I just saw that you asked me to change the readme in `scripts/docker/`, however, I misread that and added Pierre's command to the readme in `.github/workflows/`. Is this a big problem? I actually wouldn't know where this fits into the `scripts/docker/README.md`.

julien-tierny:
yep, makes perfect sense. thanks for catching that up :)

julien-tierny:
(the right spot is indeed `.github/workflows/`)

floWetzels:
Perfect!

PR #776, opened 2022-05-19
[MPI] Add MPI support for ScalarFieldCriticalPoints filter and MPI preconditioning

eve-le-guillou:
This PR adds an MPI support for the ScalarFieldCriticalPoints filter for vti files. This filter requires Ghost Points, with a Ghost Layer of 2, as well as Global Point Ids. 
It also adds preconditioning for MPI pipelines by testing whether Ghost Points, Global Point Ids and RankArray are already created. If not, Ghost Points as well as Global Point ids will be generated. A warning willl also be printed when generating Global Ids, as generating the Ids in parallel may result in different Ids as when generated in sequential.
The initialization of MPIrank_ and MPI_size_ are also modified as the previous version would impact performance greatly.
A new compilation variable has been created: TTK_ENABLE_MPI_TIME. This variable should be set to ON when the user wants a very precise time measurement of the execution when using MPI. This measurement may slow down execution overall, hence it should be set to OFF when not useful.

Here is an example of use of the ScalarFieldCriticalPoints filter, with 'data', a vti file with 'GenerateGlobalIds' and 'ArrayPreconditioning' pre-computed on one process:

    tTKScalarFieldCriticalPoints1 = TTKScalarFieldCriticalPoints(registrationName='TTKScalarFieldCriticalPoints1', Input=data)
    tTKScalarFieldCriticalPoints1.ScalarField = ['POINTS', 'NameOfScalarField']
    tTKScalarFieldCriticalPoints1.Backend = 'Default generic backend'


m-s-will:
Just a note for further discussion:
@JonasLukasczyk and I were recently talking about the necessity of the GhostPointArray in later steps, e.g. I removed it in my PathCompression filter because all functionality and more can be achieved by using the RankArray (basically we can  use `ghostPointArray[i]` <=> `rankArray[i] == ttk::MPIrank_` ) and RankArray gives us more functionality with the knowledge of the owner ranks. We might want to discuss for which one we check and which one we use for uniformity.

JonasLukasczyk:
Hi Eve, thank you for the PR. I don't understand why you need two ghost-cell layers. One layer should be sufficient where each rank only processes the vertices that are owned by the rank (`rankArray[i] == ttk::MPIrank_`).

eve-le-guillou:
> Just a note for further discussion: @JonasLukasczyk and I were recently talking about the necessity of the GhostPointArray in later steps, e.g. I removed it in my PathCompression filter because all functionality and more can be achieved by using the RankArray (basically we can use `ghostPointArray[i]` <=> `rankArray[i] == ttk::MPIrank_` ) and RankArray gives us more functionality with the knowledge of the owner ranks. We might want to discuss for which one we check and which one we use for uniformity.

Thank you for your remarks.
It is true that the GhostPointArray can be deduced from the RankArray. However, some filters, such as ScalarFieldCriticalPoints, don't require the computation of RankArray and only need to know if a point is Ghost or not, so that would require unnecessary computation. An argument can be made that RankArray will be required by so many filters that one should just compute it by default as it will surely be needed at some point in any pipeline. I don't know enough about all the filters to have an answer, we should certainly discuss more about this.
The call to a VTK ghost Generator will still be required at some point, as it doesn't just flag the ghost points/cells, it also creates the layers, etc. But maybe PointGhostArray won't be necessary in TTK. 

Here, I tried to have as little computation as possible, hence I didn't compute the RankArray, but that is likely to change as we look into more and more filters.

> Hi Eve, thank you for the PR. I don't understand why you need two ghost-cell layers. One layer should be sufficient where each rank only processes the vertices that are owned by the rank (`rankArray[i] == ttk::MPIrank_`).

The two layers are required by the ScalarFieldCriticalPoints filter. Indeed, in the case of Saddle 1 and Saddle 2 in dimension 3, it is not only required to have information on the neighbors of the point of interest, but also to know whether those neighbors are on the boundary of the global domain. With only one layer, the neighbors that are ghosts will be considered on the boundary because it is on the local boundary, as processes wouldn't have access to all the neighbors of the ghost neighbors. One way to know if the ghost neighbor of a point of interest is on the global boundary with limited communication is to have two layers of ghost cells. 
@julien-tierny thinks that a lot of filters will also need to know whether the neighbors of a point of interest are on the boundary, therefore 2 layers will always be required, as it is not very costly.

julien-tierny:
note: by "boundary", @eve-le-guillou means "global domain boundary"

eve-le-guillou:
The function isRunningWithMPI was changed in my latest commit, as it wasn't working as intended: the goal of the function was to test whether TTK was being executed using `mpirun`. It did so by checking if MPI had been initialized, with `MPI_Initialized`. 
However, Paraview will always initialize as long as it has been compiled with MPI, regardless of whether is was executed indeed with `mpirun`. Therefore, `isRunningWithMPI` would return true as long as Paraview had been compiled with MPI. In this commit, the test has been changed to `ttk::MPIsize_ > 1`, to determine whether the execution is performed on more than one process. 
If TTK is run with `mpirun`, but only on one process, the behavior will be the same as if it hasn't been run with MPI.

core/base/common/BaseClass.h
```diff
@@ -62,6 +62,11 @@ namespace ttk {
 
   class BaseClass {
   public:
+#if TTK_ENABLE_MPI
+    long int *GlobalIdsArray;
```
julien-tierny: To follow TTK's variable naming convention, these attribute names should not start with a capital letter, they should follow the camel naming and finish with a `_` postfix (to quickly differentiate variables which are attributes from local variables).
For instance: `GlobalIdsArray -> globalIdsArray_`

Moreover, these 3 attributes are related to the triangulation. However, not all TTK modules manipulate the triangulation (for instance: persistence diagram distance computation).

Thus, I'd recommend to move these attributes (with their setters) to `AbstractTriangulation.h` (see line 3357 to 3375, where other MPI attributes have been defined).

core/base/common/BaseClass.h
```diff
@@ -75,6 +80,19 @@ namespace ttk {
       return 0;
     }
 
+#if TTK_ENABLE_MPI
```
julien-tierny: see the above remark.

core/base/scalarFieldCriticalPoints/ScalarFieldCriticalPoints.h
```diff
@@ -27,6 +27,7 @@
 
 #pragma once
 
+#include <algorithm>
```
julien-tierny: why is the `<algorithm>` needed here?
(I don't see any reason)

if it builds and runs without it, thanks for removing it.

core/vtk/ttkAlgorithm/vtk.module
```diff
@@ -3,4 +3,5 @@ NAME
 DEPENDS
   VTK::FiltersCore
 PRIVATE_DEPENDS
-  VTK::CommonCore
\ No newline at end of file
+  VTK::CommonCore
+  VTK::FiltersParallelDIY2
```
julien-tierny, line 7: I don't believe `VTK::FiltersParallelDIY2` is needed.
if it builds and runs without it, thanks for removing it.

core/vtk/ttkScalarFieldCriticalPoints/ttkScalarFieldCriticalPoints.cpp
```diff
@@ -78,6 +82,17 @@ int ttkScalarFieldCriticalPoints::RequestData(
   this->preconditionTriangulation(triangulation);
   this->setOutput(&criticalPoints_);
 
+#if TTK_ENABLE_MPI_TIME
```
julien-tierny: Please create a function `startMPItimer`, to declare and implement in `core/vtk/ttkAlgorithm/ttkUtils.h` and `core/vtk/ttkAlgorithm/ttkUtils.cpp`.

This entire snippet of code (from line 85 to 94, with the `#if` and `#endif` included) should be put altogether in that function.

core/vtk/ttkScalarFieldCriticalPoints/ttkScalarFieldCriticalPoints.cpp
```diff
@@ -92,6 +107,16 @@ int ttkScalarFieldCriticalPoints::RequestData(
   if(status < 0)
     return 0;
 
+#if TTK_ENABLE_MPI_TIME
```
julien-tierny: Same remark as above. Please create a function `stopMPItimer`.

core/vtk/ttkScalarFieldCriticalPoints/ttkScalarFieldCriticalPoints.cpp
```diff
@@ -78,6 +82,17 @@ int ttkScalarFieldCriticalPoints::RequestData(
   this->preconditionTriangulation(triangulation);
   this->setOutput(&criticalPoints_);
 
+#if TTK_ENABLE_MPI_TIME
+  Timer t_mpi;
+  // Get processes information
+  vtkMPIController *controller = vtkMPIController::SafeDownCast(
+    vtkMultiProcessController::GetGlobalController());
```
julien-tierny: Can we do differently than using VTK's handles to the MPI's controller?
Isn't it possible to access MPI communications from the base code without any interaction with VTK's handles?

core/vtk/ttkScalarFieldCriticalPoints/vtk.module
```diff
@@ -2,3 +2,4 @@ NAME
   ttkScalarFieldCriticalPoints
 DEPENDS
   ttkAlgorithm
+  VTK::ParallelCore
```
julien-tierny: Are these required by the VTK's handle to the MPI controller?

core/base/common/BaseClass.h
```diff
@@ -62,6 +62,11 @@ namespace ttk {
 
   class BaseClass {
   public:
+#if TTK_ENABLE_MPI
+    long int *GlobalIdsArray;
+    unsigned char *PointGhostArray;
```
julien-tierny: Based on the discussion with @m-s-will , I'd recommend to remove that array and only use the `rankArray_` instead.

julien-tierny (changes requested):
hi @eve-le-guillou,
thanks a lot for your PR.
this is looking good. 
could you address my remarks and suggestions?
thanks!

core/vtk/ttkScalarFieldCriticalPoints/ttkScalarFieldCriticalPoints.cpp
m-s-will: If we assume to use the default communicator (MPI_COMM_WORLD) for TTK, which should always be available, we can use MPI_BARRIER(MPI_COMM_WORLD), without using the VTK functions. 

In the base layer we _need_ to use 'raw' MPI because one could use TTK standalone without VTK, in the vtk layer it should be no problem to rely on the VTK handles, however I would also use the raw MPI for consistency.

julien-tierny: that was the point of my question indeed.
I'd recommend to implement the MPI timing util functionalities at the base code level indeed.

core/base/scalarFieldCriticalPoints/ScalarFieldCriticalPoints.h
```diff
@@ -138,6 +138,7 @@ namespace ttk {
     const std::vector<std::vector<std::pair<SimplexId, SimplexId>>>
       *vertexLinkEdgeLists_{};
     std::vector<std::pair<SimplexId, char>> *criticalPoints_{};
+    int *rankArray_;
```
julien-tierny: please initialize to `nullptr`.

core/base/triangulation/Triangulation.h
```diff
@@ -2598,6 +2598,23 @@ namespace ttk {
                              const unsigned char *const ghostCellMask) {
       this->abstractTriangulation_->setGlobalIds(cellGid, ghostCellMask);
     }
+
+    inline void setRankArray(int *rankArray) {
+      this->abstractTriangulation_->setRankArray(rankArray);
```
julien-tierny: please insert the test:
```
if(this->abstractTriangulation_)
  this->abstractTriangulation_->setRankArray(rankArray);
```

core/base/triangulation/Triangulation.h
```diff
@@ -2598,6 +2598,23 @@ namespace ttk {
...
+    }
+
+    inline int *getRankArray() {
+      return this->abstractTriangulation_->getRankArray();
```
julien-tierny: same remark as above.

core/base/triangulation/Triangulation.h
```diff
@@ -2598,6 +2598,23 @@ namespace ttk {
...
+    }
+
+    inline void setGlobalIdsArray(long int *globalIds) {
+      this->abstractTriangulation_->setGlobalIdsArray(globalIds);
```
julien-tierny: same remark as above

core/base/triangulation/Triangulation.h
```diff
@@ -2598,6 +2598,23 @@ namespace ttk {
...
+    }
+
+    inline long int *getGlobalIdsArray() {
+      return this->abstractTriangulation_->getGlobalIdsArray();
```
julien-tierny: same remark as above

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -9,6 +9,10 @@
 #include <vtkCellTypes.h>
 #include <vtkCommand.h>
 #include <vtkDataSet.h>
+#include <vtkGenerateGlobalIds.h>
```
julien-tierny: I would put this include within the `#if TTK_ENABLE_MPI` (it's only used when MPI is enabled)

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -382,6 +386,42 @@ int ttkAlgorithm::RequestDataObject(vtkInformation *ttkNotUsed(request),
   return 1;
 }
 
+#if TTK_ENABLE_MPI
+void ttkAlgorithm::MPIPreconditioning(vtkInformationVector **inputVector) {
```
julien-tierny: I'd change the prototype of this function to something more intuitive, such as:
```
void ttkAlgorithm::MPIPreconditioning(vtkDataSet **input) {
```

then, when calling this function, you'd need to do:
```
MPIPreconditioning(vtkDataSet::GetData(inputVector[0]));
```

I believe this is more intuitive. Plus, we really only need the dataset inside that function.

core/vtk/ttkAlgorithm/ttkUtils.cpp
```diff
@@ -359,3 +362,26 @@ int ttkUtils::CellVertexFromPoints(vtkDataSet *const dataSet,
 
   return 1;
 }
+
```
julien-tierny: it seems that the functions `startMPITimer` and `endMPITimer` can go to `core/base/common/MPIUtils.h`.
everything that does not include VTK things can be moved down there.
(this was implied in our previous discussion with @m-s-will on the topic)

core/vtk/ttkAlgorithm/vtk.module
julien-tierny: Is that still needed???!!!

julien-tierny (changes requested):
hi @eve-le-guillou 
thanks a lot for your edits!
it's getting there :)
could you address the last few remarks I entered?
thanks!

julien-tierny (approved):

julien-tierny:
neat & tidy, awesome!

PR #777, opened 2022-05-23
[MPI] Added MPI support for ScalarFieldNormalizer

m-s-will:
As discussed for one filter of low MPI communication, I have added simple MPI support for the ScalarFieldNormalizer filter. In addition, this could be seen as a minimal example how to check for:

1. whether TTK was build with MPI
2. whether TTK is running with MPI

Another note: the ScalarFieldNormalizer filter itself was, as far as I can see, written a long time ago and not really updated. Bringing it up to modern standards could be a PR by itself, then maybe for the dev branch.


julien-tierny:
simple enough!
thanks

PR #778, opened 2022-05-25
[MPI] MPI support for Scalar field smoother

m-s-will:
This PR adds MPI support for the scalar field smoothing filter by exchanging the values of the ghost cells after every iteration step. As long as these values are exchanged, the smoothing in the next step has the correct values and it can therefore be done with as many iterations as wanted.

This also silences Open MPI warnings related to the C++ bindings on some compiler / mpi configurations as seen [here](https://github.com/open-mpi/ompi/issues/5157)


core/base/common/MPIUtils.h
```diff
@@ -11,8 +11,8 @@
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
-
 #if TTK_ENABLE_MPI
+#define OMPI_SKIP_MPICXX 1
```
julien-tierny, line 15: what is this?

core/base/scalarFieldSmoother/ScalarFieldSmoother.h
```diff
@@ -81,16 +85,26 @@ int ttk::ScalarFieldSmoother::smooth(const triangulationType *triangulation,
...
+#endif
   SimplexId vertexNumber = triangulation->getNumberOfVertices();
 
   std::vector<dataType> tmpData(vertexNumber * dimensionNumber_, 0);
 
   dataType *outputData = (dataType *)outputData_;
   dataType *inputData = (dataType *)inputData_;
-
+#if TTK_ENABLE_MPI
+  std::unordered_map<SimplexId, SimplexId> gidToLidMap;
```
julien-tierny: is that required by the `exchangeGhostCells` function?

julien-tierny: if so, I'd rather move that map to the `AbstractTriangulation` (see lines 3357 to 3396) and provide setters and getters for those. see the functions `preconditionDistributedEdges()`, `getEdgeGlobalId` for examples.

core/base/scalarFieldSmoother/ScalarFieldSmoother.h
```diff
@@ -81,16 +85,26 @@ int ttk::ScalarFieldSmoother::smooth(const triangulationType *triangulation,
...
+#endif
   // init the output
   for(SimplexId i = 0; i < vertexNumber; i++) {
+#if TTK_ENABLE_MPI
```
julien-tierny: if gidToLidMap is indeed needed, thanks for moving this to `preconditionDistributedVertices()` (cf above remark)

julien-tierny (changes requested):
hi @m-s-will,
thanks so much for taking care of this!!!!
could you please address the few remarks I entered?

core/base/scalarFieldSmoother/ScalarFieldSmoother.h
m-s-will: Yes, I think moving it the the triangulation is the right call because while there may be some filters not needing a triangulation, I think most which will want to use MPI need a triangulation and then don't need a separate map, I'll check it out.

julien-tierny:
hi @m-s-will 
thanks for the quick fix.
could you please resolve the conflict?
thanks!

m-s-will:
I was gone for the last few days but I will fix it in the next few days.

julien-tierny:
awesome!
no rush though :)
have a good weekend,
-- 
Dr Julien Tierny
CNRS Researcher
Sorbonne Universite
http://lip6.fr/Julien.Tierny

m-s-will:
I am having some problems with my adapted preconditioning of vertices, but also e.g. the `preconditionDistributedEdges` from PR #773. I am assuming it has to do with similar problems as #772, but am not 100% sure at the moment. An example how to use the e.g. `preconditionDistributedEdges()` would be great.

core/base/scalarFieldSmoother/ScalarFieldSmoother.h
```diff
@@ -48,13 +48,26 @@ namespace ttk {
...
-
       return 0;
     }
+#if TTK_ENABLE_MPI
+    template <class triangulationType>
+    int preconditionDistributedTriangulation(triangulationType *triangulation) {
+      // Pre-condition distributed functions.
+      if(triangulation) {
+        this->printMsg("Preconditioning Vertices");
+        triangulation->preconditionDistributedVertices();
```
julien-tierny: the calling program should not explicitly call this function.
this should be "hidden".

please move this piece of code to `core/base/triangulation/Triangulation.h`, in the functions `setInputPoints` (called for everything explicit) and `setInputGrid` (called for everything implicit).

core/base/explicitTriangulation/ExplicitTriangulation.cpp
```diff
@@ -864,6 +864,40 @@ int ExplicitTriangulation::preconditionDistributedTriangles() {
...
+  }
+  if(!ttk::isRunningWithMPI()) {
+    return -1;
+  }
+  if(this->cellGid_ == nullptr) {
+    this->printWrn("Missing global identifiers on cells");
+    return -2;
+  }
+
+  if(this->getDimensionality() != 2 && this->getDimensionality() != 3) {
```
julien-tierny: dimension 1 should be supported in principle.
(I know some ttk filters are buggy in 1D)

core/vtk/ttkScalarFieldSmoother/ttkScalarFieldSmoother.cpp
```diff
@@ -53,6 +53,14 @@ int ttkScalarFieldSmoother::RequestData(vtkInformation *ttkNotUsed(request),
 
   this->preconditionTriangulation(triangulation);
 
+#if TTK_ENABLE_MPI
+
+  this->MPIPreconditioning(input);
```
julien-tierny: filters should not call this function.
this should be called in ttkAlgorithm already (are you sure you've updated this branch?)

core/base/explicitTriangulation/ExplicitTriangulation.cpp
```diff
@@ -864,6 +864,40 @@ int ExplicitTriangulation::preconditionDistributedTriangles() {
   return 0;
 }
 
+int ExplicitTriangulation::preconditionDistributedVertices() {
```
julien-tierny, line 867: Please add this functionality to `core/base/compactTriangulation/` as well

core/base/implicitTriangulation/ImplicitTriangulation.cpp
```diff
@@ -3338,6 +3338,41 @@ int ttk::ImplicitTriangulation::preconditionDistributedTriangles() {
   return 0;
 }
 
+int ttk::ImplicitTriangulation::preconditionDistributedVertices() {
```
julien-tierny: Please add this functionality to `core/base/periodicTriangulation/` as well

core/base/scalarFieldSmoother/ScalarFieldSmoother.h
julien-tierny: (also consider the function `setStellarInputPoints`)

julien-tierny (changes requested):
hi @m-s-will 
thanks for your edits.
please find above another set of remarks.
thanks!

core/vtk/ttkScalarFieldSmoother/ttkScalarFieldSmoother.cpp
m-s-will: ah yes, I wasn't sure if it was necessary to call it myself, but it's called in ttkAlgorithm so it works without

core/base/explicitTriangulation/ExplicitTriangulation.cpp
m-s-will: it shouldn't really matter of what dimension the dataset is when only looking at the ids, so different dimensions shouldn't make any problems

core/base/scalarFieldSmoother/ScalarFieldSmoother.h
m-s-will: Here I looked at the non-distributed preconditioning steps and I thought that we don't want to do the preconditioning steps when they are not needed for the filter. Or should we assume that every (distributed) filter which wants a triangulation also will _probably_ need to transfer ghost cell values and therefore need to transform between local and global ids?

core/base/abstractTriangulation/AbstractTriangulation.h
```diff
@@ -2598,6 +2601,52 @@ namespace ttk {
 #endif // TTK_ENABLE_KAMIKAZE
       return this->getTriangleLocalIdInternal(gtid);
     }
+    virtual inline SimplexId getVertexGlobalId(const SimplexId &leid) const {
+#ifndef TTK_ENABLE_KAMIKAZE
+      if(this->getDimensionality() != 2 && this->getDimensionality() != 3) {
```
julien-tierny: I think you should also add the condition `&& this->getDimensionality() != 1`

core/base/abstractTriangulation/AbstractTriangulation.h
```diff
@@ -2598,6 +2601,52 @@ namespace ttk {
...
+        this->printErr("Only 2D and 3D datasets are supported");
```
julien-tierny: "Only 1D, 2D and 3D datasets are supported."

core/base/abstractTriangulation/AbstractTriangulation.h
```diff
@@ -2598,6 +2601,52 @@ namespace ttk {
...
+          "Please call preconditionDistributedVertices() in a pre-process.");
+        return -1;
+      }
+#endif // TTK_ENABLE_KAMIKAZE
+      return this->getVertexGlobalIdInternal(leid);
+    }
+    virtual inline int getVertexGlobalIdMap(
+      std::unordered_map<SimplexId, SimplexId> &map) const {
+#ifndef TTK_ENABLE_KAMIKAZE
+      if(this->getDimensionality() != 2 && this->getDimensionality() != 3) {
```
julien-tierny: same as above.

core/base/scalarFieldSmoother/ScalarFieldSmoother.h
```diff
@@ -47,14 +47,19 @@ namespace ttk {
       // Pre-condition functions.
       if(triangulation) {
         triangulation->preconditionVertexNeighbors();
+#if TTK_ENABLE_MPI
+        if(ttk::isRunningWithMPI())
+          triangulation->preconditionDistributedVertices();
```
julien-tierny: please remove this 4-liner and add a call `preconditionDistributedVertices()` at the end of the function `setInputPoints` and `setStellarInputPoints` of `core/base/triangulation`.

(that way, it is always called by default if MPI is detected).

julien-tierny (changes requested):
thanks a lot @m-s-will 
I think we're nearly getting there :)
if you could address my last remarks, that'd be awesome (we can discuss about them later today). thanks!

core/base/scalarFieldSmoother/ScalarFieldSmoother.h
julien-tierny: in short, yes :)

core/base/scalarFieldSmoother/ScalarFieldSmoother.h
m-s-will: Hmmm, `preconditionDistributedVertices()` needs the GlobalIdArray which is set in `MPIPreconditioning()`in `ttkAlgorithm.cpp`. Would it be more elegant to move the `MPIPreconditioning` into the triangulation itself, or to move a call to  `preconditionDistributedVertices()` into ttkAlgorithm?

julien-tierny: another option would be to add a call to `preconditionDistributedVertices()` right at the end of the implementation of `setGlobalIdsArray()`. I think that would make more sense. what do you think?

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -403,6 +403,7 @@ void ttkAlgorithm::MPIPreconditioning(vtkDataSet *input) {
       static_cast<long int *>(ttkUtils::GetVoidPointer(
         input->GetPointData()->GetArray("GlobalPointIds"))));
   }
+  triangulation->preconditionDistributedVertices();
```
julien-tierny, line 406: thanks to that one liner, I believe *all* other calls to `preconditionDistributedVertices()` can now be safely removed.

julien-tierny (changes requested):

julien-tierny:
alright, wonderful, let's go!
thanks a lot for your hard work!!

PR #779, opened 2022-05-27
TriangulationRequest: Extract simplices on domain boundary

pierre-guillou:
This PR extends the `TriangulationRequests` filter to extract simplices on the domain boundary (or, for simplices of the highest dimension, the cells that have a face on the boundary). It leverages the `isVertexOnBoundary`, `isEdgeOnBoundary` & `isTriangleOnBoundary` Triangulation methods.

The goal of this PR is to get an alternative to the `FeatureEdges` ParaView filter, which has a way larger memory footprint, in the relevant ttk-data state files.

Enjoy,
Pierre


julien-tierny:
looks good, thanks!

PR #780, opened 2022-05-27
PersistenceDiagramClustering: Fix (VTU) format of centroids diagrams

pierre-guillou:
This PR adds missing data arrays to the centroids diagrams of the `PersistenceDiagramClustering` filter. These data arrays have been missing following since #745. This is needed for instance to compute the distance matrix between centroids.

In a second commit, as a follow-up to #717, I modified the base module API of the `PersistenceDiagramDistanceMatrix` module to replace obscure template parameters with fully-qualified types using forward declaration. This helps readability while keeping the dependency to the `Auction` module in the .cpp file.

Enjoy,
Pierre


julien-tierny:
ok, looks good!

PR #781, opened 2022-05-27
Smoothers: Also use current vertex value

pierre-guillou:
This PR alters the `ScalarFieldSmoother`, `GeometrySmoother` and `SurfaceGeometrySmoother` filters by extending the sliding window on vertices to also take the current vertex into account (currently only the neighbors scalar field values/coordinates are used).

While these modifications may reduce the effect of individual iterations, it fixes instabilities in 1D when points/scalar field follow a triangle wave pattern: `/\/\/\` currently becomes `\/\/\/` at the next `GeometrySmoother`/`ScalarFieldSmoother` iteration. 

With this PR, the filters should behave closer to traditional mean filters.

Since these filters are largely used in the ttk-data state files & Python scripts, errors in the relevant CI workflow are expected (and should be fixed in a further ttk-data PR).

Enjoy,
Pierre

### DIFF
--- a/core/base/scalarFieldSmoother/ScalarFieldSmoother.h
+++ b/core/base/scalarFieldSmoother/ScalarFieldSmoother.h
@@ -115,16 +115,16 @@ int ttk::ScalarFieldSmoother::smooth(const triangulationType *triangulation,
         continue;
 
       for(int j = 0; j < dimensionNumber_; j++) {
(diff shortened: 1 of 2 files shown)


julien-tierny:
alright, thanks!
thanks also for the upcoming PR on ttk-data!

PR #782, opened 2022-06-01
[MPI] Introduction of PathCompression module

m-s-will:
This PR adds the PathCompression module which computes the ascending and descending manifold of a scalar field. It does 
so by, for each vertex, pointing the largest / smallest value of that vertex and then iteratively compressing the path to nearest critical point by pointing to the vertex your current target is currently pointing at. This doubles the path length in each step until each vertex stops changing. Then the whole scalar field is tesselated and each tile is represented by the global id of the critical point lying in it.

For the input we expect a rankArray and a globalIdArray similarly to other MPI filters. As an output, we receive two arrays, one for the ascending and one for the descending manifold. 

This PR also changes the `getNeighbors()` function to correctly work when there are actually only one-sided neighborhood relations e.g. rank 0 and 1 share one row of vertices, but rank 0 is the owner for all of them. This is not relevant when having _actual_ ghostcells, but still noteworthy when working with MPI enabled in general.

This module already uses the changes to triangulations in PR #778 , so the checks are expected to fail and this PR should not be merged until after that PR is merged.


julien-tierny:
will discuss this PR tomorrow with @JonasLukasczyk

PR #784, opened 2022-06-03
Merge Tree filters new input format

MatPont:
This PR modifies the mergeTree filters of TTK to change the format of the input.
Instead of having a multiblock of multiblock where each inner block is the nodes, arcs and segmentation (optional) of a tree, we now have a multiblock of three multiblocks, the first one contains all the nodes, the second one all the arcs and finally all the segmentations (optional) in the last inner multiblock.

This allows to be closer to the way ParaView/VTK manages filters output of a multiblock (e.g. computing a FTMTree of a cinema database).


julien-tierny:
wonderful, thanks!

Issue #783, opened 2022-06-02
RankArray strange behavior on MPI branch


eve-le-guillou:
Hi all,

Some points have a wrong RankArray value. This only happens sometimes, not always, maybe 1 in 20 iterations. Adding an output of the RankArray array inside the RequestData function of the TTKGhostCellPreconditioning filter makes the problem disappear. I have noticed that the problem happens for the points with the following global ids: 32640, 32639, 32897, however it may happen for more points, I don't know.

**To Reproduce**

1. Add a small snippet of code in order to see the behavior. The should be added in the file ttkScalarFieldCriticalPoints.cpp ligne 72, in the function RequestData:

          long int *globalIds2 = triangulation->getGlobalIdsArray();
          rankArray_ = triangulation->getRankArray();
          for (int i = 0; i< triangulation->getNumberOfVertices(); i++){
              if (globalIds2[i] == 32640 || globalIds2[i]== 32639 || globalIds2[i]==32897){
              printErr("Element "+std::to_string(globalIds2[i])+ " possessed with RankArray: "+std::to_string(rankArray_[i]));
            }
          }
2. Compile TTK again.

3. Retrieve the ctBones.vti data available in ttk-data: [https://github.com/topology-tool-kit/ttk-data/blob/dev/ctBones.vti](url)
4. Apply the following pipeline in sequential, without MPI:

        #### import the simple module from the paraview
        from paraview.simple import *
        
        # create a new 'XML Image Data Reader'
        ctBonesvti = XMLImageDataReader(registrationName='ctBones.vti', FileName=['ctBones.vti'])
        ctBonesvti.PointArrayStatus = ['Scalars_']
        ctBonesvti.TimeArray = 'None'
        
        # create a new 'TTK ArrayPreconditioning'
        tTKArrayPreconditioning1 = TTKArrayPreconditioning(registrationName='TTKArrayPreconditioning1', Input=ctBonesvti)
        tTKArrayPreconditioning1.PointDataArrays = ['Scalars_']
        
        # create a new 'Generate Global Ids'
        generateGlobalIds1 = GenerateGlobalIds(registrationName='GenerateGlobalIds1', Input=tTKArrayPreconditioning1)
        
        SaveData("ctBones_prepared.pvd",generateGlobalIds1)

5. Now, run in parallel, using the following command:

[...]

PR #785, opened 2022-06-08
[MPI] Fix for rare RankArray errors

m-s-will:
This PR should fix issue #783 by restructuring the GhostCellPreconditioning to work with MPI_Send instead of MPI_Isend.

It also modernizes the MPI usage in the module a bit, especially by not using MPI_COMM_WORLD, but instead duplicating it and therefore using a unique communicator for this module so no interference from the outside can happen. While this was not the source of error for this issue. This may have caused other issues in other modules and having an unique communicator should therefore be preferred. Whether there is _one_ TTK communicator or a new communicator for every module can be discussed in the future.


julien-tierny:
great, thanks!

PR #786, opened 2022-06-10
[MPI] Changing MPI_COMM_WORLD to unique TTK Communicator

m-s-will:
Using MPI_COMM_WORLD may lead to interference from VTK / Paraview / other applications using MPI, which may have already led to a few problems where we then maybe fixed the symptoms and not the cause (e.g. going from send / recv to gather / scatter). 
For using distinct communicators we have two possibilities:

1. Using one "TTK Communicator" which is created when TTK is first started and destroyed when finalizing MPI ( currently implemented in the commits)
2. creating a new communicator at the beginning of each filter and freeing it when the filter is finished. 

While creating new communicators for each filter may be cleaner, I don't think multiple filters run at the same time (which could create the same interferences between them) and additionally, creating new communicators may incur additional performance costs.

We should definitely move away from MPI_COMM_WORLD, this PR may be used as a place for discussion how exactly we will do that.

### DIFF
--- a/core/base/arrayPreconditioning/ArrayPreconditioning.h
+++ b/core/base/arrayPreconditioning/ArrayPreconditioning.h
@@ -194,7 +194,7 @@ namespace ttk {
           MPI_Send(orderResendValues[rankIdOfMaxScalar].data(),
                    orderResendValues[rankIdOfMaxScalar].size(), MPI_IT,
                    rankIdOfMaxScalar, intTag * rankIdOfMaxScalar,
-                   MPI_COMM_WORLD);
+                   ttk::MPIcomm_);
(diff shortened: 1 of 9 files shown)


julien-tierny:
looks good to me. I think this PR is ready to go.
my understanding is that as long as filters are used together in a legit VTK-pipeline, there should be no concurrency between them (thus, we should be good to go).
@JonasLukasczyk what do you think?

JonasLukasczyk:
Using the global ttk communicator seems fine to me for now.

julien-tierny:
seems like a done deal.
just a thought: I think we should soon merge the mpi branch into the main dev branch.
we had some discussions over here and we believe it's for the best to have everything in one spot.
maybe we'll do that next week.

PR #787, opened 2022-06-14
Fix ReturnSaddleConnectors interactions with DiscreteGradient cache

pierre-guillou:
Since the ReturnSaddleConnectors algorithm modifies in-place the DiscreteGradient internal structure, it is not compatible with the DiscreteGradient cache added in #744 .

To fix this issue, this PR introduces a optional bypass of the DiscreteGradient cache. The internal structure is, in this case, stored in the DiscreteGradient instance (identical to before #744). This way, using ReturnSaddleConnectors will recompute the DiscreteGradient each time and will not clobber the cache.

Enjoy,
Pierre

### DIFF
--- a/core/base/discreteGradient/DiscreteGradient.h
+++ b/core/base/discreteGradient/DiscreteGradient.h
@@ -337,7 +337,8 @@ saddle-connectors.
 triangulation.
        */
       template <typename triangulationType>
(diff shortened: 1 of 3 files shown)


julien-tierny:
nice, thanks!

PR #788, opened 2022-06-15
[MPI] Reworked GhostCellPreconditioning to use neighborhood relations and MPI_Sendrecv

m-s-will:
This PR reworks GhostCellPreconditioning to already use rank neighborhood relations which can be calculated by putting a bounding box around the points of every rank, inflating it by an epsilon and checking for overlaps. This allows us to move away from broadcasting the needed ids to every other rank and from awaiting an unknown number of receives until we have enough rankArray values. Instead we know that we expect to send one message to each neighbor and receive one message from one neighbor, which can be done with MPI_Sendrecv. 

This may allow for performance increases depending on the rank layout and the amount of ranks. In theory it would be possible to remove the received values from `currentRankUnknownIds` after each receive, but when testing we found that removing the values introduced a bigger overhead than what was gained by sending fewer values.

### DIFF
--- a/core/vtk/ttkGhostCellPreconditioning/ttkGhostCellPreconditioning.cpp
+++ b/core/vtk/ttkGhostCellPreconditioning/ttkGhostCellPreconditioning.cpp
@@ -43,6 +43,18 @@ int ttkGhostCellPreconditioning::FillOutputPortInformation(
   return 0;
 }
 
+// returns true if bounding boxes intersect, false if not
+bool checkForIntersection(double *myBB, double *theirBB) {
+  return !(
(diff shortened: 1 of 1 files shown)


julien-tierny:
ok, looks good to me!

julien-tierny:
@m-s-will a side comment:
could you extend the GhostCellPreconditioning to also produce a RankArray but for the cells?

m-s-will:
Sure, I just checked, the CellData also includes vtkGhostLayer values and generateGlobalIds also generates unique global cell ids, so doing everything done for vertices also to cells should be no problem.

PR #789, opened 2022-06-15
[MPI] Add RankArray to MPI preconditioning

eve-le-guillou:
This Pull Request improves the preconditioning for MPI by automatically computing ghost points, ghost cells, global ids and the RankArray array if they are not already computed when the execution is in parallel.

The MPI preconditioning is moved from the ProcessRequest function to the GetTriangulation function in the ttkAlgorithm class.
Two functions are added to ttkAlgorithm:
- MPIPipelinePreconditioning: before the creation/retrieval of a triangulation, ghosts, global ids and the RankArray array are added to the pipeline if not already present.
- MPITriangulationPreconditioning: after the creation/retrieval of a triangulation, informations for ghosts, global ids and the RankArray array are retrieved from the pipeline and added to TTK objects. The vertex preconditioning is executed here as well.

In order to be able to produce the RankArray from within ttkAlgorithm, the heart of the TTKGhostCellPreconditioning filter is moved to MPIUtils.h. The filter is still there and still allows of the computation of the RankArray array.

To execute a filter in parallel, such as TTKArrayPreconditioning and ScalarFieldCriticalPoints, one would now need the following pipeline:

        from paraview.simple import *
        
        # create a new 'Wavelet'
        
        # create a new 'TTK ArrayPreconditioning'
        tTKArrayPreconditioning1 = TTKArrayPreconditioning(registrationName='TTKArrayPreconditioning1', Input=wavelet1)
        
        # create a new 'TTK ScalarFieldCriticalPoints'
        tTKScalarFieldCriticalPoints1 = TTKScalarFieldCriticalPoints(registrationName='TTKScalarFieldCriticalPoints1', Input=tTKArrayPreconditioning1)
        tTKScalarFieldCriticalPoints1.ScalarField = ['POINTS', 'RTData']
        tTKScalarFieldCriticalPoints1.InputOffsetField = ['POINTS', 'RTData']
        tTKScalarFieldCriticalPoints1.Backend = 'Default generic backend'
        
        UpdatePipeline(time=0.0, proxy=tTKScalarFieldCriticalPoints1)
      
Please note that the computation of the global ids in parallel will raise a warning as it won't produce a result strictly identical to the result on one process in terms of global identifiers.

### DIFF
--- a/core/base/common/MPIUtils.h
+++ b/core/base/common/MPIUtils.h
@@ -325,7 +325,8 @@ namespace ttk {
                                unsigned char *ghostCells,
                                int nVertices,
                                double *boundingBox) {
-    std::vector<std::array<double, 6>> rankBoundingBoxes(ttk::MPIsize_);
+    std::vector<std::array<double, 6>> rankBoundingBoxes(
+      ttk::MPIsize_, std::array<double, 6>({}));
     std::copy(
       boundingBox, boundingBox + 6, rankBoundingBoxes[ttk::MPIrank_].begin());
     for(int r = 0; r < ttk::MPIsize_; r++) {
--- a/core/base/common/MPIUtils.h
+++ b/core/base/common/MPIUtils.h
@@ -9,6 +9,7 @@
 
 #include <BaseClass.h>
 #include <Timer.h>
+#include <algorithm>
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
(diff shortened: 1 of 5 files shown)


julien-tierny:
thanks for fixing the conflict :)

julien-tierny:
ok, looks good, thanks!

PR #790, opened 2022-06-16
PersistenceDiagramClustering: Reset diagram position before translation

pierre-guillou:
This PR introduces new methods in `ttkPersistenceDiagramUtils` that take care of translating a VTU diagram (in its canonical form) and resetting its position. 

This way, it is possible to apply the `PersistenceDiagramClustering` filter on the output of a previous `PersistenceDiagramClustering` instance, using the `Clusters as stars` representation, without messing with the diagram positions.

Enjoy,
Pierre


julien-tierny:
looks good, thanks!

PR #791, opened 2022-06-20
[MPI] Add rank array computation for cell data

m-s-will:
This PR adapts the Rank Array computation for point data to cell data. Cells also get unique ids and the vtkGhostType array, therefore we can use the existing methods in `MPIUtils.h`.

The RankArray for cells is automatically added in `MPIPipelinePreconditioning` in `ttkAlgorithm.cpp`, but can also be manually added by using the `GhostCellPreconditioning` filter.


julien-tierny:
alright, looks good!
thanks

PR #792, opened 2022-06-23
[MergeTreeDistanceMatrix] fix division by zero

MatPont:
When computing a distance matrix of an ensemble of less than 10 merge trees, a modulo operation (here to print the progress of the computation) causes a division by zero.

This PR fixes this problem.


julien-tierny:
neat & tidy!

PR #793, opened 2022-06-24
Add DiscreteMorseSandwich, PersistentGenerators

pierre-guillou:
This PR adds the new DiscreteMorseSandwich backend to the PersistenceDiagram module. Also included in this PR is the PersistentGenerators module.

This new backend has been set to be the default one for the PersistenceDiagram filter and standalone. This will cause errors in the CI but a later ttk-data PR should revert the PersistenceDiagram backend of relevant state files & Python scripts to FTM.

Some tweaks in the DiscreteGradient module should help it being more resilient when called inside a parallel region (the cache is disabled) or when the `preconditionTriangulation` and the `buildGradient` methods are not called from the same DiscreteGradient instance.

Enjoy,
Pierre

### DIFF
--- a/core/base/discreteGradient/DiscreteGradient.h
+++ b/core/base/discreteGradient/DiscreteGradient.h
@@ -376,8 +376,7 @@ according to them.
        */
       inline void preconditionTriangulation(AbstractTriangulation *const data) {
         if(data != nullptr) {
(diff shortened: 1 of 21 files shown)


julien-tierny:
awesome, let's go!

PR #794, opened 2022-06-24
Improve FlatJaggedArray API

pierre-guillou:
This PR extends the FlatJaggedArray API to support double indexing (`array[i][j]`) and range-based for loops (`for(auto &vec: array) {for(auto elem: vec){}}`) and make it closer to the `std::vector<std::vector<SimplexId>>` API.

The ExplicitTriangulation now makes extensive use of this new API (which is quite close to what was there before the introduction of FlatJaggedArray), without any noticeable performance impact (tested with DiscreteGradient).

Enjoy,
Pierre

### DIFF
--- a/core/base/common/FlatJaggedArray.h
+++ b/core/base/common/FlatJaggedArray.h
@@ -29,6 +29,14 @@ namespace ttk {
       this->offsets_ = std::move(offsets);
     }
 
+    /**
+     * @brief Clear the underlaying vectors
+     */
+    inline void clear() {
+      this->data_.clear();
+      this->offsets_.clear();
+    }
+
     // ############################## //
     // Mimic the vector of vector API //
     // ############################## //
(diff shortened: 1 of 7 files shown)


core/base/common/FlatJaggedArray.h
```diff
@@ -57,6 +65,61 @@ namespace ttk {
...
+      inline SimplexId operator[](const SimplexId id) {
+#ifndef TTK_ENABLE_KAMIKAZE
+        if(id < 0 || id >= len) {
+          return -1;
+        }
+#endif
+        return ptr[id];
+      }
+      inline const SimplexId *data() const {
+        return ptr;
```
vollia: missing `this` here and below

CharlesGueunet:
Is there any documentation on which methods are thread safe and which are not ?

core/base/common/FlatJaggedArray.h
pierre-guillou: Thanks! I miss a corresponding clang-tidy lint.

pierre-guillou:
> Is there any documentation on which methods are thread safe and which are not ?

All newly-introduced methods should be const except `Iterator::operator++` which IIRC is used by range-based for loops. I don't see a use-case in which this method could be called in a parallel region without abusing the API.

julien-tierny:
nice!

PR #795, opened 2022-06-27
MorseSmaleComplex: Display CellDimension, SeparatrixType by default

pierre-guillou:
This PR uses `vtkDataSetAttributes::SetScalars` to set the default scalar field for DiscreteGradient critical points (CellDimension), DiscreteGradient glyphs (PairType), Morse-Smale Complex critical points (CellDimension) and Morse-Smale Complex 1-separatrices (SeparatrixType). The FTMTree filter already makes use of these methods to provide a default coloring of nodes & arcs.

Since the filter's output are modified, changes in the Python CI hashes are expected.

Enjoy,
Pierre


julien-tierny:
looks good, thanks!

PR #796, opened 2022-06-27
kdTree: Remove dynamic allocation

pierre-guillou:
This PR replaces a dynamic allocation occuring in a hot path in the KDTree module with equivalent code that computes the index of the maximum cost in the range [|0, k|].

This modification increases the performance of the PersistenceDiagramClustering & PersistenceDiagramDistanceMatrix by up to +40%.

Since more computation is available for a given time period, the PersistenceDiagramClustering state file output goes one iteration further (screenshots have to/will be updated).

Enjoy,
Pierre


julien-tierny:
ok, thanks!

julien-tierny:
ok, thanks!

PR #797, opened 2022-06-27
MorseSmaleQuadrangulation: Remove excess points when DualQuadrangulate

pierre-guillou:
The DualQuadrangulation option of the MorseSmaleQuadrangulation filter generates a coarser quadrangulation from the Morse-Smale Complex. However, it kept the same points as the non-dual quadrangulation even though some of them have no neighbors in the dual quadrangulation.

This PR post-processes the dual quadrangulation, removing these points that have no neighbor. These additional points would have been projected to coordinates (nan, nan, nan) by the QuadrangulationSubdivision filter.

Hashes of the morseSmaleQuadrangulation.py Python script are expected to change (200 points removed).

Enjoy,
Pierre


julien-tierny:
thanks!
can you also PR updated hashes on ttk-data?

PR #798, opened 2022-06-28
Extension to an old module of mine

kappe-c:

In a nutshell, I use the information of all the vertices within the contours to compute a centroid whose position is less arbitrary compared to the original extreme point if it lies in a "flat" "plateau" region, and which is (because of that) more stable for scalar fields with a temporal dimension.
I have checked my example in the data repo and it still works. There is now a second output available for TTKContourAroundPoint filter instances (Centroids additional to the old Contours) which can be visualized e.g. with TTKIcospheresFromPoints.
(That actually happens to be a rather good example; notice my new dark blue/red centroids that consider both topology and geometry and lie nicely centrally within their contours compared to the original extreme points.)
Screenshot_20220628_111859 ContourAroundPoint


kappe-c:
Hello,
I am sorry but currently I have difficulty figuring out what is the problem here
https://github.com/topology-tool-kit/ttk/runs/7091323879?check_suite_focus=true#step:16:17551
line 17410
Failing cases: ['mergeTreeClustering', 'mergeTreeTemporalReduction']
It is not obvious that this even has anything to do with code that I touched.
Regards,
Chris

julien-tierny:
I don't believe this is the issue.
Let me restart the CI.

julien-tierny:
alright, looks good, thanks!

PR #799, opened 2022-06-29
SurfaceGeometrySmoother: Re-use container between iterations

pierre-guillou:
This PR includes a small change in the `SurfaceGeometrySmoother` base module that reduces dynamic allocations by sharing a heap-allocated container between (parallel) loop iterations (one container per thread). The goal is to re-use the buffer between iterations (by the same thread), thus reducing the number of heap allocations.

Enjoy,
Pierre


julien-tierny:
great, thanks!

PR #800, opened 2022-06-30
Fix GCC12 warnings, remove explicit O3 flag

pierre-guillou:
This PR fixes several GCC 12 warnings (-Wuse-after-free, -Warray-bounds, -Wuseless-cast, -Wmaybe-uninitialized, -Wnull-dereference, -Wduplicated-branches). This should hopefully make TTK safer.

In addition, a `-O3` flag has been removed in BaseCode.cmake that prevented users to override the optimization level for individual TTK targets (using `target_compile_options` in modules's CMakeLists.txt), which is useful for debugging modules without recompiling the whole TTK (and still keeping some level of performance).

Enjoy,
Pierre


julien-tierny:
looks good, thanks!

PR #801, opened 2022-06-30
[MPI] Automatically create order array if it does not exist

m-s-will:
Hello,

currently one needs to run ArrayPreconditioning before trying to use the order array of any scalar array. GetOrderArray actually tries to generate an order array if it is not found, however this is only correct for each rank, not between ranks. 

Therefore I needed to move much of ArrayPreconditioning to MPIUtils (similarly to the GhostCellPreconditioning), so that it can be used in ttkAlgorithm to generate an order array on the fly. Now any MPI filters should be usable without using any filters manually adding GlobalIds, GhostCells, the RankArray or an OrderArray.


julien-tierny:
hi @m-s-will 

can you please re-open this PR towards the `dev` branch (the legacy `mpi` branch has been deleted after merging into the `dev` branch)

also, we noticed that you used some copy based parameter specifications. for instance `core/base/explicitTriangulation/ExplicitTriangulation.h:572`, this function makes a full copy of a potentially very large map. in scalar field smoother, such a copy is done at each iteration. first it does not need to, it could be done out of the for loop. second, you'd better do a (const) ref copy or (const) pointer copy to speed things up.
also, the new optional parameters of scalar field smoother are not necessary (they can now be retrieved from the triangulation).

can you please double-check in other parts of your commits that you don't use any unnecessary parameter copy?

thanks a lot!

PR #802, opened 2022-07-01
 [MPI] Automatically create order array if it does not exist

m-s-will:
Hello,

this is a reopen of #801 with dev as a target. An optimization PR will come early next week.

currently one needs to run ArrayPreconditioning before trying to use the order array of any scalar array. GetOrderArray actually tries to generate an order array if it is not found, however this is only correct for each rank, not between ranks.

Therefore I needed to move much of ArrayPreconditioning to MPIUtils (similarly to the GhostCellPreconditioning), so that it can be used in ttkAlgorithm to generate an order array on the fly. Now any MPI filters should be usable without using any filters manually adding GlobalIds, GhostCells, the RankArray or an OrderArray.


julien-tierny:
please `git pull` from the `dev` branch to make sure you're on sync.

julien-tierny:
looks good, thanks!

PR #803, opened 2022-07-01
Update BottleneckDistance with new PersistenceDiagram data structure

pierre-guillou:
This PR uses the data structures & functions added in #745 to overhaul the BottleneckDistance module & its two downstream users, TrackingFromFields and TrackingFromPersistenceDiagrams.

Replacing templated tuples with structs makes the code more readable and allows the de-templating of the base layer, reducing the build times. 

The BottleneckDistance VTK layer has been modified to make it closer to PersistenceDiagramClustering/PersistenceDiagramDistanceMatrix. The filter now requires a vtkMultiBlockDataSet of PersistenceDiagrams as an input (distance is computed between the first two blocks). Diagram translation alongside the Z-axis should be identical to PersistenceDiagramClustering: the same user-provided spacing value should generate the same translation. The diagrams position is also reset before the translation, following what was done in #790. The output diagrams and matching also follow what has been done in PersistenceDiagramClustering: one output for the (translated) diagrams, one output for the matchings.

The results displayed on the command-line should be identical and are now broken down per pair type and also propagated on the Matchings output as FieldData.

Enjoy,
Pierre


julien-tierny:
ok, looks good.
(need to double check mathieu's merge tree examples)

PR #804, opened 2022-07-05
Use SetThresholdFunction instead of the old ThresholdBy*

CharlesGueunet:
Dear Julien,

We have a warning concerning a deprecated method `vtkThreshold::ThresholdBy*` for some times now. Unfortunately, this has evolved into an error with the latest ParaView has this method has been removed.
Here is the MR to use the new version.

Best,  
Charles

### DIFF
--- a/core/vtk/ttkExtract/ttkExtract.cpp
+++ b/core/vtk/ttkExtract/ttkExtract.cpp
@@ -1,16 +1,15 @@
 #include "BaseClass.h"
 #include <ttkExtract.h>
 
-#include <vtkInformationVector.h>
-
-#include <vtkMultiBlockDataSet.h>
-#include <vtkSmartPointer.h>
-#include <vtkTable.h>
-#include <vtkUnstructuredGrid.h>
-
 #include <vtkInformation.h>
 #include <vtkInformationVector.h>
+#include <vtkSmartPointer.h>
 #include <vtkStreamingDemandDrivenPipeline.h>
+#include <vtkVersion.h> // for VTK_VERSION_CHECK via ParaView 5.8.1
+
+#include <vtkMultiBlockDataSet.h>
+#include <vtkTable.h>
+#include <vtkUnstructuredGrid.h>
 
 #include <vtkCellArray.h>
 #include <vtkCellData.h>
@@ -496,7 +495,12 @@ int ttkExtract::ExtractGeometry(vtkDataObject *output,
     threshold->SetInputDataObject(maskOutput);
     threshold->SetInputArrayToProcess(
       0, 0, 0, isPointDataArray ? 0 : 1, "Mask");
+#if VTK_VERSION_NUMBER < VTK_VERSION_CHECK(9, 2, 0)
     threshold->ThresholdByUpper(0.5);
+#else
+    threshold->SetThresholdFunction(vtkThreshold::THRESHOLD_UPPER);
+    threshold->SetUpperThreshold(0.5);
+#endif
     threshold->SetAllScalars(this->CellMode == CELL_MODE::ALL);
     threshold->Update();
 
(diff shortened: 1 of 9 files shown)


CharlesGueunet:
This MR fails on elder PV / VTK. Do we want to branch the code depending on the VTK / PV version ? or do we want to stay compatible with VTK >= 9.1 and PV >= 5.10 ?

julien-tierny:
can you put preprocessor macro safeguard for the moment (such that the compiler switches between the old and new version, based on the paraview version we're building against)

core/vtk/ttkExtract/ttkExtract.cpp
```diff
@@ -496,7 +496,12 @@ int ttkExtract::ExtractGeometry(vtkDataObject *output,
     threshold->SetInputDataObject(maskOutput);
     threshold->SetInputArrayToProcess(
       0, 0, 0, isPointDataArray ? 0 : 1, "Mask");
+#if VTK_MAJOR_VERSION > 9 || (VTK_MAJOR_VERSION >= 9 && VTK_MINOR_VERSION >= 2)
```
mathstuf: There is `VTK_VERSION_CHECK` to do this. The replacement has been there since 9.1

vollia: indeed, that is cleaner. Fixed, thanks !

mathstuf: More specifically:

```cpp
#if VTK_VERSION_NUMBER < VTK_VERSION_CHECK(9, 1, 0)
  // old code
#else
  // new code
#endif
```

vollia: what is the benefits of this order compared to mine ?
I used:
```cpp
#if VTK_VERSION_NUMBER >= VTK_VERSION_CHECK(9, 2, 0)
    // new
#else
    // old
#endif
```

vollia: Unfortunately, it seems the **VTK** provided by ParaView 5.8.1 does not yet support this feature. The CI complains:
```
/home/vsts/work/1/s/core/vtk/ttkExtract/ttkExtract.cpp:498:44: error: missing binary operator before token "("
 #if VTK_VERSION_NUMBER >= VTK_VERSION_CHECK(9, 2, 0)
```

vollia: @julien-tierny do we still need to support ParaView 5.8.1 ? Or do we have a plan to move forward soon ?

mathstuf: The benefits are that the new API actually comes with 9.1, so you check that version, not against 9.2.

As for ParaView 5.8.1, the macro lived in `vtkVersion.h` then. Adding this at the top would work:

```cpp
#include "vtkVersion.h" // for VTK_VERSION_CHECK via ParaView 5.8.1
```

with a hope that it gets noticed when the macro is no longer needed for 5.8.1 support is dropped.

vollia: Your version indeed work with PV 5.8.1, this MR should be ready once the pipeline is green :)

julien-tierny:
could you please also update the source `examples/vtk-c++/main.cpp`?
is the pvpython api impacted? (we already observed a switch from 5.9 to 5.10)

CharlesGueunet:
It seems the `packaging` package is not present on Mac. If my understanding this is the standard python way to check for version, would it be possible to depends on it / install it on the CI ? I tried updating the CI myself with no avail

julien-tierny:
it looks like you forgot the `packaging` package under windows.
could you fix that really quick?
thanks!

julien-tierny:
you may need to git pull first to synchronize

julien-tierny:
can you also add the `packaging` dep in the app-veyor CI?
thanks!
otherwise, I believe we can get rid of it at this point (superseded by the github actions).

julien-tierny:
alright, the CI is no longer complaining now.
thanks!

PR #805, opened 2022-07-05
[MPI] Optimization in VertexGlobalIdMap and ScalarFieldSmoother

m-s-will:
This PR changes some copy based parameters to only return pointers, moves the vertexGlobalIdMap in ScalarFieldSmoother out of the loop and removes some unneeded parameters which can be retrieved from the triangulation instead. Due to the changes in `ttkAlgorithm.cpp` which added `MPIPipelinePreconditioning`, this filter can now be used on loaded data directly, without any manual preprocessing.
On my machine this leads to a speedup of factor 3-5, when smoothing a scalar field with 1.000.000 vertices which is smoothed in 100 iterations.

### DIFF
--- a/core/base/scalarFieldSmoother/ScalarFieldSmoother.h
+++ b/core/base/scalarFieldSmoother/ScalarFieldSmoother.h
@@ -53,9 +53,7 @@ namespace ttk {
 
     template <class dataType, class triangulationType = AbstractTriangulation>
     int smooth(const triangulationType *triangulation,
-               const int &numberOfIterations,
-               const int *rankArray = nullptr,
-               const SimplexId *globalIds = nullptr) const;
+               const int &numberOfIterations) const;
 
   protected:
     int dimensionNumber_{1};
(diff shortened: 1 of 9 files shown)


core/base/scalarFieldSmoother/ScalarFieldSmoother.h
```diff
@@ -84,14 +80,30 @@ int ttk::ScalarFieldSmoother::smooth(const triangulationType *triangulation,
   if(!outputData_)
     return -4;
 #endif
+  int *rankArray_{nullptr};
```
julien-tierny: the current convention for variable naming in the base code is to use a final `_` only for member variables (precisely to quickly identify them). could you remove this one and the next (these are local variables).

core/base/scalarFieldSmoother/ScalarFieldSmoother.h
```diff
@@ -84,14 +80,30 @@ int ttk::ScalarFieldSmoother::smooth(const triangulationType *triangulation,
...
 
 #if TTK_ENABLE_MPI
-  if(ttk::isRunningWithMPI() && rankArray != nullptr && globalIds != nullptr)
+  rankArray_ = triangulation->getRankArray();
+  globalIds_ = (SimplexId *)triangulation->getGlobalIdsArray();
+  if(ttk::isRunningWithMPI() && rankArray_ != nullptr
+     && globalIds_ != nullptr) {
     useMPI = true;
+    map = triangulation->getVertexGlobalIdMap();
+    if(map != nullptr) {
```
julien-tierny: I believe this block can be removed.

julien-tierny (changes requested):
thanks! 
could you please quickly address my couple of minor comments?
thanks!

julien-tierny:
awesome, thanks!

PR #807, opened 2022-07-07
GitHub Actions: Add Ubuntu 22.04, switch to Windows Server 2022

pierre-guillou:
This PR updates the GitHub Actions workflows `check.yml`, `test.yml`, `ccache.yml` and `package.yml` to support Ubuntu 22.04 as an additional platform (for `test.yml`, `ccache.yml` and `package.yml`) or as a replacement of Ubuntu 20.04  in `check.yml`. The Windows platforms are also updated from Windows Server 2019/Visual Studio 2019 to Windows Server 2022/Visual Studio 2022.

This PR also contains the patch that corresponds to https://github.com/topology-tool-kit/ttk-paraview/pull/6.

A few compatibility bugs were fixed in this PR:
* a Doxygen update in Ubuntu 22.04 found an error in `FTMTreeUtils.cpp`,
* the DimensionReduction module needs an absolute path to the `dimensionReduction.py` Python script at compile time but, in order to create packages, `cpack` needs only relative paths.
* the `DOUBLE_TEMPLATING` feature has been disabled in the Ubuntu and Windows jobs of the `package.yml` workflow since it was using too much RAM.
* the `test.yml` workflow can now work even without ccache archive generated by `ccache.yml`,
* Ubuntu 22.04 has a weird behavior with its OOM killer and I had to skip the execution of `mergeTreeClustering.pvsm`,
* the `persistenceDiagramClustering.pvsm` state file does more iterations in Ubuntu 22 than in previous Ubuntu versions (maybe related to #796). As a consequence, the screenshots are not tested.
* Ubuntu 22.04's `pvpython` does not embed the correct PYTHONPATH (the standard Python interpreter should work fine).

Additionally, the ttk-data Python scripts are also tested in the `package.yml` workflow.

Enjoy,
Pierre


julien-tierny:
awesome, looks great!

PR #810, opened 2022-07-13
Fix CI after new state files additions

pierre-guillou:
Following the introduction of new state files in ttk-data (in particular relative to `PersistentGenerators`), this PR skips running some memory-overflowing state files in Ubuntu 22.04 that may crash the whole workflow.

Additionally, an out-of-bound memory read has been fixed in `PersistentGenerators`: the wrong boolean vector was passed to the `DiscreteMorseSandwich` module in dimension 2, leading to reads of garbage memory. This was causing non-deterministic outputs in the relevant Python scripts.

Enjoy,
Pierre


julien-tierny:
great stuff, thanks!

PR #811, opened 2022-07-15
[MPI] Fixed bug when completing onesided neighbors

m-s-will:
Hello,

this PR fixes an oversight when calculating neighborhood relations between ranks. Depending on the number of ghost layers, ranks can have onesided relations e.g. rank A knows that it needs information from rank B, but rank B does not know that it needs to send information to rank A. There was a bug in the code calculating this, which is now fixed.


julien-tierny:
awesome, thanks!

PR #812, opened 2022-07-21
Fix build without MPI

lhofmann:
`ttkAlgorithm` currently depends on `VTK::FiltersParallelDIY2`, but this dependency is only required for MPI builds and may not be available if VTK was built without MPI.

This PR makes the dependency optional, which allows the VTK module system to silently skip linking against `VTK::FiltersParallelDIY2` if it is not available.
If the module is not required (building without MPI) but available, it is linked without being used, but I think this is the most simple solution.

See https://github.com/conda-forge/topologytoolkit-feedstock/pull/28

### DIFF
--- a/CMake/VTKModule.cmake
+++ b/CMake/VTKModule.cmake
@@ -117,7 +117,7 @@ function(ttk_get_target ttk_module ttk_target)
 
   # dependencies check
   ttk_parse_module_file(${VTKWRAPPER_DIR}/${ttk_module}/vtk.module)
-  cmake_parse_arguments("VTK" "" "NAME" "DEPENDS;PRIVATE_DEPENDS" ${moduleFileContent})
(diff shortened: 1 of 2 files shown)


julien-tierny:
thanks a lot for that hot fix @lhofmann !!!
@m-s-will are you sure the DIY dep is needed here (???!!!???)

m-s-will:
> @m-s-will are you sure the DIY dep is needed here (???!!!???)

@eve-le-guillou added this dependency when introducing the MPIPipelinePreconditioning (in this PR #776 , no idea why it was shown that I added it again..) and I think it is needed for the GhostCellsGenerator and the GenerateGlobaIds VTK filters

eve-le-guillou:
> > @m-s-will are you sure the DIY dep is needed here (???!!!???)
> 
> @eve-le-guillou added this dependency when introducing the MPIPipelinePreconditioning (in this PR #776 , no idea why it was shown that I added it again..) and I think it is needed for the GhostCellsGenerator and the GenerateGlobaIds VTK filters

Yes, that is exactly right. The dependency is required to call the GhostCellsGenerator filter from inside ttkAlgorithm.

PR #813, opened 2022-07-23
Add a CMake preset that can be use to release or help benchmark

CharlesGueunet:
Dear Julien,

This PR is a first attempt at a `CMakePreset.json`.
The goal is to help people compiling TTK to have the right default values for some variable, like the Kamikaze mode on when trying to benchmark. It may also help having a Release mode that would give users the same configuration than the release, for reproducibility or simply for testing.

All the variable I have added here are subject to discussion, so feel free to tell me for better defaults.

Best,
Charles

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,41 @@
+{
+  "version": 2,
+  "configurePresets": [
+    {
+      "name": "TTK-Default",
+      "hidden": true,
+      "binaryDir": "build",
+      "generator": "Ninja",
+      "cacheVariables": {
+        "TTK_ENABLE_DOUBLE_TEMPLATING": "ON"
+      }
+    },
+    {
+      "name": "TTK-Release",
+      "inherits": "TTK-Default",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Release",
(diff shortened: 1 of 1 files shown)


julien-tierny:
hi charles,
that sounds like a good idea, thanks.
would that file be automatically processed by CMAKE? (i.e. creating a new entry in ccmake or cmake-gui with a dropdown menu with the available options?)

CharlesGueunet:
Hi Julien,

there won't be a specific entry in `ccmake` or `cmake-gui` yet. For now on, the preset are only available with the `cmake` command:
```bash
cmake --preset=TTK-Perf #for example
```
Preset is a young feature and the integration inside the configuring tools of cmake may evolve in the future.
Best,  
Charles

julien-tierny:
actually, I have just given it a shot, and with cmake 3.22.4, the preset selection menu does appear indeed in cmake-gui :)

CharlesGueunet:
Oh cool ! that is a good news :) now, there will be no excuse to omit an important option while benchmarking against TTK !

julien-tierny:
:laughing: 
:rofl: 
never underestimate the depth of bad faith :)

PR #814, opened 2022-07-27
Triangulation: Improve MPI support

pierre-guillou:
This PR reworks the MPI support inside the Triangulation modules to enable full local to global and global to local identifiers mappings for edges and triangles. It extends #773 by exchanging missing identifier mappings between neighboring ranks. The algorithm introduced in #773 to label local edges & triangles with global identifiers has been overhauled and is now better MPI-parallel.

This new additions only work well in practice for `ImplicitTriangulation`. VTU files are currently not distributed correctly between ranks by ParaView: applying `Ghost Cells Generator` after `Redistribute Dataset` generates duplicate points that share the same global id in the same rank (implicit grids are not affected).

MPI support has been removed from `PeriodicImplicitTriangulation` since the periodic triangulation model does not fit the data distribution done by ParaView. Similarly, the few MPI support in `CompactTriangulation` has been removed (no idea in how to implement it in practice). An error message is displayed in `TriangulationManager` when trying to convert to these two Triangulation types in an MPI context.

`AbstractTriangulation` now holds the accessor methods to convert between local and global indices (except for local to global cell indices for implicit grids). Useless triangulation members have been removed (`ghostCellMask_` has been replaced by `cellRankArray_`, `vertexLidToGid_` is a copy of `vertGid_`...). The global identifiers arrays pointers types (`cellGid_` and `vertGid_`) have been switched from `long int` to `LongSimplexId`/`vtkIdType` to better match the output of the `Generate Global Ids` ParaView filter. This has led to some type modifications in `MPIUtils.h` methods, I hope without consequences.

The `ScalarFieldCriticalPoints` and `ScalarFieldSmoother` modules have been marginally modified to reduce the MPI code footprint and better take advantage of the new triangulation methods. Both should work as expected.

Enjoy,
Pierre

### DIFF
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -2555,17 +2555,7 @@ namespace ttk {
       return this->cellRankArray_;
     }
 
-    // precondition methods for distributed meshes
-
-    virtual int preconditionDistributedCells() {
-      return 0;
-    }
-    virtual int preconditionDistributedEdges() {
-      return 0;
-    }
-    virtual int preconditionDistributedTriangles() {
-      return 0;
-    }
+    // public precondition method (used in Triangulation/ttkAlgorithm)
     virtual int preconditionDistributedVertices() {
       return 0;
     }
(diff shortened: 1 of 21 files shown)


julien-tierny:
awesome, looks great!
@eve-le-guillou @m-s-will you may want to have a close look at this PR, to make sure we're all on the same page.

PR #816, opened 2022-07-29
Branch Mapping Distance and Path Mapping Distance for Merge Trees

floWetzels:
TAdded implementation of different branch decomposition-independent edit distances for merge trees to the MergeTreeDistanceMatrix module. Also changed some details in the MergeTreeClustering module, so that it actively ignores the new distances (for now).


julien-tierny:
thanks a lot @floWetzels 
coud you take care of the CI errors? thanks!

floWetzels:
> thanks a lot @floWetzels coud you take care of the CI errors? thanks!

Sure, working on it.

core/base/mergeTreeClustering/BranchMappingDistance.h
```diff
@@ -0,0 +1,522 @@
...
+      size_t dim2 = (nn1 + 1) * dim1;
+      size_t dim3 = (depth1 + 1) * dim2;
+      size_t dim4 = (nn2 + 1) * dim3;
+      // size_t dim1_ = 1;
+      // size_t dim2_ = (nn1 + 1) * dim1;
+      // size_t dim3_ = (nn1 + 1) * dim2;
+      // size_t dim4_ = (nn2 + 1) * dim3;
+
+      std::unique_ptr<dataType[]> memT(
+        new dataType[(nn1 + 1) * (depth1 + 1) * (nn2 + 1) * (depth2 + 1)]);
```
julien-tierny: please re-factor to avoid the usage of any `new` and `delete`.

core/base/mergeTreeClustering/PathMappingDistance.h
```diff
@@ -0,0 +1,360 @@
...
+      size_t dim2 = (nn1 + 1) * dim1;
+      size_t dim3 = (depth1 + 1) * dim2;
+      size_t dim4 = (nn2 + 1) * dim3;
+      // size_t dim1_ = 1;
+      // size_t dim2_ = (nn1 + 1) * dim1;
+      // size_t dim3_ = (nn1 + 1) * dim2;
+      // size_t dim4_ = (nn2 + 1) * dim3;
+
+      std::unique_ptr<dataType[]> memT(
+        new dataType[(nn1 + 1) * (depth1 + 1) * (nn2 + 1) * (depth2 + 1)]);
```
julien-tierny: please re-factor to avoid the usage of any `new` and `delete`.

core/base/mergeTreeClustering/BranchMappingDistance.h
```diff
@@ -0,0 +1,522 @@
...
+#include <AssignmentExhaustive.h>
+#include <AssignmentMunkres.h>
+#include <Debug.h>
+
+namespace ttk {
+
+  class BranchMappingDistance : virtual public Debug {
+
+  private:
+    int baseMetric = 0;
```
julien-tierny: usually, in ttk, the member variables of a class, within the `core/base/` subdirectories, end with a final `_` (in order to distinguish them from local variables, later in the functions).
see `assignmentSolverID_` for instance.
that'd be great if you could follow that convention.

core/base/mergeTreeClustering/PathMappingDistance.h
```diff
@@ -0,0 +1,360 @@
...
+/// \author Florian Wetzels (wetzels@cs.uni-kl.de)
+/// \date 2022.
+///
+/// This module defines the %PathMappingDistance class that computes distances
+/// between two merge trees.
+///
+/// \b Related \b publication \n
+/// "A Deformation-based Edit Distance for Merge Trees" \n
+/// Florian Wetzels, Christoph Garth.\n
+/// TBP.
```
julien-tierny: Thanks for updating the reference ;)

core/base/mergeTreeClustering/PathMappingDistance.h
```diff
@@ -0,0 +1,360 @@
...
+#include <AssignmentExhaustive.h>
+#include <AssignmentMunkres.h>
+#include <Debug.h>
+
+namespace ttk {
+
+  class PathMappingDistance : virtual public Debug {
+
+  private:
+    int baseMetric = 0;
```
julien-tierny: usually, in ttk, the member variables of a class, within the core/base/ subdirectories, end with a final _ (in order to distinguish them from local variables, later in the functions).
see assignmentSolverID_ for instance.
that'd be great if you could follow that convention.

core/vtk/ttkMergeTreeDistanceMatrix/ttkMergeTreeDistanceMatrix.cpp
```diff
@@ -229,3 +309,113 @@ int ttkMergeTreeDistanceMatrix::RequestData(
...
+  }
+  for(size_t i = 0; i < upIDs.size(); i++) {
+    int downId = downIDs[i];
+    int upId = upIDs[i];
+    children[upId].push_back(downId);
+    nodePersistences[downId]
+      = persistences[i]; // std::abs(nodeScalars[downId]-nodeScalars[upId]);
+  }
+
+  // random edge contractions for testing
```
julien-tierny: could you remove this commented block?

julien-tierny:
hi @floWetzels , 
sorry for the delay on this PR.
I started a review. Thanks for considering my remarks (especially regarding the usage of `new`/`delete` to be avoided, as well as naming conventions).
@MatPont could you have a quick look at the code too to make sure that everything's fine on your end?
thanks!

julien-tierny (changes requested):
Thanks for the PR. Thanks for considering the following remarks

core/vtk/ttkMergeTreeDistanceMatrix/ttkMergeTreeDistanceMatrix.cpp
```diff
@@ -229,3 +309,113 @@ int ttkMergeTreeDistanceMatrix::RequestData(
 
   return res;
 }
+
+std::tuple<std::vector<float>, std::vector<std::vector<int>>, int>
+  ttkMergeTreeDistanceMatrix::ftmToAdjList(
```
MatPont: Would not it be better to make this function returns void and put the actual output as an input parameter by reference to avoid copy?

julien-tierny: jeeze, good catch.
please do not return objects in general and prefer passing the output of your function by reference (as a parameter) as recalled by mathieu. returning outputs directly induces a copy, which can slow everything down when the object is large.

julien-tierny:
@pierre-guillou if you could also have a look, that'd be fantastic :)

core/base/mergeTreeClustering/PathMappingDistance.h
pierre-guillou: I agree. Why not use the commented `std::vector` declaration below instead?

core/vtk/ttkMergeTreeDistanceMatrix/ttkMergeTreeDistanceMatrix.cpp
pierre-guillou: Actually, compilers can use [Copy Elision](https://en.cppreference.com/w/cpp/language/copy_elision) to avoid copies in return statements (aka "Return Value Optimization").

core/base/mergeTreeDistanceMatrix/MergeTreeDistanceMatrix.h
```diff
@@ -55,6 +74,58 @@ namespace ttk {
       }
     }
 
+    template <class dataType>
+    void execute(
+      std::vector<
+        std::tuple<std::vector<float>, std::vector<std::vector<int>>, int>>
+        &trees,
```
pierre-guillou: These long types can be factored with [type aliases](https://en.cppreference.com/w/cpp/language/type_alias) (`using VectorTypeName = std::vector<std::tuple<std::vector<float>, std::vector<std::vector<int>>, int>>` or `using TupleTypeName = std::tuple<std::vector<float>, std::vector<std::vector<int>>, int>`).
Types aliases also make it easier to document a complex type.

Alternatively, I would replace the tuple with a struct because it's easier to understand named field members (with documentation) than `std::get<n>()` calls.

core/vtk/ttkMergeTreeDistanceMatrix/ttkMergeTreeDistanceMatrix.cpp
julien-tierny: is that enabled by default on all compilers?

pierre-guillou: It is explicitly required from C++17 and above, but it should be safe to assume that GCC, Clang and MSVC support it for C++11. 

[Wikipedia](https://en.wikipedia.org/wiki/Copy_elision#RVO) tells us that 

> Return value optimization is supported on most compilers.

core/base/mergeTreeClustering/PathMappingDistance.h
floWetzels: I did not use a vector because it is significantly slower due to initialization. This is an n^4 sized memoization array which is filled in a sparse manner during the bottom-up computation. Therefore, I opted for an uninitialized array. Since new and delete should not be used, I wrapped them by a smart pointer and thought that would be ok. I can easily change this of course.

core/base/mergeTreeClustering/BranchMappingDistance.h
floWetzels: Sure, I will have a look into that.

core/vtk/ttkMergeTreeDistanceMatrix/ttkMergeTreeDistanceMatrix.cpp
floWetzels: I agree that this is not good style and I can fix this. However, this function is going to be removed anyway, as soon as I push the update to the FTM data structures.

floWetzels:
Thanks for the remarks on my code! I will fix them soon. But one question beforehand: I already implemented the methods using the FTM data structures. Should I do the conversion within this PR here and then handle the remarks or should we first finish this version?

core/base/mergeTreeClustering/PathMappingDistance.h
floWetzels: Will do that, but of course have to wait until sometime after VIS.

floWetzels: Or should I add the arxiv reference?

core/base/mergeTreeClustering/PathMappingDistance.h
julien-tierny: yeah I'm afraid there might be some memory leaks there.
I'd recommend to switch to the `std::vector`

core/base/mergeTreeClustering/PathMappingDistance.h
julien-tierny: The TopoInVis paper has officially been accepted, right?
That's what the program says anyways :)  
https://topoinvis-workshop.github.io/2022/program.html

I think you can add the TopoInVis refs there.

core/vtk/ttkMergeTreeDistanceMatrix/ttkMergeTreeDistanceMatrix.cpp
julien-tierny: if you don't mind, I'd prefer that you fix that function right now (to make sure it does not stay too long in the source tree with a copy return). thanks!

julien-tierny:
> Thanks for the remarks on my code! I will fix them soon. But one question beforehand: I already implemented the methods using the FTM data structures. Should I do the conversion within this PR here and then handle the remarks or should we first finish this version?

If the FTM data structure is already implemented, then yes, I'd recommend to merge that into this PR.
thanks a lot!

floWetzels:
I think all your remarks are either fixed now or the corresponding code segment has been removed anyway with the switch to the ftm data structures. Could you have a look, @julien-tierny?

julien-tierny:
sorry for the delay @floWetzels 
I'll take care of this before the end of the week.

core/vtk/ttkMergeTreeDistanceMatrix/ttkMergeTreeDistanceMatrix.cpp
```diff
@@ -100,41 +100,85 @@ int ttkMergeTreeDistanceMatrix::run(
...
+    } else {
+      baseModule = 0;
     }
   }
-  if(not branchDecomposition_) {
-    if(normalizedWasserstein_)
-      printMsg("NormalizedWasserstein is set to false since branch "
-               "decomposition is not asked.");
-    normalizedWasserstein_ = false;
+  if(baseModule == 0) {
```
julien-tierny: Would it be possible to have this logic implemented in the base code instead?
(by logic, I mean the check for compatibility between computation options).

Otherwise, the base code becomes very difficult to use (independently of VTK).

(maybe this is not easy to do because there are so many unrelated base code objets)

core/base/mergeTreeClustering/BranchMappingDistance.h
```diff
@@ -0,0 +1,595 @@
...
+        }
+      }
+
+      size_t nn1 = tree1->getNumberOfNodes();
+      size_t nn2 = tree2->getNumberOfNodes();
+      size_t dim1 = 1;
+      size_t dim2 = (nn1 + 1) * dim1;
+      size_t dim3 = (depth1 + 1) * dim2;
+      size_t dim4 = (nn2 + 1) * dim3;
+      // size_t dim1_ = 1;
```
julien-tierny: can you please remove these comments?

julien-tierny: actually, can you remove all commented code? (not just here)
thanks!

core/base/mergeTreeDistanceMatrix/MergeTreeDistanceMatrix.h
```diff
@@ -29,6 +31,11 @@ namespace ttk {
    */
   class MergeTreeDistanceMatrix : virtual public Debug,
                                   virtual public MergeTreeBase {
+  protected:
+    int baseModule = 0;
```
julien-tierny: can you rename into `baseModule_` ?(the `_` post-fix is used throughout the base code layer to denote member variables)

julien-tierny (changes requested):
hi @floWetzels 
thanks for the great work!
I have a few minor comments left. that'd be great if you could address them!
best,

core/vtk/ttkMergeTreeDistanceMatrix/ttkMergeTreeDistanceMatrix.cpp
floWetzels: This is a good point, I think that would make sense. However, the code you are referring to here is actually Mathieu's. The only integrity check I added is in the MergeTreeClustering module (here I only added console outputs). Maybe I should discuss this with Mathieu first and change this later?
@julien-tierny @MatPont

floWetzels:
@julien-tierny The failed CI check seems to be unrelated to my commits, is there a way to restart the workflow? Or maybe I am missing something.

pierre-guillou:
> @julien-tierny The failed CI check seems to be unrelated to my commits, is there a way to restart the workflow? Or maybe I am missing something.

A recent macOS update has created some incompatibilities in the packages we use as dependencies. I'm working on a fix.

core/vtk/ttkMergeTreeDistanceMatrix/ttkMergeTreeDistanceMatrix.cpp
floWetzels: @julien-tierny I just talked to Mathieu and he agrees that it makes sense to move these checks to the base layer. But we have to move some member variables as well and do some restructuring, so we think it makes more sense to do this in another PR.

julien-tierny:
could you please sync this PR (git pull the public ttk dev branch) to retrieve the CI fix (#846 )?

julien-tierny:
ping @floWetzels :)

floWetzels:
> ping @floWetzels :)

Will do it now. Sorry, we had a public holiday here in Germany.

core/vtk/ttkMergeTreeDistanceMatrix/ttkMergeTreeDistanceMatrix.cpp
julien-tierny: sounds good for another PR then.
that'd be great not to forget about it in the long run though :D

julien-tierny:
Alright, that's all good now.
Thanks a lot for this great contribution! Let's merge that in!

julien-tierny:
As a side note, to make sure that the code for these new distances gets tested by the CI upon each PR (and to guarantee that it doesn't get broken by others over time), you want to prepare some ttk example **at some point**.

- Please see the following instructions: https://github.com/topology-tool-kit/ttk-data/blob/dev/CONTRIBUTING.md#2-adding-a-new-example
- And a representative recent example: https://topology-tool-kit.github.io/examples/compactTriangulation/

You could prepare an example as of now, which would show the differences between the 4 merge tree distances currently available in TTK. Maybe you could have a top row of 4 distance matrices (visualized with their heat maps), and a bottom row of 4 dimension reductions (like in this example https://topology-tool-kit.github.io/examples/mergeTreeClustering/), What do you think?

Another option is to prepare an example later when more code related to the last 2 distances is added. It is up to you. 
But remember that for the time being, your code isn't being tested by the CI.

PR #819, opened 2022-09-01
Use SurfaceGeometrySmoother as QuadrangulationSubdivision backend

pierre-guillou:
This PR modifies the `SurfaceGeometrySmoother` and the `QuadrangulationSubdivision` base modules in order for the latter to use the former for the relaxation/projection iterations, thus removing duplicated code.

To do so, a `Quadrangulation` class has been introduced in the `QuadrangulationSubdivision` module. This class implements parts of the `AbstractTriangulation` API (without inheriting from it thanks to duck-typing in template instantiations). The data structure is heavily inspired from `ExplicitTriangulation` without implementing all its methods (no need for now). The `precondition` methods use the `skeleton` module directly (with a slight modification to `OneSkeleton::preconditionEdges` to adapt it to quad cells).

The subdivision algorithm in `QuadrangulationSubdivision` has been adapted to use this `Quadrangulation` data structure, leading to performance improvements in this part of the algorithm: the `QuadrangulationSubdivision` main routine should now be twice as fast.

The `SurfaceGeometrySmoother` module now supports the `reverseProjection` algorithm from `QuadrangulationSubdivision`: if the regular projection onto the reference surface fails, the reverse projection will try to project onto the reference surface alongside the normal of the quad surface at the quad vertex (mean of quad normals in the vertex's star).

New arrays have been added to the PointData output:
* `ttkVertexScalarField` identifies the nearest vertex in the reference surface for each quad mesh vertex,
* `Density` shows if neighbors are far or close,
* `Deformity` measures the squareness of the quads around each vertex.

Some arrays have been removed: `Projection` and `Triangles Checked` that relates to the last projection, now outsourced to `SurfaceGeometrySmoother`.

Due to some improvements in `SurfaceGeometrySmoother`, some slight changes in the `QuadrangulationSubdivision` output around sharp edges in the reference surface are expected.

Enjoy,
Pierre

### DIFF
--- a/core/base/quadrangulationSubdivision/CMakeLists.txt
+++ b/core/base/quadrangulationSubdivision/CMakeLists.txt
@@ -1,10 +1,13 @@
 ttk_add_base_library(quadrangulationSubdivision
   SOURCES
    QuadrangulationSubdivision.cpp
+   Quadrangulation.cpp
   HEADERS
    QuadrangulationSubdivision.h
+   Quadrangulation.h
   DEPENDS
    dijkstra
    geometry
    triangulation
+   surfaceGeometrySmoother
   )
--- a/core/base/quadrangulationSubdivision/Quadrangulation.cpp
+++ b/core/base/quadrangulationSubdivision/Quadrangulation.cpp
@@ -1,0 +1,150 @@
+#include <Quadrangulation.h>
+
+#include <OneSkeleton.h>
+#include <ZeroSkeleton.h>
+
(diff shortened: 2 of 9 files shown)


julien-tierny:
The code looks good. The CI for the python scripts reports hash differences for the morse smale quadrangulation demo.
That's to be expected too I guess. I'd double-check the visual outcome though.

julien-tierny:
Thanks for filing a PR to ttk-data with the updated hashes too.

PR #820, opened 2022-09-01
GitHub Actions: Workaround Python 3.10 on macOS

pierre-guillou:
Last month, macOS brew Python has been updated from 3.9 to 3.10. Since our precompiled `ttk-paraview` used in the TTK CI depends on Python 3.9 (the Python files are installed under `$prefix/lib/python3.9/site-packages`), the TTK CI fails when trying to use the `vtk` or `paraview.simple` Python modules.

This PR offers a simple hotfix that consists in temporarily rolling back to Python 3.9 on macOS. A better fix IMO would be to recompile the `ttk-paraview` headless packages with the current macOS/brew software stack (this can be done in a later step with a revert of this commit).

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -196,18 +196,17 @@ jobs:
     - uses: actions/checkout@v2
       name: Checkout TTK source code
 
-    - name: Remove hosted Python
-      run: |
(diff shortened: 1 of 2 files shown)


julien-tierny:
thanks for the hotfix!

PR #821, opened 2022-09-02
Principal Geodesic Analysis of Merge Trees (and Persistence Diagrams)

MatPont:
This PR adds the code corresponding to the following work:

[Principal Geodesic Analysis of Merge Trees (and Persistence Diagrams)](https://arxiv.org/abs/2207.10960)
Mathieu Pont, Jules Vidal, Julien Tierny

It adds 2 filters to TTK : `MergeTreePrincipalGeodesics` and `MergeTreePrincipalGeodesicsDecoding`.

The first one, `MergeTreePrincipalGeodesics`, takes an ensemble of merge trees or persistence diagrams as input and compute an orthogonal geodesic basis that maximize the variance of the data projected onto it and hence minimize the reconstruction error.
It outputs the barycenter of the ensemble (origin of the basis), the vectors defining the basis, the coefficients of the input (i.e. the coordinates of each data in this new basis) and a correlation matrix allowing to inspect the features that are the most responsible for the variability in the ensemble.

The second filter, `MergeTreePrincipalGeodesicsDecoding`, is used to reconstruct merge trees (or persistence diagrams) on this new basis, it can reconstruct the input and can construct other trees/diagrams like sampling uniformly the surface of the first two axes.
It takes as input the outputs of the previous filter (the correlation matrix is optionnal) and output an ensemble of merge trees (or persistence diagrams).

This PR also adds some features to the `Geometry` module, like generalizing some functions to more than 3 dimensions and add some new functions like the projection of a vector into another.

It also adds some features to `ttkFTMTreeUtils.h` to convert a `vtkUnstructuredGrid` representing a diagram to a `FTMTree_MT` and adds features to the `MergeTreeVisualization` and `ttkMergeTreeVisualization` classes to output persistence diagrams (with the necessary information on points and cells like merge trees have to allow interpretability of the results).


julien-tierny:
can you please git pull the main repo to get the MacOS CI hotfix on your PRs?

core/base/geometry/Geometry.h
```diff
@@ -178,12 +178,14 @@ namespace ttk {
...
-    /// \param vB xyz coordinates of vB std::vector
+    /// Compute the dot product of two std::vectors
+    /// \param vA coordinates of vA std::vector
+    /// \param vB coordinates of vB std::vector
+    /// \param dimension Optional parameter that specifies the dimension of
+    /// the point set (by default 3).
     /// \return Returns Output dot product
     template <typename T>
-    T dotProduct(const T *vA, const T *vB);
+    T dotProduct(const T *vA, const T *vB, const int &dimension = 3);
```
julien-tierny, line 202: nice generalizations, thanks!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,2270 @@
...
+   * the barycenter as origin.
+   */
+  class MergeTreePrincipalGeodesics : virtual public Debug,
+                                      public MergeTreePrincipalGeodesicsBase {
+
+  protected:
+    bool deterministic_ = true;
+    unsigned int numberOfGeodesics_ = 1;
+    unsigned int noProjectionStep_ = 2;
+    // bool rebootDegeneratedGeod_ = true;
```
julien-tierny: could you remove unnecessary commented code?

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,2270 @@
...
+    bool doComputeReconstructionError_ = false;
+    double barycenterSizeLimitPercent_ = 0.0;
+    // TODO keepState works only when enabled before first computation
+    bool keepState_ = false;
+
+    // Gradient descent [Seguy and Cuturi, 2015]
+    // gradient descent does not work anymore
```
julien-tierny: if it is no longer supported, I'd suggest to remove all the associated code (that will ease the long term maintenance of the code)

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,2270 @@
...
+    bool useGradientDescent_ = false;
+    double beta_ = 1e-3; // gradient step
+    double lambda_ = 1; // regularizer
+
+    // Proximal Gradient Descent [Cazelles et al., 2018]
+    // not implemented
```
julien-tierny: same as above.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,2270 @@
...
+    bool useProximalGradientDescent_ = false;
+    double t0_ = 0.0;
+
+    // Advanced parameters
+    bool projectInitializedVectors_ = true;
+    int parallelMode_ = 0; // 0 : Double _ 1 : Simple _ 3 : Double triangle
+                           // inequality _ 2 : Simple triangle inequality
+
+    // Old/Testing
+    bool moveVectorsAlongGeodesics_ = true;
```
julien-tierny: if this is no longer working, I'd recommend to remove that code.
(only keep what's actually running in the examples)

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,2270 @@
...
+    double weightPropCost_ = 1.0;
+    double weightOrthoCost_ = 1.0;
+    double weightMapCost_ = 1.0;
+    double t_vectorCopy_time_ = 0.0, t_allVectorCopy_time_ = 0.0;
+
+    // Filled by the algorithm
+    std::vector<double> baryDistances_;
+    std::vector<std::vector<double>> allDistances_;
+    ftm::MergeTree<double> barycenter_, barycenter2_, barycenterBD_,
```
julien-tierny: could you make these names more explicit?
are these all needed?

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,2270 @@
...
+          v2 = v1;
+          if(trees2.size() != 0) {
+            initRandomVector(barycenter2, trees2V1, trees2Vs_, trees2V2s_);
+            trees2V2 = trees2V1;
+          }
+        }
+
+        // Verify that the two highest persistence pairs are not destroyed
+        /*verifyExtremitiesValidity<dataType>(barycenter, trees, v1, v2);
+        if(trees2.size() != 0)
```
julien-tierny: thanks for removing unnecessary commented code.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+                                                   std::vector<double> &v2,
+                                                   std::vector<double> &v1Out,
+                                                   std::vector<double> &v2Out) {
+    std::vector<double> sumV;
+    sumVector(v1, v2, sumV);
+    vectorProjection(sumV, v1, v1Out);
+    vectorProjection(sumV, v2, v2Out);
+  }
+
+  void MergeTreePrincipalGeodesicsBase::gramSchmidt(
```
julien-tierny: I believe that function should be moved to `core/base/geometry/` as well.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    std::vector<std::vector<double>> &sumV) {
+    std::vector<std::vector<double>> v1_flatten, v2_flatten;
+    multiFlatten(v1, v1_flatten);
+    multiFlatten(v2, v2_flatten);
+    multiSumVector(v1_flatten, v2_flatten, sumV);
+  }
+
+  void MergeTreePrincipalGeodesicsBase::flatten(
+    std::vector<std::vector<double>> &v, std::vector<double> &newV) {
+    newV = std::vector<double>(v.size() * v[0].size());
```
julien-tierny: why don't you `newV.resize(v.size() * v[0].size())`?
I believe that would skip an unnecessary copy of zeroes...

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    std::vector<std::vector<std::vector<double>>> &v1,
+    std::vector<std::vector<std::vector<double>>> &v2,
+    std::vector<std::vector<double>> &sumV) {
+    std::vector<std::vector<double>> v1_flatten, v2_flatten;
+    multiFlatten(v1, v1_flatten);
+    multiFlatten(v2, v2_flatten);
+    multiSumVector(v1_flatten, v2_flatten, sumV);
+  }
+
+  void MergeTreePrincipalGeodesicsBase::flatten(
```
julien-tierny: I believe that function would better be in `core/base/geometry` (maybe with a more explicit name, such as `flattenMultiDimensionalVector`).
for anything you move to geometry, thanks for providing the corresponding doxygen documentation.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    newV = std::vector<double>(v.size() * v[0].size());
+    for(unsigned int i = 0; i < v.size(); ++i)
+      for(unsigned int j = 0; j < v[0].size(); ++j)
+        newV[i * v[0].size() + j] = v[i][j];
+  }
+
+  void MergeTreePrincipalGeodesicsBase::multiFlatten(
+    std::vector<std::vector<std::vector<double>>> &v,
+    std::vector<std::vector<double>> &newV) {
+    newV = std::vector<std::vector<double>>(v.size());
```
julien-tierny: same remark as above (`resize` should be more efficient in principle).

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    for(unsigned int i = 0; i < v.size(); ++i)
+      flatten(v[i], newV[i]);
+  }
+
+  void MergeTreePrincipalGeodesicsBase::unflatten(
+    std::vector<double> &v, std::vector<std::vector<double>> &newV) {
+    newV = std::vector<std::vector<double>>(v.size() / 2);
```
julien-tierny: same remark as above about `resize`.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+  void MergeTreePrincipalGeodesicsBase::multiFlatten(
+    std::vector<std::vector<std::vector<double>>> &v,
+    std::vector<std::vector<double>> &newV) {
+    newV = std::vector<std::vector<double>>(v.size());
+    for(unsigned int i = 0; i < v.size(); ++i)
+      flatten(v[i], newV[i]);
+  }
+
+  void MergeTreePrincipalGeodesicsBase::unflatten(
```
julien-tierny: I believe that function can go to `core/base/geometry` (as discussed above, with the doxygen documentation)

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+                                                  std::vector<double> &v2,
+                                                  std::vector<double> &subV) {
+    subV = std::vector<double>(v1.size());
+    ttk::Geometry::subtractVectors(
+      v2.data(), v1.data(), subV.data(), v1.size());
+  }
+
+  void MergeTreePrincipalGeodesicsBase::multVectorByScalar(
+    std::vector<double> &v, double scalar, std::vector<double> &multV) {
+    multV = v;
```
julien-tierny: `resize` may be more efficient (in particular if multiV has the right size already).

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    double scalar,
+    std::vector<std::vector<double>> &multV) {
+    std::vector<double> v_flat, multV_flat;
+    flatten(v, v_flat);
+    multVectorByScalar(v_flat, scalar, multV_flat);
+    unflatten(multV_flat, multV);
+  }
+
+  double
+    MergeTreePrincipalGeodesicsBase::sumVectorElements(std::vector<double> &v) {
```
julien-tierny: could you move that to `geometry` (with doxygen doc)

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+  void MergeTreePrincipalGeodesicsBase::unflatten(
+    std::vector<double> &v, std::vector<std::vector<double>> &newV) {
+    newV = std::vector<std::vector<double>>(v.size() / 2);
+    for(unsigned int i = 0; i < v.size(); i += 2)
+      newV[i / 2] = std::vector<double>{v[i], v[i + 1]};
+  }
+
+  bool
+    MergeTreePrincipalGeodesicsBase::isVectorUniform(std::vector<double> &v) {
```
julien-tierny: please move to `core/base/geometry` (with doxygen doc)

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    for(unsigned int i = 0; i < v.size() - 1; ++i)
+      if(not(std::abs(v[i] - v[i + 1]) < 1e-6))
```
julien-tierny: I don't believe `1e-6` is sufficient when using double precision (`std::vector<double> &v`).
please double check wikipedia (https://en.wikipedia.org/wiki/Double-precision_floating-point_format) but I believe `1e-15` looks more appropriate.

julien-tierny: maybe the level of precision `1e-15` could be encoded as a constant in `core/base/geometry`.
@pierre-guillou, didn't we have such constants at some point?

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+        return false;
+    return true;
+  }
+
+  bool MergeTreePrincipalGeodesicsBase::isVectorNull(std::vector<double> &v) {
+    for(unsigned int i = 0; i < v.size(); ++i)
+      if(not(std::abs(v[i]) < 1e-6))
```
julien-tierny: same remark as above.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+  bool
+    MergeTreePrincipalGeodesicsBase::isVectorUniform(std::vector<double> &v) {
+    for(unsigned int i = 0; i < v.size() - 1; ++i)
+      if(not(std::abs(v[i] - v[i + 1]) < 1e-6))
+        return false;
+    return true;
+  }
+
+  bool MergeTreePrincipalGeodesicsBase::isVectorNull(std::vector<double> &v) {
```
julien-tierny: -> `core/base/geometry`

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
julien-tierny: found them, in `core/base/geometry/Geometry.cpp`

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    for(unsigned int i = 0; i < v.size(); ++i)
+      if(not(std::abs(v[i]) < 1e-6))
+        return false;
+    return true;
+  }
+
+  bool MergeTreePrincipalGeodesicsBase::isVectorNullFlatten(
```
julien-tierny: -> geometry

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+  void
+    MergeTreePrincipalGeodesicsBase::vectorToPointer(std::vector<double> &vec,
+                                                     double *&pVec) {
+    pVec = vec.data();
+  }
+
+  void MergeTreePrincipalGeodesicsBase::vectorsToPointers(
+    std::vector<std::vector<double>> &vec, std::vector<double *> &pVec) {
+    pVec = std::vector<double *>(vec.size());
```
julien-tierny: `resize()`

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    for(unsigned int i = 0; i < vec.size(); ++i)
+      vectorToPointer(vec[i], pVec[i]);
+  }
+
+  void MergeTreePrincipalGeodesicsBase::vectorOfVectorsToPointers(
+    std::vector<std::vector<std::vector<double>>> &vS,
+    std::vector<std::vector<double *>> &pVS) {
+    pVS = std::vector<std::vector<double *>>(vS.size());
```
julien-tierny: `resize()`

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    for(unsigned int i = 0; i < vS.size(); ++i)
+      vectorsToPointers(vS[i], pVS[i]);
+  }
+
+  void MergeTreePrincipalGeodesicsBase::pointerToVector(
+    double *pVec, size_t size, std::vector<double> &vec) {
+    vec = std::vector<double>(size);
```
julien-tierny: `resize()`

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+  }
+
+  void MergeTreePrincipalGeodesicsBase::vectorsToPointers(
+    std::vector<std::vector<double>> &vec, std::vector<double *> &pVec) {
+    pVec = std::vector<double *>(vec.size());
+    for(unsigned int i = 0; i < vec.size(); ++i)
+      vectorToPointer(vec[i], pVec[i]);
+  }
+
+  void MergeTreePrincipalGeodesicsBase::vectorOfVectorsToPointers(
```
julien-tierny, line 34: I am not sure if all these `vector(...)ToPointers()` are really needed.
in principle you can access the vector like a standard array so I'm a bit puzzled about these functions.
I'm asking because there are lots of copies involved. things would probably go much faster without these copies.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+  double
+    MergeTreePrincipalGeodesicsBase::sumVectorElements(std::vector<double> &v) {
+    double sum = 0;
+    for(auto e : v)
+      sum += e;
+    return sum;
+  }
+
+  double MergeTreePrincipalGeodesicsBase::sumVectorElementsFlatten(
```
julien-tierny: -> geometry?

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    std::vector<double> v_flatten;
+    flatten(v, v_flatten);
+    return sumVectorElements(v_flatten);
+  }
+
+  void MergeTreePrincipalGeodesicsBase::multiSumVector(
+    std::vector<std::vector<double>> &v1,
+    std::vector<std::vector<double>> &v2,
+    std::vector<std::vector<double>> &sumV) {
+    sumV = std::vector<std::vector<double>>(v1.size());
```
julien-tierny: `resize()`?

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    vec = std::vector<double>(size);
+    for(unsigned int i = 0; i < size; ++i)
+      vec[i] = pVec[i];
+  }
+
+  void MergeTreePrincipalGeodesicsBase::pointersToVectors(
+    std::vector<double *> &pVec,
+    std::vector<size_t> sizes,
+    std::vector<std::vector<double>> &vec) {
+    vec = std::vector<std::vector<double>>(pVec.size());
```
julien-tierny: `resize()`?

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    size_t size,
+    std::vector<std::vector<double>> &vec) {
+    std::vector<size_t> sizes(pVec.size(), size);
+    pointersToVectors(pVec, sizes, vec);
+  }
+
+  // ----------------------------------------------------------------------------
+  // Matrix Utils
+  // ----------------------------------------------------------------------------
+  void MergeTreePrincipalGeodesicsBase::matrixDot(
```
julien-tierny: -> `geometry()` definitely

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    std::vector<std::vector<double>> &newM) {
+    newM = std::vector<std::vector<double>>(
+      m1.size(), std::vector<double>(m2[0].size(), 0.0));
+    for(unsigned int i = 0; i < newM.size(); ++i)
+      for(unsigned int j = 0; j < newM[i].size(); ++j)
+        for(unsigned int k = 0; k < m1[i].size(); ++k)
+          newM[i][j] += m1[i][k] * m2[k][j];
+  }
+
+  void MergeTreePrincipalGeodesicsBase::subMatrix(
```
julien-tierny: -> `geometry()` definitely

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+  }
+
+  // ----------------------------------------------------------------------------
+  // Matrix Utils
+  // ----------------------------------------------------------------------------
+  void MergeTreePrincipalGeodesicsBase::matrixDot(
+    std::vector<std::vector<double>> &m1,
+    std::vector<std::vector<double>> &m2,
+    std::vector<std::vector<double>> &newM) {
+    newM = std::vector<std::vector<double>>(
```
julien-tierny: resize

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+      for(unsigned int j = 0; j < newM[i].size(); ++j)
+        for(unsigned int k = 0; k < m1[i].size(); ++k)
+          newM[i][j] += m1[i][k] * m2[k][j];
+  }
+
+  void MergeTreePrincipalGeodesicsBase::subMatrix(
+    std::vector<std::vector<double>> &m1,
+    std::vector<std::vector<double>> &m2,
+    std::vector<std::vector<double>> &newM) {
+    newM = std::vector<std::vector<double>>(
```
julien-tierny: resize()

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+      m1.size(), std::vector<double>(m1[0].size()));
+    for(unsigned int i = 0; i < m1.size(); ++i)
+      for(unsigned int j = 0; j < m1[0].size(); ++j)
+        newM[i][j] = m1[i][j] - m2[i][j];
+  }
+
+  void MergeTreePrincipalGeodesicsBase::sumMatrix(
```
julien-tierny: -> `geometry()` definitely

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    std::vector<std::vector<double>> &m1,
+    std::vector<std::vector<double>> &m2,
+    std::vector<std::vector<double>> &newM) {
+    newM = std::vector<std::vector<double>>(
```
julien-tierny: resize()

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+      m1.size(), std::vector<double>(m1[0].size()));
+    for(unsigned int i = 0; i < m1.size(); ++i)
+      for(unsigned int j = 0; j < m1[0].size(); ++j)
+        newM[i][j] = m1[i][j] + m2[i][j];
+  }
+
+  void MergeTreePrincipalGeodesicsBase::multMatrix(
```
julien-tierny: -> `geometry()` definitely

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    std::vector<std::vector<double>> &m1,
+    double mult,
+    std::vector<std::vector<double>> &newM) {
+    newM = m1;
+    for(unsigned int i = 0; i < newM.size(); ++i)
+      for(unsigned int j = 0; j < newM[i].size(); ++j)
+        newM[i][j] *= mult;
+  }
+
+  void MergeTreePrincipalGeodesicsBase::transpose(
```
julien-tierny: -> `geometry()` definitely

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    std::vector<std::vector<double>> &m,
+    std::vector<std::vector<double>> &newM) {
+    std::vector<std::vector<double>> mTemp(
+      m[0].size(), std::vector<double>(m.size()));
+    for(unsigned int i = 0; i < m.size(); ++i)
+      for(unsigned int j = 0; j < m[0].size(); ++j)
+        mTemp[j][i] = m[i][j];
+    newM = mTemp;
```
julien-tierny: why not just work with `newM` and avoid the copy altogether?

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+  }
+
+  // ----------------------------------------------------------------------------
+  // Statistics Utils
+  // ----------------------------------------------------------------------------
+  double MergeTreePrincipalGeodesicsBase::mean(std::vector<double> &v) {
```
julien-tierny: -> `geometry()` definitely

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    double mean = 0.0;
+    for(auto e : v)
+      mean += e;
+    return mean / v.size();
+  }
+
+  double MergeTreePrincipalGeodesicsBase::var(std::vector<double> &v) {
```
julien-tierny: -> `geometry()` definitely

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    return cov(v, v);
+  }
+
+  double MergeTreePrincipalGeodesicsBase::cov(std::vector<double> &v1,
```
julien-tierny: -> `geometry()` definitely

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+                                              std::vector<double> &v2) {
+    double cov = 0.0;
+    double meanV1 = mean(v1);
+    double meanV2 = mean(v2);
+    for(unsigned int i = 0; i < v1.size(); ++i)
+      cov += (v1[i] - meanV1) * (v2[i] - meanV2);
+    return cov / (v1.size() - 1);
+  }
+
+  double MergeTreePrincipalGeodesicsBase::corr(std::vector<double> &v1,
```
julien-tierny: -> `geometry()` definitely

julien-tierny: maybe `core/base/statistics/`?

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    std::vector<double> &v1,
+    std::vector<double> &v2,
+    std::vector<double> &projec) {
+    projec = std::vector<double>(v1.size(), 0.0);
+    ttk::Geometry::vectorProjection(
+      v2.data(), v1.data(), projec.data(), v1.size());
+  }
+
+  void
+    MergeTreePrincipalGeodesicsBase::sumProjection(std::vector<double> &v1,
```
julien-tierny: --> geometry?

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,2270 @@
...
+          }
+          for(unsigned int j = 0; j < trees2V.size(); ++j) {
+            trees2V[j] = vConcat[v.size() + j];
+            trees2V2[j] = v2Concat[v.size() + j];
+          }
+          t_vectorCopy_time_ += t_vectorCopy.getElapsedTime();
+        }
+
+        // --- Projection convergence
+        /*if(projectionStepConvergence(geodesicNumber, v, v2, vS, v2s, vOld,
```
julien-tierny: thanks for removing unnecessary commented code.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,2270 @@
...
+          break;*/
+      }
+      return optMapCost;
+    }
+
+    //----------------------------------------------------------------------------
+    // Assignment
+    //----------------------------------------------------------------------------
+    template <class dataType>
+    void assignmentDoublePara(
```
julien-tierny: could you use a more explicit name for that functioN?

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    std::vector<double> v_flatten;
+    flatten(v, v_flatten);
+    return norm(v_flatten);
+  }
+
+  void MergeTreePrincipalGeodesicsBase::vectorProjection(
+    std::vector<double> &v1,
+    std::vector<double> &v2,
+    std::vector<double> &projec) {
+    projec = std::vector<double>(v1.size(), 0.0);
```
julien-tierny: resize?

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+  }
+
+  double MergeTreePrincipalGeodesicsBase::normFlatten(
+    std::vector<std::vector<double>> &v) {
+    std::vector<double> v_flatten;
+    flatten(v, v_flatten);
+    return norm(v_flatten);
+  }
+
+  void MergeTreePrincipalGeodesicsBase::vectorProjection(
```
julien-tierny: -> `geometry()` definitely

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    flatten(v1, v1_flatten);
+    flatten(v2, v2_flatten);
+    return scalarProduct(v1_flatten, v2_flatten);
+  }
+
+  double MergeTreePrincipalGeodesicsBase::norm(std::vector<double> &v) {
+    return ttk::Geometry::magnitude(v.data(), v.size());
+  }
+
+  double MergeTreePrincipalGeodesicsBase::normFlatten(
```
julien-tierny: -> `geometry()` ?

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+    return distanceL2(v1_flatten, v2_flatten);
+  }
+
+  double
+    MergeTreePrincipalGeodesicsBase::scalarProduct(std::vector<double> &v1,
+                                                   std::vector<double> &v2) {
+    return ttk::Geometry::dotProduct(v1.data(), v2.data(), v1.size());
+  }
+
+  double MergeTreePrincipalGeodesicsBase::scalarProductFlatten(
```
julien-tierny: -> `geometry()` ? (name it `dotProduct` to be consistent)

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
```diff
@@ -0,0 +1,334 @@
...
+namespace ttk {
+  // ----------------------------------------------------------------------------
+  // Vector Utils
+  // ----------------------------------------------------------------------------
+  double MergeTreePrincipalGeodesicsBase::distanceL2(std::vector<double> &v1,
+                                                     std::vector<double> &v2) {
+    return ttk::Geometry::distance<double>(v1.data(), v2.data(), v1.size());
+  }
+
+  double MergeTreePrincipalGeodesicsBase::distanceL2Flatten(
```
julien-tierny: -> `geometry()`?

julien-tierny: warning, for all these move to geometry, thanks for templatizing your implementation (not restricted to double)

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,2270 @@
...
+#ifdef TTK_ENABLE_OPENMP
+        } // pragma omp single nowait
+      } // pragma omp parallel
+#endif
+
+      // std::cout << "t_para : " << t_para.getElapsedTime() << std::endl;
+    }
+
+    template <class dataType>
+    void assignmentSimplePara(
```
julien-tierny: it seems like you have two parallel implementations of the same thing here.
is that the case?
are the two of them actually run by the code?
(if not, thanks for removing the unused one).

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,2270 @@
...
+        ssCumulT << "Cumulative explained T-Variance : "
+                 << round(cumulTVariancePercent * 100.0) / 100.0 << " %";
+        printMsg(ssCumulT.str());
+      }
+    }
+
+    //----------------------------------------------------------------------------
+    // Utils
+    //----------------------------------------------------------------------------
+    template <class dataType>
```
julien-tierny: wouldn't that util function be more relevant in `core/base/mergeTreeClustering/`?

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,2270 @@
...
+        }
+      }
+
+      if(cptBelowDiagonal != 0)
+        printErr(std::to_string(cptBelowDiagonal)
+                 + " points in interpolation below diagonal.");
+    }
+
+    template <class dataType>
+    void makeMyTest(std::vector<ftm::MergeTree<dataType>> &trees,
```
julien-tierny: does this need to be on the git ?

julien-tierny (changes requested):
here's a first batch of reviews, thanks!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Done in the next commit!

core/base/geometry/Geometry.h
MatPont: Thanks!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Done in the next commit! (removed the unused one)

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Actually this function is not used anymore and will be removed in the next commit.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Done in the next commit! (removed)

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: It was in case we want to use the same matrix as input and output (to do the transpose inplace). We can keep this like that to allow the modification iplace, or remove this feature (meaning that the input and output can not be the same object).

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Indeed the `pointer(...)ToVectors()` functions make a copy, but they are used with this purpose of making a copy (to not modify the original).
The motivation is the following, in the decoding we use pointers to the arrays of the vtkTable representing the geodesic vectors (to avoid to copy them permanently). But when we compute an interpolation (for example creating the trees of the surface) the vectors are potentially temporarily modified to ensure nesting condition and birth < death. 
We need that the original vectors are not modified (moreover since there is some parallelism involved we also need that each thread modifying its vectors do not modify the vectors of another thread).

MatPont: Indeed the `pointer(...)ToVectors()` functions make a copy, but they are only used with this purpose of making a copy (to not modify the original).
The motivation is the following, in the decoding we use pointers to the arrays of the vtkTable representing the geodesic vectors (to avoid to copy them permanently). But when we compute an interpolation (for example creating the trees of the surface) the vectors are potentially temporarily modified to ensure nesting condition and birth < death. 
We need that the original vectors are not modified (moreover since there is some parallelism involved we also need that each thread modifying its vectors do not modify the vectors of another thread hence the need of a copy).
But perhaps I'm missing an easiest solution.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit! (new module `statistics` created)

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit! (in statistics)

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit! (in statistics)

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit! (in statistics)

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Removed in the next commit.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Removed in the next commit.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Renamed in the next commit! (they are indeed still needed)

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.cpp
MatPont: The new definition now avoid copy. It is left to the used of the function.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Done in the next commit!

core/base/geometry/Geometry.cpp
```diff
@@ -481,59 +517,353 @@ T Geometry::magnitude(const T *o, const T *d) {
...
+  return scaleVector(a.data(), factor, out.data(), a.size());
+}
+
+template <typename T>
+int Geometry::vectorProjection(const T *a,
+                               const T *b,
+                               T *out,
+                               const int &dimension) {
+  T dotProdBB = dotProduct(b, b, dimension);
+  if(dotProdBB > 1e-12) {
```
julien-tierny: instead of `1e-12`, you want to use one of the constants used in TTK (e.g. `1e-DBL_DIG` or `PREC_DBL`)

julien-tierny: you may want to double-check for other occurences of `1e-12`.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.cpp
```diff
@@ -0,0 +1,45 @@
...
+      return 0.0;
+    if(doPrint)
+      printMsg("Scalar Products:");
+    double cost = 0.0;
+    std::vector<std::vector<double>> sumVs;
+    ttk::Geometry::multiAddVectorsFlatten(vS, v2s, sumVs);
+    for(unsigned int i = 0; i < geodesicNumber; ++i) {
+      double scalarProd
+        = ttk::Geometry::dotProduct(sumVs[i], sumVs[geodesicNumber]);
+      cost += std::pow(scalarProd, 2);
```
julien-tierny: please avoid using `std::pow` when the exponent is an integer.
instead, thanks for using @pierre-guillou faster lambdas (`TTK_POW_LAMBDA`). this comes with performance improvements. thanks for double-checking your code for other occurences.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,1652 @@
...
+    double orthogonalCost(std::vector<std::vector<std::vector<double>>> &vS,
+                          std::vector<std::vector<std::vector<double>>> &v2s,
+                          std::vector<std::vector<double>> &v,
+                          std::vector<std::vector<double>> &v2) {
+      return verifyOrthogonality(vS, v2s, v, v2, false);
+    }
+
+    double regularizerCost(std::vector<std::vector<double>> &v,
+                           std::vector<std::vector<double>> &v2) {
+      return std::pow((ttk::Geometry::dotProductFlatten(v, v2)
```
julien-tierny: same remark as before with `std::pow`

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,1652 @@
...
+        for(unsigned int j = 0; j < vS.size(); ++j) {
+          vSConcat[j].insert(
+            vSConcat[j].end(), trees2Vs[j].begin(), trees2Vs[j].end());
+          v2sConcat[j].insert(
+            v2sConcat[j].end(), trees2V2s[j].begin(), trees2V2s[j].end());
+        }
+        t_vectorCopy_time_ += t_vectorCopy.getElapsedTime();
+      }
+
+      // double oldCost = -1, oldCost2 = -1;
```
julien-tierny: thanks for removing commented code.
thanks for double-checking that in other places of your code.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,1652 @@
...
+            birthDeath
+              = trees[i].tree.template getBirthDeath<dataType>(matched);
+          births[i] = std::get<0>(birthDeath);
+          deaths[i] = std::get<1>(birthDeath);
+          pers[i] = deaths[i] - births[i];
+        }
+
+        // Compute correlations
+        for(unsigned int g = 0; g < numberOfGeodesics_; ++g) {
+          double birthCorr = Statistics::corr(births, allTs_[g]);
```
julien-tierny: could you please change to `ttk::Statistics::corr` to make it more explicit.
could you please double-check that in the rest of the code?

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.h
```diff
@@ -0,0 +1,893 @@
...
+    void computeOneDistance(
+      ftm::MergeTree<dataType> &tree1,
+      ftm::MergeTree<dataType> &tree2,
+      std::vector<std::tuple<ftm::idNode, ftm::idNode, double>> &matching,
+      dataType &distance,
+      bool isCalled = false,
+      bool useDoubleInput = false,
+      bool isFirstInput = true) {
+      MergeTreeDistance mergeTreeDistance;
+      mergeTreeDistance.setDebugLevel(2);
```
julien-tierny: what if someone wants to run ttk with no debug information at all?
I suggest to use a predefined value (in this example 2), unless `debugLevel_` is already set at a lower value (for instance 1 or 0). Could you address this remark throughout your code?

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.h
```diff
@@ -0,0 +1,893 @@
...
+    //----------------------------------------------------------------------------
+    double getGeodesicVectorMiddle(std::vector<double> &v,
+                                   std::vector<double> &v2) {
+      std::vector<double> vProj, v2Proj;
+      ttk::Geometry::addVectorsProjection(v, v2, vProj, v2Proj);
+
+      double alpha = 0.0;
+      int cptDivide = 0;
+      for(unsigned int i = 0; i < vProj.size(); ++i) {
+        if(std::abs(v2Proj[i]) < 1e-12)
```
julien-tierny: same remark as before regarding `1e-12`

core/base/mergeTreePrincipalGeodesicsDecoding/MergeTreePrincipalGeodesicsDecoding.h
```diff
@@ -0,0 +1,484 @@
...
+      std::array<double, 2> middle;
+      getGeodesicsMiddle<dataType>(barycenter, vS, v2s, vSizeToUse, middle);
+
+      // Get ellipses
+#ifdef TTK_ENABLE_OPENMP
+#pragma omp parallel for schedule(dynamic) num_threads(this->threadNumber_)
+#endif
+      for(unsigned int i = 0; i < noSample; ++i) {
+        double angle = 360.0 / noSample * i;
+        double pi = 3.14159265359;
```
julien-tierny: please use the system constant `M_PI` instead.

julien-tierny (changes requested):
Thanks a lot for this great work @MatPont 
Could you please address my final remarks?
Thanks!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: I did another check for removing old code in the next commit, thanks!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.h
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesicsDecoding/MergeTreePrincipalGeodesicsDecoding.h
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.cpp
MatPont: Modified in the next commit!
As suggested by @pierre-guillou I simply used a multiplication, since the power is fixed, the call to `TTK_POW_LAMBDA` could be less efficient than just a multiplication. The multiplication is still better than `std::pow`.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Modified in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.h
MatPont: Done in the next commit!

core/base/geometry/Geometry.cpp
MatPont: Done in the next commit!

By the way, in the meantime I changed the function to adopt the convention used in the [Gram-Schmidt algorithm](https://en.wikipedia.org/wiki/Gram%E2%80%93Schmidt_process).
The convention is the following: the projection of a vector on a null vector is a null vector.
Therefore the function will gives a null vector in that case instead of returning failure (useful for the `gramSchmidt` function).

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.h
MatPont: Done in the next commit!

Note: since we can not access `PREC_DBL` outside `Geometry` I use its definition like for `FiberSurface`.

julien-tierny:
could you please sync this PR (git pull the public ttk dev branch) to retrieve the CI fix (https://github.com/topology-tool-kit/ttk/pull/846 )?

julien-tierny:
ping @MatPont :)

MatPont:
done!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,1646 @@
...
+      ftm::MergeTree<dataType> &tree,
+      std::vector<std::tuple<ftm::idNode, ftm::idNode, double>> &matching,
+      std::vector<std::vector<double>> &v) {
+      ftm::FTMTree_MT *barycenterTree = &(barycenter.tree);
+      ftm::FTMTree_MT *treeTree = &(tree.tree);
+
+      std::vector<ftm::idNode> matchingVector;
+      getMatchingVector<dataType>(barycenter, tree, matching, matchingVector);
+
+      v = std::vector<std::vector<double>>(
```
julien-tierny: `v.resize()`
you may want to search and replace all occurrences of `= std::vector`.
my understanding is that it will make unnecessary copies.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,1646 @@
...
+                          std::vector<std::vector<std::vector<double>>> &v2s) {
+      // Get average norm of the previous vectors
+      std::vector<std::vector<double>> sumVs;
+      ttk::Geometry::multiAddVectorsFlatten(vS, v2s, sumVs);
+      double newNorm = 0;
+      for(auto &sumVi : sumVs)
+        newNorm += ttk::Geometry::magnitude(sumVi) / sumVs.size();
+
+      // Generate random vector
+      v = std::vector<std::vector<double>>(
```
julien-tierny: resize()

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,1646 @@
...
+      mergeTreeBary.setPostprocess(false);
+      mergeTreeBary.setBranchDecomposition(true);
+      mergeTreeBary.setNormalizedWasserstein(normalizedWasserstein_);
+      mergeTreeBary.setKeepSubtree(false);
+      mergeTreeBary.setAssignmentSolver(assignmentSolverID_);
+      mergeTreeBary.setThreadNumber(this->threadNumber_);
+      mergeTreeBary.setDeterministic(deterministic_);
+      mergeTreeBary.setBarycenterSizeLimitPercent(barycenterSizeLimitPercent_);
+
+      matchings = std::vector<
```
julien-tierny: resize()
you may want to search and replace all occurrences of = std::vector.
my understanding is that it will make unnecessary copies.

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,1646 @@
...
+      dataType distance;
+      std::vector<ftm::idNode> matchingVector;
+
+      if(extremityTree->getRealNumberOfNodes() != 0) {
+        computeOneDistance(barycenter, extremity, matching, distance, true,
+                           useDoubleInput, isFirstInput);
+        getMatchingVector(barycenter, extremity, matching, matchingVector);
+      } else
+        matchingVector
+          = std::vector<ftm::idNode>(barycenterTree->getNumberOfNodes(), -1);
```
julien-tierny: v.resize()
you may want to search and replace all occurrences of = std::vector.
my understanding is that it will make unnecessary copies.

I'll stop reporting here by it looks like there's 60 occurrences of that in this PR.

core/base/mergeTreePrincipalGeodesicsDecoding/MergeTreePrincipalGeodesicsDecoding.h
```diff
@@ -0,0 +1,484 @@
...
+                            std::array<double, 2> &middle) {
+      int cptDivide = 0;
+      std::vector<double> alpha(2, 0.0);
+      for(unsigned int i = 0; i < 2; ++i) {
+        cptDivide = 0;
+        for(unsigned int j = 0; j < vSize; ++j) {
+          if(barycenter.tree.isNodeAlone(j))
+            continue;
+          for(unsigned int k = 0; k < 2; ++k) {
+            if(std::abs(v2s[i][k][j]) < 1e-12)
```
julien-tierny: there's still an `1e-12` that can be replaced by `PREC_DBL` or so.
thanks for search and replace :)

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
```diff
@@ -0,0 +1,1646 @@
...
+      ssOptMapCost << "Map. cost   = " << optMapCost;
+      printMsg(ssOptMapCost.str());
+
+      // Detect convergence
+      double tol = 0.01;
+      tol = oldFrechetEnergy / 125.0;
+      converged = std::abs(frechetEnergy - oldFrechetEnergy) < tol;
+      oldFrechetEnergy = frechetEnergy;
+
+      if(frechetEnergy + 1e-6 < minFrechetEnergy) {
```
julien-tierny: you may want to define a constant at the top of the file (which could be reused or easily changed) instead of using  a numerical value `1e-6`

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.h
```diff
@@ -0,0 +1,894 @@
...
+        interpolationVector[nodeDeath] = interDeath;
+
+        // - Verify NaN
+        if(std::isnan(interpolationVector[nodeBirth])
+           or std::isnan(interpolationVector[nodeDeath])) {
+          printMsg(ssT.str());
+          printErr("NOT A NUMBER");
+        }
+        // - Verify points below diagonal
+        double eps = 1e-4;
```
julien-tierny: you may want to define a constant at the top of the file (which could be reused or easily changed) instead of using a numerical value `1e-4`

core/vtk/ttkPlanarGraphLayout/ttkMergeTreeVisualization.h
```diff
@@ -941,9 +986,28 @@ class ttkMergeTreeVisualization : public ttk::MergeTreeVisualization {
           for(auto child : children)
             queue.emplace(child);
 
+          /*if(isPersistenceDiagram) {
```
julien-tierny: can you please remove commented code?

core/vtk/ttkPlanarGraphLayout/ttkMergeTreeVisualization.h
```diff
@@ -991,37 +1056,71 @@ class ttkMergeTreeVisualization : public ttk::MergeTreeVisualization {
...
-            /*and !isLeaf(trees[i], node)
-            and isBranchOrigin(trees[i], node)*/
-            ;
-          if(dummyNode)
-            treeDummySimplexId[node] = points->InsertNextPoint(
-              point); // will be modified when processing son
+          bool dummyNode
+            = PlanarLayout and not branchDecompositionPlanarLayout_
+              and (!trees[i]->isRoot(node) or isPersistenceDiagram);
+          /*and !isLeaf(trees[i], node)
```
julien-tierny: please remove commented code.

julien-tierny (changes requested):
ok, thanks @MatPont !
could you take care of the last minor comments I entered? thanks!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Sorry for the missing `resize`.
Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Done in the next commit!

core/vtk/ttkPlanarGraphLayout/ttkMergeTreeVisualization.h
MatPont: Done in the next commit!

core/vtk/ttkPlanarGraphLayout/ttkMergeTreeVisualization.h
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesicsDecoding/MergeTreePrincipalGeodesicsDecoding.h
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesics.h
MatPont: Done in the next commit!

core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.h
MatPont: Done in the next commit!

julien-tierny:
alright \0/
it looks like we're good to go.
thanks again for this awesome work!!!

julien-tierny:
thanks for opening your PR on ttk-data with examples!!!!

julien-tierny:
This PR is triggering many warnings with gcc 11. For instance:
```
/home/julien/Pro/git/github/tierny/private/ttk-tierny/core/base/mergeTreePrincipalGeodesics/MergeTreePrincipalGeodesicsBase.h:754:65: warning: comparison of unsigned expression in '>= 0' is always true [-Wtype-limits]
  754 |            and match1 < tree.tree.getNumberOfNodes() and match1 >= 0)
```
@MatPont could you fix those?

@pierre-guillou aren't these supposed to be caught by the CI?

pierre-guillou:
These warnings are specific to GCC and appear only on Ubuntu 22.04 and newer. The CI prevents warnings that are found with Clang + `libstdc++`. Actually, other warnings are also found in the macOS and Windows builds (I plan to fix them soon).

It is quite hard currently to enforce no warnings in the build tasks since the GCC of Ubuntu 18.04 mixes variables and functions in `-Wshadow`. Once Ubuntu 18.04 jobs are removed, we can build TTK with `-Werror` enabled (at least on Ubuntu).

julien-tierny:
> These warnings are specific to GCC and appear only on Ubuntu 22.04 and newer. The CI prevents warnings that are found with Clang + `libstdc++`. Actually, other warnings are also found in the macOS and Windows builds (I plan to fix them soon).
> 
> It is quite hard currently to enforce no warnings in the build tasks since the GCC of Ubuntu 18.04 mixes variables and functions in `-Wshadow`. Once Ubuntu 18.04 jobs are removed, we can build TTK with `-Werror` enabled (at least on Ubuntu).

I see, thanks for the clarification!

PR #822, opened 2022-09-05
DiscreteMorseSandwich: Fix IgnoreBoundary

pierre-guillou:
This PR modifies the handling of the `IgnoreBoundary` option. Now, the option removes only the saddle-max pair that involves the global maximum of the dataset. This should reduce the distance to FTM diagrams.

Also, a bug involving (non) memory cleanup for 1D datasets was fixed.

Enjoy,
Pierre


julien-tierny:
looks good, thanks!

PR #823, opened 2022-09-07
DiscreteGradient: Add cell ids to gradient glyphs

pierre-guillou:
This PR extends the Gradient Glyphs output of the DiscreteGradient module with 2 new PointData arrays: CellIds and CellDimension (similar arrays are already on the Critical Points output). This is a way to expose the internal structure of the DiscreteGradient to the VTK layer.

Enjoy,
Pierre


julien-tierny:
alright, thanks!

PR #818, opened 2022-09-01
Update PersistenceDiagramClustering, DistanceMatrix with new PersistenceDiagram structure

pierre-guillou:
As a follow-up to #745 (and as a complement to #803), this PR makes the `PersistenceDiagramClustering`, `PersistenceDiagramDistanceMatrix` & `PersistenceDiagramAuction` modules use the new PersistenceDiagram structure defined in core/base/PersistenceDiagram/PersistencDiagramUtils.h instead of a templated `std::tuple` (which is quite harder to understand & manipulate). Within this new data structure, a boolean field identifies infinite pairs that can be produced by the DiscreteMorseSandwich algorithm.

The unified PersistenceDiagram data structure also allows to de-template the code in the concerned base layer modules, leading to better parallelism during the build process and more precise code lint and static code analysis with `clang-tidy`.

In `PersistenceDiagramAuction`, the `GoodDiagram` and `BidderDiagram` data structures have been respectively aliased to `std::vector<Good>` and `std::vector<Bidder>` to further reduce their footprint in terms of LoC.

A lot of commented code in the base layer has also been removed.

Per diagram aggregated matching costs has been added as FieldData to the Clustered Diagrams and Cluster Centroids outputs. The global cost is already present on the FieldData of the Matchings output.

No change is to be expected on the two ttk-data state files that use these modules. However, newly added matchings costs should change the outputs of the corresponding Python scripts.

Enjoy,
Pierre

PR #824, opened 2022-09-07
Fix segfault, other small changes

pierre-guillou:
This PR fixes a buffer overflow introduced in #818 in `PersistenceDiagramClustering` (when aggregating matchings costs per diagram).

Other small fixes include:
* removal of useless casts in `MPIUtils.h`,
* removal of a dynamic allocation in `Debug.h`,
* `ScalarFieldCriticalPoints` now correctly identifies saddles on (non-manifold) 1D datasets.

Enjoy,
Pierre

### DIFF
--- a/core/base/common/Debug.h
+++ b/core/base/common/Debug.h
@@ -25,6 +25,7 @@
 #include <BaseClass.h>
 
 #include <algorithm>
+#include <array>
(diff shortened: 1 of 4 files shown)


julien-tierny:
wonderful, thanks!

PR #825, opened 2022-09-08
MergeTreeClustering and MergeTreeDistanceMatrix field data

MatPont:
This PR fixes the field data output of the `MergeTreeClustering` and `MergeTreeDistanceMatrix` filters.

Before, only the nodes field data were copied to the output. Since `FTMTree` copies the field data in the segmentation output it could result in a loss of information during the pipeline.

Now, the field data of the segmentation output are also copied. However, the segmentation is optional for these two filters mentioned at the beginning, therefore a user that wants to keep the field data should provide the segmentation to these filters (no impact on the performance).


julien-tierny:
looks good, thanks!

PR #826, opened 2022-09-16
Enable per-dimension computation for DiscreteMorseSandwich Persistence Diagrams

pierre-guillou:
This PR enables per-dimension computation of the persistence pairs for the DiscreteMorseSandwich backend of the PersistenceDiagram module.

This behavior is controlled in the GUI by a dropdown menu that let the user choose between computing pairs of all dimensions or if specific dimensions should be requested. If the second option is chosen, three checkbox appear (one to select each dimension/pair type).

Enjoy,
Pierre


julien-tierny:
awesome, thanks!

PR #827, opened 2022-09-16
[MergeTreeDistanceMatrix] general field data output management

MatPont:
Since #825, the `mergeTreeClustering` state and python file of ttk-data cause a segfault.

The problem was due to the way `MergeTreeDistanceMatrix` manages field data.

Since the PR mentionned above, the original field data (like `TimeStep`, `FILE` etc.) are passed to the output of merge tree related filters. But, the centroids computed by `MergeTreeClustering` does not have these field data, therefore, when creating the distance matrix (of the input trees and the centroids trees) it causes a segfault by trying to find these field data in the centroids.

To summary, the problem occurs when a tree has some field data that another tree doesn't have.

In order to still have the field data in the distance matrix, the solution proposed here is to add "empty" value to the arrays that are present in some trees and not in other ones, `nan` for `dataArray` and `""` for `stringArray`.

Actually, since `MergeTreeDistanceMatrix` was inspired by `PersistenceDiagramDistanceMatrix` the problem also occurs with the latter (for example, if we load `Isabel.cdb`, compute a clustering then compute a distance matrix with the input diagrams and the centroids it will causes a segfault). Another PR inspired by this one could fix this.

### DIFF
--- a/core/vtk/ttkMergeTreeDistanceMatrix/ttkMergeTreeDistanceMatrix.cpp
+++ b/core/vtk/ttkMergeTreeDistanceMatrix/ttkMergeTreeDistanceMatrix.cpp
@@ -8,6 +8,7 @@
 
 #include <vtkDoubleArray.h>
 #include <vtkInformation.h>
+#include <vtkStringArray.h>
 #include <vtkTable.h>
 
 using namespace ttk;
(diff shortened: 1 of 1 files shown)


julien-tierny:
thanks!
@pierre-guillou you may want to do something similar for the persistence diagram clustering.

julien-tierny:
looks great, thanks!

PR #828, opened 2022-09-20
PersistenceDiagramDistanceMatrix: Ensure all FieldData arrays are forwarded

pierre-guillou:
This PR extends #827 to the PersistenceDiagramDistanceMatrix filter. Also, the CinemaQuery filters now supports NaN values.

Enjoy,
Pierre


julien-tierny:
thanks!
(I take back my prior comment)

PR #831, opened 2022-09-21
[MPI] Added dimensionNumber to exchangeGhostCells

m-s-will:
This PR extends the exchangeGhostCells function to use the optional dimensionNumber parameter. Similarly to the `ttkGeometrySmoother` filter, this allows it to be used for exchanging e.g. 3D point arrays which are compressed into an 1D array.

### DIFF
--- a/core/base/common/MPIUtils.h
+++ b/core/base/common/MPIUtils.h
@@ -115,7 +115,8 @@ namespace ttk {
                           const std::unordered_set<int> &neighbors,
                           const int rankToSend,
                           const IT nVerts,
-                          MPI_Comm communicator) {
+                          MPI_Comm communicator,
+                          const int dimensionNumber) {
 
     if(!ttk::isRunningWithMPI()) {
       return -1;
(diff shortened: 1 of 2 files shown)


julien-tierny:
@eve-le-guillou does this fix work for you?

eve-le-guillou:
Thank you Michael,
My example unfortunately uses codes that I have not yet merged into TTK, as it uses a distributed version of TTKIntegralLines. It does not seem to work with your fix.
Here is the results I obtain from my Pipeline (on 8 processes): 
[results.zip](https://github.com/topology-tool-kit/ttk/files/9624212/results.zip)
It gives the following output: 
Screenshot from 2022-09-22 11-04-38
And the output, with the Geometry Smoother added with only one iteration:
Screenshot from 2022-09-22 11-12-23
As can be seen some lines diverge completely from what they should be.
I have not yet been able to pin point where exactly the problem is, however it is possible that the problem comes from somewhere else in TTK, as the managing of Global Ids for Edges is fairly new. Please let me know if you feel that it is the case.

m-s-will:
No, sorry that was probably a mistake from me, the dimensionNumber wasn't given to the smooth call as a parameter. You file crashes on my machine while preconditioning the vertex neighbors in the triangulation which sounds like a different problem, but now I get the same results for the wavelet whether smoothing in sequential mode or in MPI, please do try it again. @eve-le-guillou 

Bildschirmfoto 2022-09-22 um 12 03 59

eve-le-guillou:
> You file crashes on my machine while preconditioning the vertex neighbors in the triangulation which sounds like a different problem

Yes, I changed a thing or two in ttkAlgorithm, so this is most likely why, you can disregard this.

> now I get the same results for the wavelet whether smoothing in sequential mode or in MPI, please do try it again

Indeed, that was it, now my example works well, I obtain the correct results.
Thank you!

julien-tierny:
alright, I let the CI finish and I'll merge that in then.

PR #832, opened 2022-09-23
[MPI] Add boolean attribute to track MPI support in modules

eve-le-guillou:
Add a boolean attribute to ttkAlgorithm called `hasMPISupport_`, set to `false` by default. In modules with MPI support, this attribute should be set to `true` in the constructor, as done here for ArrayPreconditioning, GeometrySmoother, ScalarFieldCriticalPoints, GhostCellPreconditioning, ScalarFieldNormalizer and ScalarFieldSmoother.

When `hasMPISupport_` is equal to `false`, TTK will print the following error message: "MPI is not supported for this filter, the results will be incorrect", as shown here for the filter DiscreteGradient executed on 4 processes:
Screenshot from 2022-09-23 11-08-46

### DIFF
--- a/core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
+++ b/core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
@@ -42,6 +42,10 @@ ttk::Triangulation *ttkAlgorithm::GetTriangulation(vtkDataSet *dataSet) {
                  ttk::debug::Priority::DETAIL);
 #if TTK_ENABLE_MPI
   if(ttk::isRunningWithMPI()) {
+    if(!hasMPISupport_) {
+      printErr(
+        "MPI is not supported for this filter, the results will be incorrect");
+    }
     this->MPIPipelinePreconditioning(dataSet);
   }
 #endif
--- a/core/vtk/ttkAlgorithm/ttkAlgorithm.h
+++ b/core/vtk/ttkAlgorithm/ttkAlgorithm.h
@@ -208,9 +208,6 @@ public:
 protected:
   ttkAlgorithm();
   ~ttkAlgorithm() override;
-#ifdef TTK_ENABLE_MPI
-  bool hasMPISupport_{false};
-#endif
 
   /**
    * This method is called in GetTriangulation, after the triangulation as been
--- a/core/vtk/ttkAlgorithm/ttkAlgorithm.h
+++ b/core/vtk/ttkAlgorithm/ttkAlgorithm.h
@@ -208,6 +208,9 @@ public:
 protected:
   ttkAlgorithm();
   ~ttkAlgorithm() override;
+#ifdef TTK_ENABLE_MPI
+  bool hasMPISupport_{false};
+#endif
 
   /**
    * This method is called in GetTriangulation, after the triangulation as been
--- a/core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
+++ b/core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
@@ -15,7 +15,9 @@
 vtkStandardNewMacro(ttkArrayPreconditioning);
 
 ttkArrayPreconditioning::ttkArrayPreconditioning() {
+#ifdef TTK_ENABLE_MPI
   hasMPISupport_ = true;
+#endif
   this->SetNumberOfInputPorts(1);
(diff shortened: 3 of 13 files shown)


core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -42,6 +42,10 @@ ttk::Triangulation *ttkAlgorithm::GetTriangulation(vtkDataSet *dataSet) {
                  ttk::debug::Priority::DETAIL);
 #if TTK_ENABLE_MPI
   if(ttk::isRunningWithMPI()) {
+    if(!hasMPISupport_) {
```
julien-tierny, line 45: ok, that's the rightspot. 
the constructor could have been a (more verbose) alternative.
like this, it is fine :)

core/vtk/ttkAlgorithm/ttkAlgorithm.h
```diff
@@ -208,6 +208,9 @@ class TTKALGORITHM_EXPORT ttkAlgorithm : public vtkAlgorithm,
 protected:
   ttkAlgorithm();
   ~ttkAlgorithm() override;
+#ifdef TTK_ENABLE_MPI
```
julien-tierny: I would declare that boolean much higher in the hierarchy (not to exclude the possibility for the future to maybe use mpi outside of paraview/vtk).
I'd recommend to move that to `core/base/common/BaseClass.h`

core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
```diff
@@ -15,6 +15,9 @@
 vtkStandardNewMacro(ttkArrayPreconditioning);
 
 ttkArrayPreconditioning::ttkArrayPreconditioning() {
+#ifdef TTK_ENABLE_MPI
+  hasMPISupport_ = true;
```
julien-tierny: I'd recommend to move that to the constructor of the *base* class when available.
in this example `core/base/arrayPreconditioning/`

core/vtk/ttkGeometrySmoother/ttkGeometrySmoother.cpp
```diff
@@ -11,7 +11,9 @@ vtkStandardNewMacro(ttkGeometrySmoother);
 
 ttkGeometrySmoother::ttkGeometrySmoother() {
   this->setDebugMsgPrefix("GeometrySmoother");
-
+#ifdef TTK_ENABLE_MPI
+  hasMPISupport_ = true;
```
julien-tierny: I'd recommend to move that to the constructor of the *base* class when available.
in this example `core/base/scalarFieldSmoother/`

core/vtk/ttkScalarFieldCriticalPoints/ttkScalarFieldCriticalPoints.cpp
```diff
@@ -20,7 +20,9 @@ using namespace ttk;
 vtkStandardNewMacro(ttkScalarFieldCriticalPoints);
 
 ttkScalarFieldCriticalPoints::ttkScalarFieldCriticalPoints() {
-
+#ifdef TTK_ENABLE_MPI
+  hasMPISupport_ = true;
```
julien-tierny: I'd recommend to move that to the constructor of the *base* class when available.
in this example `core/base/scalarFieldCriticalPoints/`

core/vtk/ttkScalarFieldSmoother/ttkScalarFieldSmoother.cpp
```diff
@@ -17,6 +17,9 @@ vtkStandardNewMacro(ttkScalarFieldSmoother);
 ttkScalarFieldSmoother::ttkScalarFieldSmoother() {
   this->SetNumberOfInputPorts(1);
   this->SetNumberOfOutputPorts(1);
+#ifdef TTK_ENABLE_MPI
+  hasMPISupport_ = true;
```
julien-tierny: I'd recommend to move that to the constructor of the *base* class when available.
in this example `core/base/scalarFieldSmoother/`

julien-tierny (changes requested):
thanks @eve-le-guillou !
could you please consider the remarks from my review?
thanks!

julien-tierny:
ok, looks good. thanks!

PR #833, opened 2022-09-23
GitHub Actions: Enforce correct execution of Python scripts

pierre-guillou:
This PR splits the CI steps that concerns ttk-data Python scripts in two:
1. the first step is to check that all Python scripts run without segfaults, otherwise it will stop the CI with an error
2. the second step will check the outputs of the Python scripts and compare them to the hashes database. If a difference is detected, the current behavior applies: the CI will exit without an error (`continue-on-error: true`), but the workflow summary will still display the errors.

The `geometryApproximation` Python script has been troublesome on Windows, since no OpenGL implementation was available on GitHub's VMs. To avoid segfaults in this particular case, a new check has been added to the `CinemaImaging` module.

Some non-deterministic state file outputs were disabled to avoid overloading the CI servers.

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -154,12 +154,16 @@ jobs:
             # Ubuntu 22.04 performs more iterations in
             # PersistenceDiagramClustering than older Ubuntu VMs
             rm -f output_screenshots/persistenceDiagramClustering*.png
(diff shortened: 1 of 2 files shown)


julien-tierny:
wonderful, thanks!

PR #834, opened 2022-09-29
Fixes for DiscreteMorseSandwich in ttk-data

pierre-guillou:
This PR contains fixes that allow the FTM to DMS switch in the ttk-data state files and Python scripts. This is the companion PR to https://github.com/topology-tool-kit/ttk-data/pull/134.

Among those fixes:
* a new option for the `PersistenceDiagram` filter allows to clear the `DiscreteGradient` cache after the computation in order to release memory;
* a segfault in libc++ `std::sort` has been fixed in `DiscreteMorseSandwich`;
* `TrackingFromFields` now uses `DiscreteMorseSandwich` instead of FTM;
* the `Laplacian` module now correctly uses the calling module thread numbers and a critical section in the main parallel loop has been removed, making the filter more deterministic;
* checks were added for the `TTK_DCG_OPTIMIZE_MEMORY` option of the DiscreteGradient module, ensuring that only valid local simplex indices were passed to the triangulation;
* for the GitHub Actions Windows job, the `ctBones.vti` dataset has been resampled to reduce it size (the VMs don't have enough RAM);
* the GitHub Actions CI now executes the Python scripts in parallel.

This PR also includes a change in the internal structure of the Persistence Diagram: the `persistence` class member has been removed; instead, an accessor returns the `death - birth` in order to enforce this invariant.

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -182,8 +182,6 @@ jobs:
       run: |
         cd ttk-data
         python3 -u python/run.py
-      env:
-        OMP_NUM_THREADS: 1
 
     - name: Test ttk-data Python scripts results [NOT ENFORCED]
       continue-on-error: true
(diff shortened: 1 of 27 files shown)


julien-tierny:
ok, thanks. I'll double check carefully the ttk-data PR later (hopefully today).

PR #836, opened 2022-10-05
PersistenceCurve: Take PersistenceDiagram as input

pierre-guillou:
This PR does breaking changes to the `PersistenceCurve` filter: now, it applies only on Persistence Diagrams (instead of computing them internally). This allows to get rid of a lot of code (which was in fact a duplicate of `PersistenceDiagram`).

Some checks in `ttkPersistenceDiagramUtils` have been moved up since segfaults could occur before hitting them...

A corresponding ttk-data PR should follow shortly.

Enjoy,
Pierre

### DIFF
--- a/core/base/persistenceCurve/CMakeLists.txt
+++ b/core/base/persistenceCurve/CMakeLists.txt
@@ -4,7 +4,6 @@ ttk_add_base_library(persistenceCurve
   HEADERS
     PersistenceCurve.h
   DEPENDS
-    discreteGradient
-    triangulation
-    ftmTreePP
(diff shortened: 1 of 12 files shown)


julien-tierny:
I believe I should wait for a companion PR on ttk-data before merging that it, right?

pierre-guillou:
The corresponding ttk-data PR conflicts with https://github.com/topology-tool-kit/ttk-data/pull/134 so I'll do it after the latter is merged.

julien-tierny:
wonderful, thanks!

PR #837, opened 2022-10-05
GitHub Actions: Fasten Windows builds with sccache

pierre-guillou:
This PR uses [sccache](https://github.com/mozilla/sccache) in the Windows jobs of the `test_build ` and `store_ccache` workflows to cache generated binary files between workflow runs. This is useful to reduce the time spent waiting for the CI to complete.

To do so, CMake must use Ninja as its generator. To keep OpenMP multithreading working, the `clang-cl` compiler is manually set and compiler flags are edited in the generated `build.ninja` file.
Binary examples also need to be built with the same compiler and flags in order to prevent mixing the LLVM and the MSVC OpenMP libraries.

This compiler change did cause an unexpected segfault in the `LDistanceMatrix` filter (through the `persistentGenerators_householdAnalysis.py` script) that involves two nested parallel loops: the inner in `LDistance` performs a reduction over two datasets and the outer in `LDistanceMatrix` is parallel over the (upper) distance matrix. Although the inner loop parallelism should be disabled (`LDistance::threadNumber_` is set to 1), a segfault occurs because, among others, the outer loop uses the dynamic schedule. This PR fixes the segfault by switching back to the default schedule for the outer loop.

The `store_ccache` workflow now automatically deletes the `ccache` Release of the current repository before creating it.
Since the default graphviz package provided by conda is already at 2.5.0, there is no need to force its version anymore.

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/ccache.yml
+++ b/.github/workflows/ccache.yml
@@ -275,6 +275,19 @@ jobs:
     needs: [ccache-ubuntu, ccache-macos, ccache-windows]
     steps:
 
+    - name: Delete previous release
+      uses: actions/github-script@v6
+      continue-on-error: true
+      with:
+        script: |
+          const { owner, repo } = context.repo
+          const { data: { id } } = await github.rest.repos.getReleaseByTag({
+            owner,
+            repo,
+            tag: "ccache"
+          })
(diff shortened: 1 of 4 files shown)


julien-tierny:
looks good, thanks!

PR #838, opened 2022-10-06
ExplicitTriangulation: Add ASCII writer

pierre-guillou:
This PR complements https://github.com/topology-tool-kit/ttk/pull/598 with an ASCII backend for the `TriangulationWriter` module. Be aware that there is no corresponding reader in TTK for this file particular format. The format version of this particular writer is set to 2 (the binary writer is still at version 1).

The writer basically exports the internal state of the ExplicitTriangulation into a file (one value per line). Sections are prefixed by the name of the corresponding ExplicitTriangulation class member. The non-owned data structure `cellArray` (which decomposes VTK cells into vertices) is also included.

To better accommodate between `std::vector<std::array>>` and `FlatJaggedArray`, some methods of the latter have been slightly modified to return `size_t` instead of `SimplexId`.

Additionally, a MSVC warning in `OneSkeleton` has been fixed by using templated functions instead of if statements.

Enjoy,
Pierre

### DIFF
--- a/core/base/common/FlatJaggedArray.h
+++ b/core/base/common/FlatJaggedArray.h
@@ -66,11 +66,11 @@ namespace ttk {
     }
 
     struct Slice {
+      const size_t len;
       const SimplexId *const ptr;
-      const SimplexId len;
-      inline SimplexId operator[](const SimplexId id) {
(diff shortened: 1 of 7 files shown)


julien-tierny:
great, thanks!

PR #840, opened 2022-10-10
[MPI] MPI preconditioning will execute when using 1 process

eve-le-guillou:
For now, the MPI preconditioning only executes when TTK is executed on more than one process. This means that when TTK is compiled with MPI, it won't execute correctly on one process, preventing the output for one and more than one process to be strictly identical.
In this pull request, we deleted the tests preventing the preconditioning, as well as modified the filter ScalarFieldCriticalPoints, to ensure that the output will always be identical, regardless of the number of processes. When TTK is used as standalone, instead of checking if the process is executed on more than one process, we check if MPI has been initialized. This will always be true when using TTK with Paraview.

### DIFF
--- a/core/base/explicitTriangulation/ExplicitTriangulation.cpp
+++ b/core/base/explicitTriangulation/ExplicitTriangulation.cpp
@@ -309,6 +309,7 @@ int ExplicitTriangulation::preconditionEdgesInternal() {
 
 #ifdef TTK_ENABLE_MPI
     if(this->getDimensionality() == 2 || this->getDimensionality() == 3) {
+      this->preconditionTriangleEdges();
       return this->preconditionDistributedEdges();
     }
 #endif // TTK_ENABLE_MPI
--- a/core/base/explicitTriangulation/ExplicitTriangulation.cpp
+++ b/core/base/explicitTriangulation/ExplicitTriangulation.cpp
@@ -591,7 +591,9 @@ int ExplicitTriangulation::preconditionDistributedCells() {
   if(this->hasPreconditionedDistributedCells_) {
     return 0;
   }
-
+  if(!ttk::hasInitializedMPI()) {
+    return -1;
+  }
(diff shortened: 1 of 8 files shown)


core/base/explicitTriangulation/ExplicitTriangulation.cpp
```diff
@@ -309,6 +309,7 @@ int ExplicitTriangulation::preconditionEdgesInternal() {
 
 #ifdef TTK_ENABLE_MPI
     if(this->getDimensionality() == 2 || this->getDimensionality() == 3) {
+      this->preconditionTriangleEdges();
```
julien-tierny: is this call really needed?

if this call is indeed required in distributed mode, then, I would suggest to move it inside the call `preconditionDistributedEdges()`.

julien-tierny: moreover, if this call is indeed needed, it should be needed in the other triangulation types (implicit, etc.)
if this is the case, thanks for adding it to the `preconditionEdgesInternal` of the implicit triangulation.
(please remind me to discuss the periodic implicit triangulation and the implicit triangulations with preconditions)

julien-tierny (changes requested):
thanks for addressing my one remark up there.
thanks.

core/base/implicitTriangulation/ImplicitTriangulation.cpp
```diff
@@ -3414,6 +3414,10 @@ int ttk::ImplicitTriangulation::preconditionDistributedEdges() {
     return -3;
   }
 
+  if(this->getDimensionality() == 2) {
```
julien-tierny, line 3417: please do the same in ImplicitTriangulationWithPrecondition

julien-tierny (changes requested):
thanks

julien-tierny:
ok, thanks!

PR #842, opened 2022-10-13
Optimize power function with templates, improve kdTree performance

pierre-guillou:
This PR optimizes modules that uses power functions with a "stable" integer exponent (the exponent doesn't change during the computation) by passing lambdas to the call site instead of calling `Geometry::pow`/`Geometry::powInt`. The lambdas are inlined and allow to skip the `if` statement in `Geometry::powInt`. So far, only `LDistance` and `KDTree` benefit from this optimization, which might be applicable to `BottleneckDistance` too.

The `KDTree` module has been overhauled. Duplicate code inside `buidWeights.h` has been merged back into `KDTree.h`. Copies and heap allocations were removed, methods have been inlined. To still allow for usability outside the context of Persistence Diagrams (5D: birth, death + some 3D coords), the class has been templated over the container type (previously `std::vector<double>`, now `std::array<double, 5>` for Persistence Diagrams).

With this PR, `LDistance` and `PersistenceDiagramDistanceMatrix` should be resp. around 15% and 2.5 times faster. 

Enjoy,
Pierre


julien-tierny:
ok, looks good thanks!

PR #844, opened 2022-10-17
CompactTriangulation: Fix segfault by resizing vectors

pierre-guillou:
The `CompactTriangulation::initCache` method is called when creating the compact triangulation using the `TriangulationManager` filter. The `caches_` and `cacheMaps_` member variables were not initialized (they were empty), causing a segfault when trying to request the compact triangulation, for instance with `ScalarFieldCriticalPoints` applied after `TriangulationManager`.

This PR fixes the segfault.

Enjoy,
Pierre


julien-tierny:
awesome, thanks for the hotfix!

PR #846, opened 2022-10-25
GitHub Actions: Fix macOS 12/OpenMP compatibility

pierre-guillou:
Following the LLVM 15 release, the brew-provided `libomp` became somewhat incompatible with Apple Clang (the default macOS compiler).
As a consequence, this PR switches the compiler to use directly Clang/LLVM 15 instead of Apple Clang.

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -268,6 +268,11 @@ jobs:
         cmake --build . --parallel
         sudo cmake --build . --target install
 
+    - name: Set PYTHONPATH for macOS pvpython
+      run: |
+        # pvpython does not embed the correct PYTHONPATH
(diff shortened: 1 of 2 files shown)


julien-tierny:
thanks for the PR!
it seems that some checks are still failing though.
maybe it is time to disable any appveyor ci...

pierre-guillou:
There's still a Python dependency problem on macOS: default Python is (now) 3.10 but the ParaView headless binaries were built with Python 3.9. The #820 workaround doesn't work anymore. The ParaView headless packages need to be rebuilt in order to better fix this dependency issue.

julien-tierny:
thanks, I rebuilt the headless packages and restarted failed CI jobs.

julien-tierny:
it seems like there's still an issue with some environment variable (the ci cannot find vtk's python support when running the `examples/python` test).

julien-tierny:
great, thanks!

PR #847, opened 2022-10-26
[MPI] Add custom global id generator in MPI preconditioning

eve-le-guillou:
In this PR is added a custom global cell and point id generator for PolyData and ImageData.

The current VTK filter `vtkGenerateGlobalIds` suffers from a bug at the moment (Paraview 5.10.1) that prevents its execution for data sets of 1024^3 and over. The filter Identifier implements a global id generator that works for data sets of all dimensions and produces global identifiers that are continuous at least on each rank.

It is triggered during MPI preconditioning, after the potential generation of ghost simplices. If global ids are already defined, it is checked if they are valid. Valid identifiers are identifiers for which the highest global id is equal to the number of identifiers minus one and for which the lowest global id is 0. If the global ids are undefined or invalid, the generation is triggered.

The generation is different depending on the type of data set. For an ImageData, the computation is done by communicating at the beginning the bounds of the global data set. From this is determined the offset of each simplex. The result of this computation will always be the same, regardless of the number of processes.

For a PolyData, the global ids are generated locally using the computation of a prefix sum on all processes to compute an offset, ensuring that each global id is unique. The result of this computation will depend on the number of processes, as the value of the offset will depend on the number of processes. Then, ghost simplices are exchanged to be matched to the generated global ids. This implementation takes advantage of previous invalid global ids or RankArray to accelerate the matching of ghosts.

No action is required from the user to trigger the generation of global ids. However, if the user wishes to use the VTK filter `vtkGenerateGlobalIds` for their pipeline, it is possible, it simply needs to be done before the use of a TTK filter. TTK will then simply retrieve the global ids computed by the VTK filter, as long as they are valid.

In case TTK is not compiled with `TTK_ENABLE_MPI=ON`, the global ids will not be computed automatically, as the MPI preconditioning is not performed. 

Calls to `vtkDataSet::GetGlobalIds` have been replaced by `vtkDataSet::GetArray(“Global[Cell/Point]Ids”)`, as the output array no longer is recognized by Paraview as a true global id array.

No standalone was created for this filter, as it felt not very appropriate for this feature.

A warning has also been added during the MPI preconditioning when producing a RankArray or Global Ids for a Poly Data, as at the moment (Paraview 5.10.1) there are a few bug in the distribution of the data, specifically in ghost simplices ownership, that can cause the RankArray and Global Ids to be partly incorrect. It is not the case for ImageData data sets.


julien-tierny:
could you please sync this PR (git pull the public ttk dev branch) to retrieve the CI fix (https://github.com/topology-tool-kit/ttk/pull/846 )?

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -428,18 +429,213 @@ int ttkAlgorithm::RequestDataObject(vtkInformation *ttkNotUsed(request),
...
+    status = identifiers.executeSequential();
+  }
+
+  if(status < 1) {
+    printErr("Global identifier generation failed");
+    return;
+  }
+#endif
+  // Remove outdated global Ids
+  input->GetPointData()->RemoveArray("GlobalPointIds");
```
julien-tierny: I believe you want to use the const char `ttk::VertexScalarFieldName` instead of writing down (multiple times, error-prone) the const `"GlobalPointIds"`.

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -428,18 +429,213 @@ int ttkAlgorithm::RequestDataObject(vtkInformation *ttkNotUsed(request),
...
+  input->GetCellData()->RemoveArray("GlobalCellIds");
```
julien-tierny: same as above but with the const char buffer `ttk::CellScalarFieldName`

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
julien-tierny: Thanks for systematically using these pre-defined strings throughout.

core/vtk/ttkGhostCellPreconditioning/ttkGhostCellPreconditioning.cpp
```diff
@@ -64,12 +64,12 @@ int ttkGhostCellPreconditioning::RequestData(
   this->printMsg("#Points: " + std::to_string(nVertices));
   this->printMsg("#Cells: " + std::to_string(nCells));
 
-  auto *verticesGlobalIds
-    = ttkUtils::GetPointer<ttk::LongSimplexId>(pointData->GetGlobalIds());
+  auto *verticesGlobalIds = ttkUtils::GetPointer<ttk::LongSimplexId>(
+    pointData->GetArray("GlobalPointIds"));
```
julien-tierny: see remark about pre-defined const char arrays.

core/vtk/ttkGhostCellPreconditioning/ttkGhostCellPreconditioning.cpp
```diff
@@ -64,12 +64,12 @@ int ttkGhostCellPreconditioning::RequestData(
...
   auto *verticesGhostCells
     = ttkUtils::GetPointer<unsigned char>(pointData->GetArray("vtkGhostType"));
-  auto *cellsGlobalIds
-    = ttkUtils::GetPointer<ttk::LongSimplexId>(cellData->GetGlobalIds());
+  auto *cellsGlobalIds = ttkUtils::GetPointer<ttk::LongSimplexId>(
+    cellData->GetArray("GlobalCellIds"));
```
julien-tierny: see remark about pre-defined const char arrays.

core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
```diff
@@ -82,7 +82,7 @@ int ttkArrayPreconditioning::RequestData(vtkInformation *ttkNotUsed(request),
     }
   }
 
-  auto vtkGlobalPointIds = pointData->GetGlobalIds();
+  auto vtkGlobalPointIds = pointData->GetArray("GlobalPointIds");
```
julien-tierny: see remark about pre-defined const char arrays.

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -514,7 +764,7 @@ void ttkAlgorithm::MPITriangulationPreconditioning(
     // provide "GlobalPointIds" & "RankArray" point data arrays to the
     // triangulation
     triangulation->setVertsGlobalIds(
-      ttkUtils::GetPointer<ttk::LongSimplexId>(pd->GetGlobalIds()));
+      ttkUtils::GetPointer<ttk::LongSimplexId>(pd->GetArray("GlobalPointIds")));
```
julien-tierny: see remark about pre-defined const char arrays.

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -452,19 +648,74 @@ void ttkAlgorithm::MPIPipelinePreconditioning(vtkDataSet *input) {
...
-      ttk::produceRankArray(rankArray,
-                            ttkUtils::GetPointer<ttk::LongSimplexId>(
-                              input->GetPointData()->GetGlobalIds()),
-                            ttkUtils::GetPointer<unsigned char>(
-                              input->GetPointData()->GetArray("vtkGhostType")),
-                            vertexNumber, boundingBox);
+      ttk::produceRankArray(
+        rankArray,
+        ttkUtils::GetPointer<ttk::LongSimplexId>(
+          input->GetPointData()->GetArray("GlobalPointIds")),
```
julien-tierny: see remark about pre-defined const char arrays.

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -452,19 +648,74 @@ void ttkAlgorithm::MPIPipelinePreconditioning(vtkDataSet *input) {
...
+               "incorrect for those simplices");
+      printWrn(
+        "Beware when using TTK for such data set types, some results may "
+        "be false");
+    }
+  }
+
+  // Checks if global ids are valid
+  ttk::LongSimplexId *globalPointIds = ttkUtils::GetPointer<ttk::LongSimplexId>(
+    input->GetPointData()->GetArray("GlobalPointIds"));
```
julien-tierny: see remark about pre-defined const char arrays.

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -452,19 +648,74 @@ void ttkAlgorithm::MPIPipelinePreconditioning(vtkDataSet *input) {
     input->GetCellData()->AddArray(
       generator->GetOutputDataObject(0)->GetGhostArray(1));
   }
+  ttk::SimplexId vertexNumber = input->GetNumberOfPoints();
+  ttk::SimplexId cellNumber = input->GetNumberOfCells();
+  if((input->GetDataObjectType() == VTK_POLY_DATA
+      || input->GetDataObjectType() == VTK_UNSTRUCTURED_GRID)) {
+    if((input->GetCellData()->GetArray("GlobalCellIds") == nullptr)
+       || (input->GetPointData()->GetArray("GlobalPointIds") == nullptr)) {
```
julien-tierny: see remark about pre-defined const char arrays.

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -428,18 +429,213 @@ int ttkAlgorithm::RequestDataObject(vtkInformation *ttkNotUsed(request),
...
+  input->GetPointData()->RemoveArray("GlobalPointIds");
+  input->GetCellData()->RemoveArray("GlobalCellIds");
+
+  // Create VTK objects and add them to the data set
+
+  vtkSmartPointer<vtkIdTypeArray> vtkVertexIdentifiers
+    = vtkSmartPointer<vtkIdTypeArray>::New();
+  vtkSmartPointer<vtkIdTypeArray> vtkCellIdentifiers
+    = vtkSmartPointer<vtkIdTypeArray>::New();
+  vtkVertexIdentifiers->SetName("GlobalPointIds");
```
julien-tierny: see remark about pre-defined const char arrays.

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -428,18 +429,213 @@ int ttkAlgorithm::RequestDataObject(vtkInformation *ttkNotUsed(request),
...
+  double *boundingBox = input->GetBounds();
+  identifiers.setBounds(boundingBox);
+  if(ttk::isRunningWithMPI()) {
+    switch(input->GetDataObjectType()) {
+      case VTK_UNSTRUCTURED_GRID:
+      case VTK_POLY_DATA: {
+
+        identifiers.setOutdatedGlobalPointIds(
+          ttkUtils::GetPointer<ttk::LongSimplexId>(
+            input->GetPointData()->GetArray("GlobalPointIds")));
```
julien-tierny: see remark about pre-defined const char arrays.

julien-tierny: see remark about pre-defined const char arrays.

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -181,7 +181,8 @@ vtkDataArray *ttkAlgorithm::GetOrderArray(vtkDataSet *const inputData,
         this->MPIPipelinePreconditioning(inputData);
       }
       if(ttk::isRunningWithMPI()) {
-        auto vtkGlobalPointIds = inputData->GetPointData()->GetGlobalIds();
+        auto vtkGlobalPointIds
+          = inputData->GetPointData()->GetArray("GlobalPointIds");
```
julien-tierny: see remark about pre-defined const char arrays.

core/vtk/ttkAlgorithm/ttkAlgorithm.h
```diff
@@ -216,6 +216,25 @@ class TTKALGORITHM_EXPORT ttkAlgorithm : public vtkAlgorithm,
    */
   void MPIPipelinePreconditioning(vtkDataSet *input);
 
+  /**
+   * This method checks the validity of the global identifiers given in
+   * argument. A set of global identifiers is valid if the highest global id is
+   * equal to the global number of simplices (without ghosts) + 1 and if the
```
julien-tierny: "number of simplices + 1"
two things:
- isn't it "-1" instead of "+1"?
- "simplices" is ambiguous here. are you checking the vertex identifiers? (if so "simplices" should be "vertices"). if you are checking the cell identifiers, "simplices" should "simplices of highest dimension".

julien-tierny: or maybe this is a generic function that covers both cases? (thanks for clarifying)

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -428,18 +429,213 @@ int ttkAlgorithm::RequestDataObject(vtkInformation *ttkNotUsed(request),
...
+
+  vtkCellIdentifiers->SetName("GlobalCellIds");
+  vtkCellIdentifiers->SetNumberOfComponents(1);
+  vtkCellIdentifiers->SetNumberOfTuples(input->GetNumberOfCells());
+
+#ifdef TTK_ENABLE_OPENMP
+#pragma omp parallel for
+#endif
+  for(ttk::SimplexId i = 0; i < vertexNumber; i++) {
+    vtkVertexIdentifiers->SetTuple1(i, vertexIdentifiers[i]);
```
julien-tierny: that copy can be avoided!
simply declare the vtk version of that array at the very beginning and allocate it appropriately.
then you can pass a pointer to its content to the base code class and then the base code class will simply fill up the buffer pointed by the provided pointer.

julien-tierny: see `core/vtk/ttkScalarFieldSmoother/ttkScalarFieldSmoother.cpp:94` for an example.

julien-tierny: please use this recommended strategy throughout your PR.
(i.e. when you can know in advance the size of an output buffer, just allocate it beforehand with a VTK data structure and pass a raw core pointer to the base class).

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -428,18 +429,213 @@ int ttkAlgorithm::RequestDataObject(vtkInformation *ttkNotUsed(request),
...
+        unsigned char *cellGhost = ttkUtils::GetPointer<unsigned char>(
+          input->GetCellData()->GetArray("vtkGhostType"));
+        identifiers.setCellGhost(cellGhost);
+        vtkPointSet *pointSet = vtkPointSet::SafeDownCast(input);
+        identifiers.setPointSet(static_cast<float *>(
+          ttkUtils::GetVoidPointer(pointSet->GetPoints())));
+        vtkCellArray *cells = nullptr;
+        switch(input->GetDataObjectType()) {
+          case VTK_UNSTRUCTURED_GRID: {
+            auto dataSetAsUG = static_cast<vtkUnstructuredGrid *>(input);
```
julien-tierny: I believe it is recommended to use `vtkUnstructuredGrid::SafeDownCast` instead of `static_cast`

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -428,18 +429,213 @@ int ttkAlgorithm::RequestDataObject(vtkInformation *ttkNotUsed(request),
...
+          input->GetPointData()->GetArray("RankArray")));
+        identifiers.setCellRankArray(ttkUtils::GetPointer<ttk::SimplexId>(
+          input->GetCellData()->GetArray("RankArray")));
+        identifiers.setVertGhost(ttkUtils::GetPointer<unsigned char>(
+          input->GetPointData()->GetArray("vtkGhostType")));
+        unsigned char *cellGhost = ttkUtils::GetPointer<unsigned char>(
+          input->GetCellData()->GetArray("vtkGhostType"));
+        identifiers.setCellGhost(cellGhost);
+        vtkPointSet *pointSet = vtkPointSet::SafeDownCast(input);
+        identifiers.setPointSet(static_cast<float *>(
```
julien-tierny, line 535: is the cast needed here?

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -428,18 +429,213 @@ int ttkAlgorithm::RequestDataObject(vtkInformation *ttkNotUsed(request),
...
+          ttkUtils::GetVoidPointer(pointSet->GetPoints())));
+        vtkCellArray *cells = nullptr;
+        switch(input->GetDataObjectType()) {
+          case VTK_UNSTRUCTURED_GRID: {
+            auto dataSetAsUG = static_cast<vtkUnstructuredGrid *>(input);
+            cells = dataSetAsUG->GetCells();
+            break;
+          }
+          case VTK_POLY_DATA: {
+            auto dataSetAsPD = static_cast<vtkPolyData *>(input);
```
julien-tierny: `SafeDownCast`

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -428,18 +429,213 @@ int ttkAlgorithm::RequestDataObject(vtkInformation *ttkNotUsed(request),
...
+              return;
+            }
+          } else {
+            this->printErr(
+              "Cannot convert the provided cell array to 64-bit storage");
+            return;
+          }
+        }
+
+        identifiers.setConnectivity(static_cast<ttk::LongSimplexId *>(
```
julien-tierny: cast needed?

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -480,13 +731,12 @@ void ttkAlgorithm::MPIPipelinePreconditioning(vtkDataSet *input) {
...
   // If the RankArray array doesn't exist for celldata, it is created
   if(input->GetCellData()->GetArray("RankArray") == nullptr) {
-    int cellNumber = input->GetNumberOfCells();
     std::vector<int> cellsRankArray(cellNumber, 0);
     if(ttk::isRunningWithMPI()) {
       double *boundingBox = input->GetBounds();
       ttk::produceRankArray(cellsRankArray,
                             ttkUtils::GetPointer<ttk::LongSimplexId>(
-                              input->GetCellData()->GetGlobalIds()),
+                              input->GetCellData()->GetArray("GlobalCellIds")),
```
julien-tierny: see remark about pre-defined const char arrays.

core/vtk/ttkIdentifiers/ttkIdentifiers.h
```diff
@@ -1,31 +1,24 @@
 /// \ingroup vtk
 /// \class ttkIdentifiers
-/// \author Julien Tierny <julien.tierny@lip6.fr>
-/// \date August 2015.
+/// \author Eve Le Guillou <eve.le-guillou@lip6.fr>
```
julien-tierny, line 3: in principle (mostly for future reference), do not erase the name of previous contributors but just add yours to the list.

core/base/identifiers/Identifiers.h
```diff
@@ -0,0 +1,959 @@
...
+
+    ~Identifiers() override = default;
+
+  protected:
+    ttk::SimplexId vertexNumber_{};
+    ttk::SimplexId cellNumber_{};
+    std::vector<ttk::SimplexId> *vertexIdentifiers_;
+    std::vector<ttk::SimplexId> *cellIdentifiers_;
+#ifdef TTK_ENABLE_MPI
+    std::map<ttk::SimplexId, ttk::SimplexId> vertGtoL_;
```
julien-tierny: the (abstract) triangulation class already contains such a map (`vertexGidToLid_`).
couldn't you use this map instead? (that way it will be filled once for all -- currently, I have the impression that information is filled twice: once here, once later when attaching the result of this computation to the triangulation).

core/base/identifiers/Identifiers.h
```diff
@@ -0,0 +1,959 @@
...
+    unsigned char *vertGhost_{nullptr};
+    unsigned char *cellGhost_{nullptr};
+    std::vector<std::vector<ttk::SimplexId>> pointsToCells_;
+    float *pointSet_;
+    ttk::LongSimplexId *connectivity_;
+    ttk::LongSimplexId *outdatedGlobalPointIds_{nullptr};
+    ttk::LongSimplexId *outdatedGlobalCellIds_{nullptr};
+    std::map<ttk::LongSimplexId, ttk::SimplexId> vertOutdatedGtoL_;
+    std::map<ttk::LongSimplexId, ttk::SimplexId> cellOutdatedGtoL_;
+    KDTree<float, std::array<float, 3>> kdt_;
```
julien-tierny: could you write `ttk::KDTree` to be more explicit (and consistent)? thanks!

julien-tierny (changes requested):
alright, thanks a lot @eve-le-guillou !
this is very clean!
could you simply address my few comments?
thanks a lot!

julien-tierny:
@pierre-guillou this one looks really nearly to go (cf my email from october 13, regarding our earlier discussions regarding implicit identifiers for simplices of intermediate dimensions)

core/base/identifiers/Identifiers.h
eve-le-guillou: Here, the triangulation is not created yet, therefore vertexGidToLid_ doesn't exist.
Indeed, the map is created twice, but I don't know how to transfer the map from identifier to a triangulation that isn't created yet.

core/vtk/ttkAlgorithm/ttkAlgorithm.h
eve-le-guillou: Indeed, it is -1, I made a mistake there. This function is a generic function that covers both cases. I will update the comments to make it clearer.

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
eve-le-guillou: I believe it is, as the function ttkUtils::GetPointer cannot be used here, since it only take vtkDataArray as argument.

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
julien-tierny: outdated when `SetGlobalIds` is merged in the PR.

julien-tierny:
Use `SetGlobalIds()` to set the newly created array (with a one-time string).
Remove the const string from `DataTypes.h`

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
julien-tierny: `SetGlobalIds` && `GetGlobalIds`

core/vtk/ttkGhostCellPreconditioning/ttkGhostCellPreconditioning.cpp
julien-tierny: `SetGlobalIds` && `GetGlobalIds`

core/vtk/ttkGhostCellPreconditioning/ttkGhostCellPreconditioning.cpp
julien-tierny: `SetGlobalIds` && `GetGlobalIds`

core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
julien-tierny: `SetGlobalIds` && `GetGlobalIds`

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
julien-tierny: `SetGlobalIds` && `GetGlobalIds`

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
julien-tierny: `SetGlobalIds` && `GetGlobalIds`

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
julien-tierny: `SetGlobalIds` && `GetGlobalIds`

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
julien-tierny: `SetGlobalIds` && `GetGlobalIds`

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
julien-tierny: `SetGlobalIds` && `GetGlobalIds`

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
julien-tierny: `SetGlobalIds` && `GetGlobalIds`

julien-tierny: `SetGlobalIds` && `GetGlobalIds`

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
julien-tierny: `SetGlobalIds` && `GetGlobalIds`

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
julien-tierny: Todo: `GetPointer()`

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
julien-tierny: Todo: `GetPointer()`

julien-tierny:
Use `ttkUtils::GetPointer()` instead of casting (simpler to read)

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
julien-tierny: SetGlobalIds && GetGlobalIds

julien-tierny:
- Revise `MPIPipelinePreconditioning` to remove the GhostCellGeneration.
- Do the GhostCellGeneration before the allocation by the factory
- Do the rest of  `MPIPipelinePreconditioning` with the triangulation as an argument (use setters and getters to access the map, update creating code in triangulation)

core/base/identifiers/Identifiers.h
julien-tierny: see 4th point in the comments below.

eve-le-guillou:
`MPIPipelinePreconditioning` may be called when the triangulation has not been created (see ttkAlgorithm::GetOrderArray), therefore is that case, the map for global id to local id is not set in the triangulation.
Otherwise, it is passed to the triangulation using a setter.

julien-tierny:
> `MPIPipelinePreconditioning` may be called when the triangulation has not been created (see ttkAlgorithm::GetOrderArray), therefore is that case, the map for global id to local id is not set in the triangulation. Otherwise, it is passed to the triangulation using a setter.

@eve-le-guillou we'd need to discuss this I'm not sure I understand.
Is `MPIPipelinePreconditioning` called twice in a single pipeline?

I was actually about to suggest to merge the two functions `MPIPipelinePreconditioning` and `MPITriangulationPreconditioning` into a single function: they have the same arguments and they're only meant to be called one after the other, only once at the beginning of the pipeline.

what do you think?

(NB: it seems that the argument `triangulation` is used several times in `MPIPipelinePreconditioning` without checking if it's `nullptr` or not)

eve-le-guillou:
> Is `MPIPipelinePreconditioning `called twice in a single pipeline?

MPIPipelinePreconditioning is called in GetTriangulation, but it is also called in `GetOrderArray`. I don't recall adding it to `GetOrderArray`, maybe @pierre-guillou is the one who added it? It looks like it is there because the order array can be created without needing to create a triangulation. In which case,         `MPIGhostPipelinePreconditioning(inputData)` and `MPIPipelinePreconditioning` are called , but not `MPITriangulationPreconditioning`, to create the necessary data arrays.

Do you think the call to `MPIPipelinePreconditioning` are useless in `GetOrderArray`?

> I was actually about to suggest to merge the two functions MPIPipelinePreconditioning and MPITriangulationPreconditioning into a single function: they have the same arguments and they're only meant to be called one after the other, only once at the beginning of the pipeline.

Because of my remark above, this is not possible.

> (NB: it seems that the argument triangulation is used several times in MPIPipelinePreconditioning without checking if it's nullptr or not)

No, it is only used once, and the check is done. Triangulation is however used multiple times in MPITriangulationPreconditioning without test, but this method is only called if the triangulation exists, therefore it is safe.

pierre-guillou:
Git blame tells me that commit 74ec4301dbe7 is the one that introduced `MPIPipelinePreconditioning` in `GetOrderArray`.

julien-tierny:
thanks @eve-le-guillou 
I can see now that `MPIPipelinePreconditioning` could also be called independently from `GetOrderArray` (and I agree that some filters may need to call this function without handling a triangulation). 
so it is indeed important that `MPIPipelinePreconditioning` is called from both functions.
However, we need to make sure that if this function is run twice within a single pipeline, nothing gets recomputed upon the second call. Is that the case?

eve-le-guillou:
Yes, it is the case.

On the second execution, the software will detect that the necessary arrays are present, and won't compute them again. 
However, with regards to class attributes such as `vertexGidToLid_`, if it is computed when the triangulation hasn't been created, it won't be saved in the triangulation. It will only be saved in the triangulation if the triangulation exists.

But all other arrays added in `MPIPipelinePreconditioning` will always be saved for the whole pipeline.

julien-tierny:
Consider the following pipeline:
- a first filter which calls `GetOrderArray`
- followed by a second filter which calls `GetTriangulation`

Can you double check that the map `vertexGidToLid_` will be correctly passed when `GetTriangulation` is called by the second filter?

eve-le-guillou:
No, in this case the map `vertexGidToLid_` will not be passed from the first filter to the second, it will be computed again, as the triangulation is used to store `vertexGidToLid_` through the pipeline. If the triangulation does not exist, then it will not be able to store the map.

The feature as we discussed doesn't not allow for that, it would require a different implementation.

julien-tierny:
Here, when the second filter is called, you want the call to `GetTriangulation` to make sure that the triangulation is indeed equipped with the map `vertexGidToLid_`. If that's not the case, you want to re-compute it. Do you see what I mean?
Thanks!

julien-tierny:
(in short, in the above scenario, you still want the map `vertexGidToLid_` to be filled in the triangulation, for the algorithms to work properly, even if that implies re-computing it).
if you detect that precise scenario (`GetOrderArray` has been called *before* `GetTriangulation`), you may want to print a warning in the command line, to mention that it would be more efficient to do it the other way around (in case a developer called these two functions in the wrong order in their filter)

eve-le-guillou:
> Here, when the second filter is called, you want the call to GetTriangulation to make sure that the triangulation is indeed equipped with the map vertexGidToLid_. If that's not the case, you want to re-compute it. Do you see what I mean?

Yes, it is clearer. This is indeed what happens, as the flag checking if the map has been computed or not is not changed when the triangulation is not created, so it does recompute it.

eve-le-guillou:
> if you detect that precise scenario (GetOrderArray has been called before GetTriangulation), you may want to print a warning in the command line, to mention that it would be more efficient to do it the other way around (in case a developer called these two functions in the wrong order in their filter)

I'm not sure how to detect this particular use case. I could maybe add this to the documentation of `GetOrderArray`. Would it be enough for you, as it is a note for developers?

julien-tierny:
Based on our discussion, thanks for revising the way the `vertexGlobalId2LocalId` is retrieved from the triangulation (with write-mode getter).

julien-tierny:
to test that the map is correctly filled in the above scenario, take ttkScalarFieldCriticalPoints, switch the calls `GetOrderArray` and `GetTriangulation` and make sure the map is correctly set (and that the critical points are well computed).
- `GetTriangulation` then `GetOrderArray`, the map should be computed upon `GetTriangulation` only
- `GetOrderArray` then `GetTriangulation`, the map should be re-computed upon `GetTriangulation` and valid afterwards.

eve-le-guillou:
Done, and tested!

julien-tierny:
Looks good, thanks a lot!

PR #848, opened 2022-10-31
DiscreteMorseSandwich: Add parallelism with locks for saddle-saddle pairs

pierre-guillou:
This PR updates the `DiscreteMorseSandwich::eliminateBoundariesSandwich` method to use the parallel algorithm described in the "Towards Lockfree Persistent Homology" article.

Due to their very late introduction in OpenMP, atomic compare-and-swap operations have been replaced by locks, without notable impact of performance.

There is no need for the first parallel pass any more. The 2-saddle boundaries now expand during the (only) parallel pass. The pairs are created once this pass is done.

Due to the slight increase in memory usage (a new vector of the size of the edges is used), the ctBones.vti data-set had to be resized for the CI to complete without errors on Ubuntu VMs.

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -128,6 +128,17 @@ jobs:
         path: "ttk-data"
       name: Checkout ttk-data
 
+    - name: Resample large datasets
+      shell: python3 {0}
+      working-directory: ./ttk-data
+      run: |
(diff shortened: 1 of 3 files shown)


julien-tierny:
fantastic, thanks!

PR #850, opened 2022-11-02
GitHub Actions: Update clang tools

pierre-guillou:
This PR updates the `check_code` workflow to use the latest Clang tools available in the Ubuntu 22.04 repositories (v14) instead of the pinned v11 packages. This also allows to use directly `clang-format`, `clang-check` and `clang-tidy` in the workflow instead of hiding them in environment variables, thus making it easier to copy-paste the commands into a terminal to use them locally.

Updating the tools revealed a new use-after-move warning in the move constructor of the Triangulation class (which does not seem to be used much in the TTK codebase). The PR fixes this warning and extends the constructors and operators= to use the implicit/periodic triangulations with preconditions. The if/else was replaced by a switch, as it emits a warning if an enumeration value is missing.

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -89,6 +89,8 @@ jobs:
           - {tidy: true, sa: false, dox: false}
           - {tidy: false, sa: true, dox: false}
           - {tidy: false, sa: false, dox: true}
+    env:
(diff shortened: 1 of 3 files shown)


julien-tierny:
thanks!

PR #853, opened 2022-11-07
Reduce build times of PersistenceDiagram modules

pierre-guillou:
This short PR reduces the build times of the `ttkPersistenceDiagram`, `ttkPersistenceDiagramApproximation` and `ttkTrackingFromFields` modules through these two optimizations:

- the computation of saddle-saddle pairs with the FTM backend has been deleted, since it is now superseded by the DiscreteMorseSandwich algorithm.
- `PersistenceDiagramApproximation` now calls only `executeApproximateTopology` in the base class instead of the way heavier `execute`. The filter has been made to be executed only on regular grids in order to get rid of its fallback to FTM.

The observed reduction in build times is as follow (on my computer with some template instantiations disabled):
* `ttkPersistenceDiagram`: from 180 to 110s,
* `ttkTrackingFromFields`: from 185 to 110s,
* `ttkPersistenceDiagramApproximation`: from 180 to 30s.

Enjoy,
Pierre

### DIFF
--- a/core/base/persistenceDiagram/PersistenceDiagram.h
+++ b/core/base/persistenceDiagram/PersistenceDiagram.h
@@ -159,9 +159,6 @@ namespace ttk {
 
     PersistenceDiagram();
 
-    inline void setComputeSaddleConnectors(bool state) {
-      ComputeSaddleConnectors = state;
-    }
     inline void setBackend(const BACKEND be) {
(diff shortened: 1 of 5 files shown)


julien-tierny:
looks good, thanks!

julien-tierny:
looks good, thanks!

PR #854, opened 2022-11-07
Update Github Actions

pierre-guillou:
This PR updates the GitHub Actions workflows:

* ZFP and Spectra have been updated to their latest release,
* the packaged version of Websocket++ is used on Ubuntu and macOS,
* the `actions/checkout`, `actions/download-artifact` and `actions/upload-artifact` were updated to their latest version (this should fix a NodeJS deprecation warning),
* the third-party action `dsaltares/fetch-gh-release-asset` has been replaced with a `wget` call,
* the deprecated `actions/create-release` has been replaced with some GitHub script,
* GitHub script is also used to delete existing releases before creating new ones with the same tag,
* the `ccache` workflow has been updated to reflect the last changes in the `test_build` workflow.

Enjoy,
Pierre

### DIFF
--- a/.github/actions/install-deps-unix/action.yml
+++ b/.github/actions/install-deps-unix/action.yml
@@ -7,20 +7,20 @@ runs:
     - name: Install Spectra dependency
       shell: bash
       run: |
-        wget https://github.com/yixuan/spectra/archive/v1.0.0.tar.gz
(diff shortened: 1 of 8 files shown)


julien-tierny:
ok, looks good. thanks!

PR #855, opened 2022-11-08
MergeTreeDistanceMatrix Persistence Diagram Support

MatPont:
This PR adds the support for persistence diagrams to the MergeTreeDistanceMatrix filter.

At the same time it fixes the weird behavior related to the custom backend. For example, before this PR, when we select the custom backend, uncheck the normalization option, execute, select the "Wasserstein Distance (IEEE TVCG 2021)" backend, execute and then go again with the custom backend, in the GUI the normalization will still be uncheck whereas the class will have this option set to true (due to the fact that we have used the Wasserstein backend just before). Meaning that the execution will use the normalization whereas the GUI would say that the normalization is uncheck hence misleading the user.


julien-tierny:
looks good, thanks!

PR #857, opened 2022-11-08
Use DiscreteMorseSandwich to return saddle connectors

pierre-guillou:
This PR updates the return saddle connectors algorithm in the MorseSmaleComplex module to use DiscreteMorseSandwich saddle-saddle pairs.

This new algorithm is very simple: since DiscreteMorseSandwich finds saddle-saddle pairs in ascending persistence order, the pairs are processed until the threshold is hit. For each pair, we find the saddle connector them reverse the v-path in the DiscreteGradient.

To avoid computing the DiscreteGradient twice (a first time for the MorseSmaleComplex and a second time for DiscreteMorseSandwich), move semantics is used to pass the DiscreteGradient instance in and out of the DiscreteMorseSandwich module. To allow it, the `contourTree_` member variable of the `DiscreteGradient` class has been made a (smart) pointer to provide an indirection since `FTMTree` instances cannot be copied or moved. This should have no impact on the rest of the code since after #853 and this PR, corresponding code should not be in use.

Enjoy,
Pierre

### DIFF
--- a/core/base/discreteGradient/DiscreteGradient.h
+++ b/core/base/discreteGradient/DiscreteGradient.h
@@ -669,6 +669,12 @@ in the gradient.
                             const triangulationType &triangulation);
 
       /**
+       * @brief Initialize/Allocate discrete gradient memory
+       */
+      void initMemory(const AbstractTriangulation &triangulation);
+
+    public:
+      /**
        * Compute the difference of function values of a pair of cells.
        */
       template <typename dataType, typename triangulationType>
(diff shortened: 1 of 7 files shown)


julien-tierny:
Could you please make sure that the persistence threshold for the saddle connector reversal is expressed as a fraction of the function span? (i.e. from 0 to 1)
At the moment, I have the impression that this is an absolute value.
Maybe you could do something similar to "TopologicalSimplificationByPersistence" (where the two options are offered based on a checkbox). Thanks!

julien-tierny:
looks good, thanks!

Pat-on-back:
could you give me the demo code? Thanks

PR #858, opened 2022-11-14
DiscreteGradient: Remove old reverseGradient code

pierre-guillou:
Following #853 and #857, this PR removes the now unused DiscreteGradient methods that computed the saddle-saddle pairs for FTM and returned them for MorseSmaleComplex.

Enjoy,
Pierre


julien-tierny:
looks good, thanks!

PR #859, opened 2022-11-14
Switch to C++14

pierre-guillou:
This PR switches the TTK codebase to use the C++14 standard rather than C++11. This is partly due to Boost requiring C++14 in a few releases.

In this PR, I also used one shiny new feature of the new standard: the function `std::make_unique` that replaces the call to the `std::unique_ptr` constructor (and avoids an explicit `new`).

@CharlesGueunet do you think this can impact the ParaView superbuild?
We are also thinking of switching to C++17. What's your though?

Enjoy,
Pierre

### DIFF
--- a/CMake/config.cmake
+++ b/CMake/config.cmake
@@ -1,6 +1,6 @@
 # --- Prerequisites
 
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 14)
 
 # --- Global Options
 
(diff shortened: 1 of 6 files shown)


CharlesGueunet:
For C++ 14, we already have VTK-m that use it. It may prevent few people of compiling TTK themselves but ... if you want to be on the state of the art, you should not stay with a 10 years old compiler.
For C++ 17, we may want to try it on a paraview superbuild branch to check if Linux / Mac and Windows build passes, I have no clue. If they pass, here again I see no real drawbacks for now on.

Once TTK will be more MPI aware, more people may be interested to use it on supercomputer, which are usually the bottleneck in terms of update. So on one side now is the good time to update, but on the other we may not want ask for gcc 12 if we want to be able to try TTK on HPC system soon :)

julien-tierny:
@eve-le-guillou let us know if you run into issues with this update!

PR #860, opened 2022-11-15
[MPI] Cache neighboring ranks in memory

eve-le-guillou:
Hi all,

This Pull Request uses the already existing attribute of the class `AbstractTriangulation` `neighborRanks_` to store neighboring ranks all through a pipeline.

There are two existing ways to compute neighbors: 
   - Through intersections of bounding box. This can always be done for a data set, as bounding boxes are always available. However, for unstructured grids, this may result in false positives, as bounding boxes may intersect without the actual domain intersecting.
   - Through the RankArray. This is more precise, however requires the RankArray to have been computed for vertices.

Both functions are defined in MPIUtils.h, and named `preconditionNeighborUsing[RankArray/BoundingBox]`. These functions already existed and have simply been renamed and slightly modified to fit the type of `neighborRanks_`.

During MPIPreconditioning, when the RankArray is not defined, Bounding boxes are used to compute the neighbors. Once RankArray is indeed computed, neighbors are computed again using RankArray.

If at the beginning of the MPI preconditioning, neighbors are already computed, no computation is done again.

Filters that use neighbors have been modified to not recompute each time: ArrayPreconditioning, ScalarFieldSmoother, GhostCellPreconditioning.

Any feedback is welcome,


julien-tierny:
thanks, ok, let's go (the CI was successful prior to the sync)

PR #861, opened 2022-11-17
Examples: Remove the Threshold compatibility layer

pierre-guillou:
This PR removes the ParaView 5.9 vs. 5.10 Threshold compatibility layer in the TTK examples. The latest is preferred over the oldest.

This also allows us to remove the now unused Python `packaging` dependency in the CI pipelines. It was used in the examples to compare the VTK versions.

Enjoy,
Pierre


julien-tierny:
wonderful, thanks!

PR #863, opened 2022-11-21
[MPI] Boundary preconditioning

eve-le-guillou:
Hi all,

Up until now, the MPI preconditioning would automatically compute 2 layers of ghost cells. This was due to the fact that numerous filters required not only information on neighbor simplices but also whether those simplices were on the global boundary. 
The global boundary is the real boundary of the data set if it had been executed sequentially. A local boundary is a boundary created by the distribution of the data set over multiple computing nodes. 

Having two layers of ghosts allowed each process to know whether the first layer of ghosts are on the boundary or not. However, this is memory consuming, it may increase communication costs when the exchange of ghost value is needed and it created problems on other filters that are being implemented using MPI. Therefore, a new solution has been implemented in this Pull Request.
Now, there is only one layer of ghost instead of two, and MPI support has been added to the boundary preconditioning to ensure that boundary preconditioning only refers to the global boundary.

For Explicit Triangulations, in sequential, the boundary preconditioning is stored in a boolean vector, with each element indicating whether the associated simplex is on the boundary or not. The MPI implementation builds on top of that: the boundary preconditioning is computed as it would be in sequential and then the values of ghost simplices for the boundary vector are exchanged between processes. The value of the owner of the simplex is the correct value of the boundary preconditioning.  This is done for vertices, edges and triangles.
The exchange of ghosts requires a RankArray for the corresponding simplices. For vertices and cells, those arrays are already automatically computed. For edges and triangles, those arrays can be computed using `preconditionEdgeRankArray` and `preconditionTriangleRankArray`. An edge or a triangle has the same RankArray value than the simplex of highest dimension of its star with the lowest global id.

For Implicit Triangulations, the computation of boundary in sequential is done on the fly. To compute the global boundary using MPI, we add a preconditioning step: the computation of `isOnGlobalBoundary_`. This is an array of size 6 of booleans. Each element stores the information of whether the face of the data set associated is on the global boundary or not. The array is organized as follows: LEFT, RIGHT, TOP, BOTTOM, FRONT, BACK. A face is considered on the global boundary if the distance between the global boundary and the local boundary is smaller than the spacing of one cell for that direction. 
When evaluating if a vertex is on the boundary or not, if the vertex is on the local boundary and is a ghost, the appropriate elements of `isOnGlobalBoundary_` are evaluated. For edges and triangle, an element is considered on the  global boundary using MPI if each of its vertices are on the global boundary.

Global Edge Ids and Global Triangle Ids have been changed of SimplexId to LongSimplexId for coherency (Global Vertex Ids and Global Cell Ids are LongSimplexId).

### DIFF
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -2564,6 +2564,15 @@ namespace ttk {
       return this->cellRankArray_;
     }
 
+    inline void setLocalBound(double bound[6]) {
+      this->localBound_[0] = bound[0];
+      this->localBound_[1] = bound[1];
+      this->localBound_[2] = bound[2];
+      this->localBound_[3] = bound[3];
+      this->localBound_[4] = bound[4];
+      this->localBound_[5] = bound[5];
+    };
+
     // public precondition method (used in Triangulation/ttkAlgorithm)
     virtual int preconditionDistributedVertices() {
       return 0;
@@ -3651,6 +3660,9 @@ namespace ttk {
     std::unordered_map<SimplexId, SimplexId> edgeGidToLid_{};
     std::vector<SimplexId> triangleLidToGid_{};
     std::unordered_map<SimplexId, SimplexId> triangleGidToLid_{};
+
+    double localBound_[6];
+    double globalBound_[6];
 
     bool hasPreconditionedDistributedCells_{false};
     bool hasPreconditionedDistributedEdges_{false};
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -2548,11 +2548,11 @@ namespace ttk {
       return this->vertGid_;
     }
 
-    inline const ttk::LongSimplexId *getEdgesGlobalIds() const {
+    inline const ttk::SimplexId *getEdgesGlobalIds() const {
       return this->edgeLidToGid_.data();
     }
 
-    inline const ttk::LongSimplexId *getTrianglesGlobalIds() const {
+    inline const ttk::SimplexId *getTrianglesGlobalIds() const {
       return this->triangleLidToGid_.data();
     }
 
@@ -2572,13 +2572,8 @@ namespace ttk {
       return this->cellRankArray_;
     }
 
-    inline void setLocalBound(double bound[6]) {
-      this->localBounds_[0] = bound[0];
-      this->localBounds_[1] = bound[1];
-      this->localBounds_[2] = bound[2];
-      this->localBounds_[3] = bound[3];
-      this->localBounds_[4] = bound[4];
-      this->localBounds_[5] = bound[5];
+    inline void setLocalBound(std::array<double, 6> &bound) {
+      this->localBounds_ = bound;
     };
 
     /// Pre-process the global boundaries when using MPI. Local bounds should
@@ -2808,38 +2803,6 @@ namespace ttk {
       return &(this->vertexGidToLid_);
     }
 
-    virtual inline const std::unordered_map<SimplexId, SimplexId> &
-      getEdgeGlobalIdMap() const {
-#ifndef TTK_ENABLE_KAMIKAZE
-      if(this->getDimensionality() != 1 && this->getDimensionality() != 2
-         && this->getDimensionality() != 3) {
-        this->printErr("Only 1D, 2D and 3D datasets are supported");
-      }
-      if(!this->hasPreconditionedDistributedEdges_) {
-        this->printErr("VertexGlobalMap query without pre-process!");
-        this->printErr(
-          "Please call preconditionDistributedEdges() in a pre-process.");
-      }
-#endif // TTK_ENABLE_KAMIKAZE
-      return this->edgeGidToLid_;
-    }
-
-    virtual inline const std::unordered_map<SimplexId, SimplexId> &
-      getTriangleGlobalIdMap() const {
-#ifndef TTK_ENABLE_KAMIKAZE
-      if(this->getDimensionality() != 1 && this->getDimensionality() != 2
-         && this->getDimensionality() != 3) {
-        this->printErr("Only 1D, 2D and 3D datasets are supported");
-      }
-      if(!this->hasPreconditionedDistributedEdges_) {
-        this->printErr("VertexGlobalMap query without pre-process!");
-        this->printErr(
-          "Please call preconditionDistributedTriangles() in a pre-process.");
-      }
-#endif // TTK_ENABLE_KAMIKAZE
-      return this->triangleGidToLid_;
-    }
-
     virtual inline std::vector<int> *getNeighborRanksWriteMode() {
       return &(this->neighborRanks_);
     }
@@ -2885,7 +2848,7 @@ namespace ttk {
       return it->second;
     }
 
-    inline LongSimplexId getEdgeGlobalIdInternal(const SimplexId leid) const {
+    inline SimplexId getEdgeGlobalIdInternal(const SimplexId leid) const {
       return this->edgeLidToGid_[leid];
     }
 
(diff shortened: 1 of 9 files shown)


core/base/abstractTriangulation/AbstractTriangulation.h
```diff
@@ -2548,6 +2548,14 @@ namespace ttk {
...
     }
 
+    inline const ttk::LongSimplexId *getEdgesGlobalIds() const {
+      return this->edgeLidToGid_.data();
+    }
+
+    inline const ttk::LongSimplexId *getTrianglesGlobalIds() const {
+      return this->triangleLidToGid_.data();
+    }
+
```
pierre-guillou: This seems to be incompatible with what I'm working on in the `ImplicitTriangulation`, meaning the on-the-fly computation of the edges and triangles global (resp. local) identifiers from their local (resp. global) identifiers. My changes would make these vectors useless (except for the `ExplicitTriangulation`).

eve-le-guillou: I use these functions to execute the function `ttk::exchangeGhostCells`, in the `ExplicitTriangulation` case, to exchange values of boundary preconditioning. For now, it is never used in the `ImplicitTriangulation` case. I also compute on the fly for the implicit case.
`ttk::exchangeGhostCells` may be subject to change in the future so it doesn't require an array of global ids to work. This would be compatible with your approach. That change has not been made yet.

julien-tierny: it sounds from this discussion that #866 should be merged prior to this PR.
do you both agree?

eve-le-guillou: I looked at #866, and it appears that this PR only implements the changes for cells, and not all simplices. Before #866, `exchangeGhostCells` could be used with any type of simplex or cell. Indeed, it would be good to merge it before this one, however the case of intermediary simplices (edges and triangles) would still be unchanged, therefore I would either need to change exchangeGhostCells again or keep things as they are, with a RankArray vector, for those simplices.

core/base/abstractTriangulation/AbstractTriangulation.h
```diff
@@ -2564,11 +2572,51 @@ namespace ttk {
...
     }
 
+    inline void setLocalBound(double bound[6]) {
+      this->localBounds_[0] = bound[0];
+      this->localBounds_[1] = bound[1];
+      this->localBounds_[2] = bound[2];
+      this->localBounds_[3] = bound[3];
+      this->localBounds_[4] = bound[4];
+      this->localBounds_[5] = bound[5];
+    };
```
pierre-guillou: I would suggest to use `std::array<double, 6>` instead of C arrays: `std::array`s implement a copy assignment operator, are zero-initialized and don't suffer from array-to-pointer decay.
Besides, they implement a lot of the `std::vector` API (`size()`, iterators) and can be converted to pointers with `data()`.

core/base/abstractTriangulation/AbstractTriangulation.h
```diff
@@ -3647,15 +3730,19 @@ namespace ttk {
     // (neighboring) ranks (per MPI rank)
     std::vector<std::vector<SimplexId>> remoteGhostCells_{};
 
-    std::vector<SimplexId> edgeLidToGid_{};
+    std::vector<ttk::LongSimplexId> edgeLidToGid_{};
     std::unordered_map<SimplexId, SimplexId> edgeGidToLid_{};
-    std::vector<SimplexId> triangleLidToGid_{};
+    std::vector<ttk::LongSimplexId> triangleLidToGid_{};
```
pierre-guillou: You should be able to keep the `SimplexId` type here if you (re-)template `exchangeGhostCells`.

core/base/abstractTriangulation/AbstractTriangulation.h
```diff
@@ -3647,15 +3730,19 @@ namespace ttk {
...
     std::unordered_map<SimplexId, SimplexId> triangleGidToLid_{};
 
+    double localBounds_[6]{0, 0, 0, 0, 0, 0};
+    double globalBounds_[6]{0, 0, 0, 0, 0, 0};
```
pierre-guillou: Use `std::array` here.

core/base/abstractTriangulation/AbstractTriangulation.h
```diff
@@ -2760,6 +2808,38 @@ namespace ttk {
...
+      }
+      if(!this->hasPreconditionedDistributedEdges_) {
+        this->printErr("VertexGlobalMap query without pre-process!");
+        this->printErr(
+          "Please call preconditionDistributedTriangles() in a pre-process.");
+      }
+#endif // TTK_ENABLE_KAMIKAZE
+      return this->triangleGidToLid_;
+    }
+
```
pierre-guillou: These methods provide an external access to the corresponding class members, which might be overkill if only used inside `ExplicitTriangulation`.

core/base/implicitTriangulation/ImplicitTriangulation.h
```diff
@@ -275,6 +279,8 @@ namespace ttk {
     // the cellGid_ array only applies on cubic cells, not on
     // simplicial ones...
     std::vector<SimplexId> cellLidToGid_{};
+    bool isOnGlobalBoundary_[6];
```
pierre-guillou: Use `std::array`

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -818,6 +818,9 @@ void ttkAlgorithm::MPITriangulationPreconditioning(
     triangulation->setCellRankArray(
       ttkUtils::GetPointer<int>(cd->GetArray("RankArray")));
   }
+  double bounds[6];
```
pierre-guillou: Use `std::array`.

pierre-guillou (commented):
I have listed some minor improvements that should reduce a bit the diff.

core/base/abstractTriangulation/AbstractTriangulation.h
eve-le-guillou: Is it a good idea to have an id type different for different simplices? That would require to cast all remaining ids when the domain changes when only simplices of smaller dimensions are kept. For example, when executing IntegraLines, the cells become the edges. `setGlobalIds` takes LongSimplexId types, so to add the array, it would needed to convert it to LongSimplexId. 
Is the cost of doing that smaller than the cost of storing the ids of intermediary simplices as LongSimplexId?

core/base/abstractTriangulation/AbstractTriangulation.h
eve-le-guillou: I will move them to `ExplicitTriangulation`.

pierre-guillou: If you don't need these methods outside the `ExplicitTriangulation`, I think there is no need to expose the class members this way. I would suggest to remove these methods.

core/base/abstractTriangulation/AbstractTriangulation.h
pierre-guillou: Considering that the `IntegralLines` output should be a small part of the input dataset, I think that the memory gains of compressing the intermediate identifiers of the input dataset (and keeping the consistency with the rest of the Triangulation API) should outweigh the conversion cost.

eve-le-guillou: Understood, thank you for the feedback

julien-tierny:
thanks @pierre-guillou for this careful review.
@eve-le-guillou could you ping me when you have addressed pierre's suggestions?
thanks!

eve-le-guillou:
@julien-tierny I have implemented Pierre's suggestions!

pierre-guillou:
Thanks ! It looks better now !

julien-tierny:
alright, looks good, thanks.
I'll merge that in.

@pierre-guillou could you take it over from here as discussed to integrate:
- your implicit local2global mechanism for simplices of intermediate dimensions
- your improved implicit boundary condition (based on the meta grid)

PR #865, opened 2022-11-21
GitHub Actions: Enable -Werror when building TTK

pierre-guillou:
This PR enables the `-Werror` compiler flag when building TTK in the `test_build` workflow.
All warnings are fixed:
- on Ubuntu 18.04, variable renaming fixed -Wshadow,
- on Ubuntu 22.04, tautological comparisons  were removed,
- on macOS, unused variables (written but never read) were removed,
- also on macOS, the `std::random_shuffle` method was replaced by `std::shuffle` in `ContourTreeAlignment` (a updated seed is needed in the corresponding state file & Python script),
- on Windows, a missing attribute has been added to the `WRLExporter` module,
- also on Windows, the `_CRT_SECURE_NO_WARNINGS` definition is used to silence warnings on purportedly more secure alternatives to standard POSIX functions (`fopen`).

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -98,6 +98,8 @@ jobs:
           -DTTK_IMPLICIT_PRECONDITIONS_THRESHOLD=64*64*64 \
           -GNinja \
           $GITHUB_WORKSPACE
+      env:
+        CXXFLAGS: -Werror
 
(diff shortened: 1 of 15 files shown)


julien-tierny:
could you sync this PR please?

julien-tierny:
I'll over-ride the CI this time as it was green earlier.

PR #866, opened 2022-11-22
[MPI] Get GhostCells from Triangulation

m-s-will:
Hi all,

similarly to #860 , the `AbstractTriangulation` already has the attributes `ghostCellPerOwner_` and `remoteGhostCells_`, which are however not used after getting computed.

This PR adapts `getGhostCellScalars(..)` to get the ghostCell information from the triangulation and to use the triangulation-methods to e.g. get the local id for global ids. For this to work correctly, the triangulation needs to precondition the cells beforehand, this is done in `MPITriangulationPreconditioning()`where similar methods already reside.

However, not all classes actually have a triangulation when they try to exchange ghost cell information (for example, when creating an order array for an existing scalar array), therefore the old functionality still exists in the `getGhostCellScalarsWithoutTriangulation(..)` method. 

Any feedback is welcome.

### DIFF
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -2818,10 +2818,23 @@ namespace ttk {
       return &(this->remoteGhostCells_);
     }
 
+    virtual inline const std::vector<std::vector<SimplexId>> *
+      getGhostVerticesPerOwner() const {
+      return &(this->ghostVerticesPerOwner_);
+    }
+
+    virtual inline const std::vector<std::vector<SimplexId>> *
+      getRemoteGhostVertices() const {
+      return &(this->remoteGhostVertices_);
+    }
+
     virtual inline void setHasPreconditionedDistributedVertices(bool flag) {
       this->hasPreconditionedDistributedVertices_ = flag;
     }
 
+    virtual inline bool hasPreconditionedDistributedVertices() const {
+      return this->hasPreconditionedDistributedVertices_;
+    }
     virtual inline bool hasPreconditionedDistributedCells() const {
       return this->hasPreconditionedDistributedCells_;
     }
@@ -3665,6 +3678,11 @@ namespace ttk {
     // global ids of local (owned) cells that are ghost cells of other
     // (neighboring) ranks (per MPI rank)
     std::vector<std::vector<SimplexId>> remoteGhostCells_{};
+    // global ids of (local) ghost vertices per each MPI (neighboring) rank
+    std::vector<std::vector<SimplexId>> ghostVerticesPerOwner_{};
+    // global ids of local (owned) vertices that are ghost cells of other
+    // (neighboring) ranks (per MPI rank)
+    std::vector<std::vector<SimplexId>> remoteGhostVertices_{};
 
     std::array<double, 6> localBounds_;
     std::array<double, 6> globalBounds_;
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -2787,7 +2787,7 @@ namespace ttk {
       this->hasPreconditionedDistributedVertices_ = flag;
     }
 
-    virtual inline bool getHasPreconditionedDistributedCells() const {
+    virtual inline bool hasPreconditionedDistributedCells() const {
       return this->hasPreconditionedDistributedCells_;
     }
 
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -2607,7 +2607,6 @@ namespace ttk {
       return 0;
     }
 
-
     // global <-> local id mappings
 
     virtual inline SimplexId getVertexGlobalId(const SimplexId lvid) const {
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -2775,7 +2775,7 @@ namespace ttk {
 
     virtual inline const std::vector<std::vector<SimplexId>> *
       getGhostCellsPerOwner() const {
-      return &(this->ghostCellPerOwner_);
+      return &(this->ghostCellsPerOwner_);
     }
 
     virtual inline const std::vector<std::vector<SimplexId>> *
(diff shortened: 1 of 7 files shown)


core/base/abstractTriangulation/AbstractTriangulation.h
```diff
@@ -2768,10 +2773,24 @@ namespace ttk {
...
+    virtual inline const std::vector<std::vector<SimplexId>> *
+      getRemoteGhostCells() const {
+      return &(this->remoteGhostCells_);
+    }
+
     virtual inline void setHasPreconditionedDistributedVertices(bool flag) {
       this->hasPreconditionedDistributedVertices_ = flag;
     }
 
+    virtual inline bool getHasPreconditionedDistributedCells() const {
```
julien-tierny: can you please simply rename this function into `hasPreconditionedDistributedCells()` (to follow the rest of the preconditioning API)? thanks!

core/base/triangulation/Triangulation.h
```diff
@@ -1424,6 +1424,10 @@ namespace ttk {
       abstractTriangulation_->setHasPreconditionedDistributedVertices(flag);
     }
 
+    inline bool getHasPreconditionedDistributedCells() const override {
```
julien-tierny: same remark as above.

julien-tierny (changes requested):
ok @m-s-will, that looks good to me! thanks a lot for the quick PR!
could you simply take care of my two little remarks?

also @eve-le-guillou, could you confirm this is in line with what you had in mind?
thanks!

eve-le-guillou:
Thanks for this PR!

I am not quite sure what it is you are exchanging with `exchangeGhostCells`: is it cells or vertices? At line 126 of MPIUtils.h, `ghostCellPerOwner` is used to predict how much data will be received, however it is the function `getVertexLocalId` that is used to turn a global id into a local id.

Furthermore, it is my bad, I wasn't clear enough, but I wanted to use exchangeGhostCells to exchange any type of simplex (vertex, edge, triangle or cell). It appears that the changes here only correspond to cells or vertices, am i correct?
`getGhostCellScalarsWithoutTriangulation` is compatible with any type of simplex, but requires the computation of `RankArray`, which is not ideal in the Implicit case.

julien-tierny (approved):

m-s-will:
> I am not quite sure what it is you are exchanging with `exchangeGhostCells`: is it cells or vertices? At line 126 of MPIUtils.h, `ghostCellPerOwner` is used to predict how much data will be received, however it is the function `getVertexLocalId` that is used to turn a global id into a local id.

Thanks, for the catch, it seems like I misunderstood ghostCellPerOwner and accidentally tested it with the wrong version, in the current form it does not work! 
Generally, most of the functions should be moved to the triangulation, so we would have global<->local functions for every type of simplex and also information for each type where the ghost cells / vertices reside. Currently only the ghostCell owner information is computed (in preconditionDistributedCells), while only the vertices has the needed exchange methods (getVertexGlobalId, getVertexLocalId). Therefore large changes to the triangulations (especially the implicit, where I'm currently not 100% sure how it's handled / planned) would be needed and #863 should probably be merged beforehand.
 
> Furthermore, it is my bad, I wasn't clear enough, but I wanted to use exchangeGhostCells to exchange any type of simplex (vertex, edge, triangle or cell). It appears that the changes here only correspond to cells or vertices, am i correct? `getGhostCellScalarsWithoutTriangulation` is compatible with any type of simplex, but requires the computation of `RankArray`, which is not ideal in the Implicit case.

`getGhostCellScalarsWithoutTriangulation` should remain the generic case where you can just supply the correct arrays and it should be agnostic of the simplex type and even be doable without a triangulation. However I don't currently see how that could be done without a RankArray (which does not necessarily need a triangulation).

eve-le-guillou:
> Generally, most of the functions should be moved to the triangulation, so we would have global<->local functions for every type of simplex and also information for each type where the ghost cells / vertices reside. Currently only the ghostCell owner information is computed (in preconditionDistributedCells), while only the vertices has the needed exchange methods (getVertexGlobalId, getVertexLocalId). 

Currently, global<->local functions exist for vertices, edges, triangles and cells.
Pierre Guillou is working on on-the-fly computation for those functions for the Implicit case, but the functions already exist and should not change in terms of input/output. Can you confirm @pierre-guillou ?

Therefore you should be able to build on top of what already exists, regardless of the changes that will come later for global<->local functions.

> `getGhostCellScalarsWithoutTriangulation` should remain the generic case where you can just supply the correct arrays and it should be agnostic of the simplex type and even be doable without a triangulation. However I don't currently see how that could be done without a RankArray (which does not necessarily need a triangulation).

I had in mind to replace the RankArray by the equivalent of `ghostCellPerOwner` for other simplices (which is not implemented yet) and have the preconditioning done somewhere else. But writing functions `exchangeGhostVertices`, `exchangeGhostEdges` and `exchangeGhostTriangles` that use a triangulation and are similar to `exchangeGhostCells` would work just as well.

pierre-guillou:
My current work will keep the `getEdgeGlobal`/`getEdgeLocal`/`getTriangleGlobal`/`getTriangleLocal` API in the `abstractTriangulation` module but the class members `edgeLidToGid_`/`edgeGidToLid_`/`triangleLidToGid_`/`triangleGidToLid_` (among others) will be moved to the `ExplicitTriangulation` since they are not used anymore by the `ImplicitTriangulation`.

julien-tierny:
hi @m-s-will 
the mpi triangulation mpi changed a bit with the PR #871 
could you please update this PR accordingly (i.e. by solving conflicts)?

please don't hesitate to let us know if you have any questions!

thanks a lot!

m-s-will:
@eve-le-guillou I added `exchangeGhostVertices` and replaced the necessary calls with it. While doing this I found weird behaviour of the boundary preconditioning, especially `preconditionBoundaryVerticesInternal`. I thought it may have been an error in my branch, but I checked out ttk/dev and it is in there too. 

Attached are two images of two runs for ScalarFieldCriticalPoints. Disregarding the char typing, if a vertex is **not** on the boundary (so IsOnBoundary != 1), it seems like the values change randomly. Similarly, the critical type is inconsistent for these vertices. 

Do you have an idea where the problems could be?

Bildschirm­foto 2022-11-29 um 10 33 44
Bildschirm­foto 2022-11-29 um 10 34 20

eve-le-guillou:
Hi Michael,

I had noticed this bug as well, and it only appears in the Spreadsheet view. If you color your figure on the left according to CriticalType or onBoudary, you'll notice the coloring is correct.
Also, if you save the data to file, the values won't be random, they'll also be correct.
I think the Char typing and the random values are connected, but I haven't really looked into the problem in details.

m-s-will:
Ah ok, then it's probably just a display bug because of the signed char / char / unsigned char differences. 
I changed the function for the more generic data exchange from `exchangeGhostCellsWithoutTriangulation` to `exchangeGhostDataWithoutTriangulation`, because it is exactly this that we want, more generic data exchange and nothing specialized to cells, vertices, etc. 

The specialized functions for cells and vertices with triangulation are done and for edges and triangles in the works.

julien-tierny:
@eve-le-guillou @pierre-guillou 
is the PR in line with what you had in mind or do we need more discussions?

eve-le-guillou:
Yes, this is what I had in mind, it looks good to me!

pierre-guillou:
Looks good to me too

julien-tierny:
rock'n'roll
let's go then :)

PR #867, opened 2022-11-22
Reduce build times of mergeTree filters

pierre-guillou:
This PR removes the template instantiations in the `mergeTree` modules, thus reducing their compilation time. The concerned modules are:

- `ttkMergeTreeClustering` (from 130s to 25s),
- `ttkMergeTreeDistanceMatrix` (from 70s to 15s),
- `ttkMergeTreePrincipalGeodesics` (from 130s to 25s),
- `ttkMergeTreePrincipalGeodesicsDecoding` (from 110s to 25s),
- `ttkMergeTreeTemporalDecoding` (from 110s to 25s),
- `ttkMergeTreeTemporalEncoding` (from 120s to 25s) and
- `ttkPlanarGraphLayout` (from 65s to 30s).

All of these modules converts merge trees in VTK format (as outputted by `ttkFTMTree`) to the internal merge tree representation used in the base layer and defined in the `ftmTree` module.

Using `vtkTemplateMacro` in these filters was not necessary since:

1. the data are copied from the VTK arrays to the internal structure (c.f. `ttk::ftm::makeTree` in ttkFTMTreeUtils.h) using `vtkDataArray::GetTuple` that casts the data in `double` (no `ttkUtils::GetPointer` or `ttkUtils::GetVoidPointer` used),
https://github.com/topology-tool-kit/ttk/blob/2a1f2b72cfdbfbd802e05f5c7db33f71685041b9/core/vtk/ttkFTMTree/ttkFTMTreeUtils.h#L38
2. the `vtkDataArray` on which the templating is performed, the `Scalar` Point Data array on `ttkFTMTree`'s `Ouput Nodes` is defined to be a `vtkFloatArray`. https://github.com/topology-tool-kit/ttk/blob/2a1f2b72cfdbfbd802e05f5c7db33f71685041b9/core/vtk/ttkFTMTree/ttkFTMStructures.h#L117

To keep the compatibility with the previous implementation, the `float` variant is explicitly chosen in the filters.

This is quite similar to what has been done to the `ttkPersistenceDiagramClustering` module several years ago.

Aggressive de-templating of base modules could lead to additional improvements in the build times but a lot more work is needed.

Enjoy,
Pierre

### DIFF
--- a/core/vtk/ttkMergeTreeClustering/ttkMergeTreeClustering.cpp
+++ b/core/vtk/ttkMergeTreeClustering/ttkMergeTreeClustering.cpp
@@ -4,7 +4,6 @@
 #include <MergeTreeUtils.h>
 #include <MergeTreeVisualization.h>
 #include <ttkFTMTreeUtils.h>
-#include <ttkMacros.h>
 #include <ttkMergeTreeClustering.h>
 #include <ttkMergeTreeVisualization.h>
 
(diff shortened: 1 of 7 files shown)


julien-tierny:
looks good, thanks!

PR #869, opened 2022-11-25
GitHub Actions: More aggressive deletion of Python 3.11 on macOS

pierre-guillou:
This PR deletes another directory in which an installation of Python 3.11 in the macOS VM can be found. This is to ensure that the correct Python (3.10) is selected by CMake to be linked by TTK. We already use Python 3.10 for the ParaView headless binaries and the `scikit-learn` package. Mixing Python versions can lead to errors in the `DimensionReduction` module.

Enjoy,
Pierre


julien-tierny:
alright, let's give it a shot.

PR #870, opened 2022-11-25
MorseSmaleComplex: Support non-manifold datasets

pierre-guillou:
This PR reworks how the ascending 1-separatrices are computed, in order to fix a segfault happening on non-manifold datasets.

Old assumptions in the code that counted 2 ascending separatrices per saddle are invalid for non-manifold datasets. For instance, in a 2D Rips Complex, an edge is often shared by more than two triangles. This PR relaxes these assumptions, the code doesn't expect an arbitrary number of ascending 1-separatrices per saddle any more.

Since this change may degrade the performance of the filter, I performed some code cleaning, especially around the `Separatrix` data-structure, which possessed unused or ill-used class members. `std::set`s were also replaced with `std::vector`s in `DiscreteGradient` for storing a de-duplicated list of saddles on walls.

Finally, the computation of `separatrixFunctionMin`, `separatrixFunctionMax` and `numberOfCriticalPointsOnBoundary` has been modified to take into account that there may be no saddles on the boundary on a wall.

Enjoy,
Pierre

### DIFF
--- a/core/base/discreteGradient/DiscreteGradient.h
+++ b/core/base/discreteGradient/DiscreteGradient.h
@@ -251,7 +251,8 @@ in the gradient.
                             VisitedMask &mask,
                             const triangulationType &triangulation,
                             std::vector<Cell> *const wall = nullptr,
-                            std::set<SimplexId> *const saddles = nullptr) const;
+                            std::vector<SimplexId> *const saddles
+                            = nullptr) const;
 
       /**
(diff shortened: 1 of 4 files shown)


julien-tierny:
wonderful, looks great. thanks!

julien-tierny:
(let's wait for the CI to complete though)

PR #871, opened 2022-11-25
ImplicitTriangulation: Compute on-the-fly local/global identifiers for edges and triangles

pierre-guillou:
This PR reworks the `getEdgeGlobalId`/`getTriangleGlobalId`/`getEdgeLocalId`/`getTriangleLocalId` methods for the `ImplicitTriangulation` by computing the values on-the-fly instead of storing them into the `edgeGidToLid_`/`edgeLidToGid_`/`triangleGidToLid_`/`triangleLidToGid_` class members during precondition.

The main idea is to use a "meta" grid whose dimensions are those of the input dataset (before being distributed by ParaView) and global to local vertices mappings (which are stored anyway) to query, for instance, the identifier of the edge in the "meta" grid that matches given global vertices identifiers. The "meta" grid dimensions are provided by the `Identifiers` module, called from `ttkAlgorithm`.

A consequence of these changes is that preconditions introduced in #773 and #814 are now useless for `ImplicitTriangulation`. They have been removed from this module and the corresponding class members inside `AbstractTriangulation` (including `edgeGidToLid_`, `edgeLidToGid_`, `triangleGidToLid_` and `triangleLidToGid_` ) have been moved in `ExplicitTriangulation`.

Some cleaning has been performed following the last MPI PRs, in particular around preferring references to pointers, function overloading with and without `const` qualifier and using `inline` to factor implementations.

A further PR will extend this scheme of using a "meta" grid to detect boundary vertices, edges, triangles & cells.

@eve-le-guillou Is this good for you?

Enjoy,
Pierre


core/base/identifiers/Identifiers.h
```diff
@@ -887,6 +892,12 @@ namespace ttk {
       int height
         = static_cast<int>((globalBounds[3] - globalBounds[2]) / spacing_[1])
           + 1;
+      int depth
+        = static_cast<int>((globalBounds[5] - globalBounds[4]) / spacing_[2])
+          + 1;
+
+      // store global grid dimensions
+      this->globalDims_ = {width, height, depth};
```
eve-le-guillou: This line is not necessarily executed, in the case the global ids were created previously, so this line may never be reached.

I also just noticed that the global boundary is computed twice: in `GenerateGlobalIds` and in `preconditionGlobalBoundary`, that computes the global boundary in case boundary preconditioning needs to be done. From what I understood, with your PR, the computation of the global boundary should always be done, to create the meta grid.

If it is moved before GenerateGlobalIds, the computation is not done twice. The computation of `globalDims_` could be done at the same time.
What do you think?

eve-le-guillou:
This all looks great! 
I just had one remark, but everything else seems good for me.
Great job!

core/base/identifiers/Identifiers.h
pierre-guillou: If I understand correctly, you suggest to move the computation of the global dimensions out of `Identifiers`, which may not be called if global identifiers are present on cells & vertices?

pierre-guillou: I copied the `Identifiers` code inside `ImplicitTriangulation` and now `ImplicitTriangulation::createMetaGrid` only depends on the distributed dataset bounds.

julien-tierny:
alright, looks good. thanks a lot!
looking forward to the next PR handling boundary conditions with the meta grid :)

PR #872, opened 2022-11-30
[MPI] Add MPI support to the discrete gradient filter

eve-le-guillou:
Hi all,

This PR adds MPI support to the discrete gradient filter.

The gradient is computed by iterating on each vertex of a data set. When the vertex is a ghost vertex, all simplices of its lower star are set to ghosts by setting their gradient to the value GHOST_GRADIENT (-2).

The data array CellId and ttkVertexScalarField are field using global ids when using MPI.

Changes to `ImplicitTriangulation` were also made to fix a bug introduced in #871, that would result in all global ids of intermediary simplices being equal to 1 when executed on one process.

Any feedback is welcome.

### DIFF
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -2858,6 +2858,30 @@ namespace ttk {
       return this->hasPreconditionedDistributedCells_;
     }
 
+    ttk::SimplexId inline getDistributedGlobalCellId(ttk::SimplexId localCellId,
+                                                     int cellDim) const {
+      if(ttk::hasInitializedMPI()) {
+        switch(cellDim) {
+          case 0:
+            return this->getVertexGlobalIdInternal(localCellId);
+          case 1:
+            return this->getEdgeGlobalIdInternal(localCellId);
+          case 2:
+            if(getDimensionality() == 2) {
+              return this->getCellGlobalIdInternal(localCellId);
+            } else {
+              return this->getTriangleGlobalIdInternal(localCellId);
+            }
+          case 3: {
+            return this->getCellGlobalIdInternal(localCellId);
+          }
+        }
+        return -1;
+      } else {
+        return localCellId;
+      }
+    }
+
   protected:
     inline SimplexId getVertexGlobalIdInternal(const SimplexId lvid) const {
       return this->vertGid_[lvid];
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -2550,11 +2550,11 @@ namespace ttk {
 
     // RankArray on points & cells
 
-    inline void setVertRankArray(const int *const rankArray) {
-      this->vertRankArray_ = rankArray;
+    inline void setVertexRankArray(const int *const rankArray) {
+      this->vertexRankArray_ = rankArray;
     }
-    inline const int *getVertRankArray() const {
-      return this->vertRankArray_;
+    inline const int *getVertexRankArray() const {
+      return this->vertexRankArray_;
     }
 
     inline void setCellRankArray(const int *const rankArray) {
@@ -3692,7 +3692,7 @@ namespace ttk {
     // "GlobalPointIds" from "Generate Global Ids"
     const LongSimplexId *vertGid_{};
     // PointData "RankArray" from "TTKGhostCellPreconditioning"
-    const int *vertRankArray_{};
+    const int *vertexRankArray_{};
     // CellData "RankArray" from "TTKGhostCellPreconditioning"
     // (warning: for Implicit/Periodic triangulations, concerns
     // "squares"/"cubes" and not "triangles"/"tetrahedron")
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -2807,8 +2807,35 @@ namespace ttk {
       return this->neighborRanks_;
     }
 
+    virtual inline const std::vector<std::vector<SimplexId>> *
+      getGhostCellsPerOwner() const {
+      return &(this->ghostCellsPerOwner_);
+    }
+
+    virtual inline const std::vector<std::vector<SimplexId>> *
+      getRemoteGhostCells() const {
+      return &(this->remoteGhostCells_);
+    }
+
+    virtual inline const std::vector<std::vector<SimplexId>> *
+      getGhostVerticesPerOwner() const {
+      return &(this->ghostVerticesPerOwner_);
+    }
+
+    virtual inline const std::vector<std::vector<SimplexId>> *
+      getRemoteGhostVertices() const {
+      return &(this->remoteGhostVertices_);
+    }
+
     virtual inline void setHasPreconditionedDistributedVertices(bool flag) {
       this->hasPreconditionedDistributedVertices_ = flag;
+    }
+
+    virtual inline bool hasPreconditionedDistributedVertices() const {
+      return this->hasPreconditionedDistributedVertices_;
+    }
+    virtual inline bool hasPreconditionedDistributedCells() const {
+      return this->hasPreconditionedDistributedCells_;
     }
 
   protected:
@@ -3646,10 +3673,15 @@ namespace ttk {
     // list of neighboring ranks (sharing ghost cells to current rank)
     std::vector<int> neighborRanks_{};
     // global ids of (local) ghost cells per each MPI (neighboring) rank
-    std::vector<std::vector<SimplexId>> ghostCellPerOwner_{};
+    std::vector<std::vector<SimplexId>> ghostCellsPerOwner_{};
     // global ids of local (owned) cells that are ghost cells of other
     // (neighboring) ranks (per MPI rank)
     std::vector<std::vector<SimplexId>> remoteGhostCells_{};
+    // global ids of (local) ghost vertices per each MPI (neighboring) rank
+    std::vector<std::vector<SimplexId>> ghostVerticesPerOwner_{};
+    // global ids of local (owned) vertices that are ghost cells of other
+    // (neighboring) ranks (per MPI rank)
+    std::vector<std::vector<SimplexId>> remoteGhostVertices_{};
 
     std::array<double, 6> localBounds_;
     std::array<double, 6> globalBounds_;
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -2858,28 +2858,35 @@ namespace ttk {
       return this->hasPreconditionedDistributedCells_;
     }
 
-    ttk::SimplexId inline getDistributedGlobalCellId(ttk::SimplexId localCellId,
-                                                     int cellDim) const {
+    inline int getDistributedGlobalCellId(const SimplexId &localCellId,
+                                          const int &cellDim,
+                                          SimplexId &globalCellId) const {
       if(ttk::hasInitializedMPI()) {
         switch(cellDim) {
           case 0:
-            return this->getVertexGlobalIdInternal(localCellId);
+            globalCellId = this->getVertexGlobalIdInternal(localCellId);
+            break;
           case 1:
-            return this->getEdgeGlobalIdInternal(localCellId);
+            globalCellId = this->getEdgeGlobalIdInternal(localCellId);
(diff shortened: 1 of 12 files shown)


core/base/discreteGradient/DiscreteGradient.cpp
```diff
@@ -164,21 +164,27 @@ bool DiscreteGradient::isCellCritical(const int cellDim,
   }
 
   if(cellDim == 0) {
-    return ((*gradient_)[0][cellId] == -1);
+    return ((*gradient_)[0][cellId] != GHOST_GRADIENT)
```
julien-tierny, line 167: if I understand correctly your code, `GHOST_GRADIENT` is equal to -2 and NULL_GRADIENT to -1.
so if the gradient for that cell is equal to -1, it cannot be equal to -2.
so that original line does not even need to be modified, right?

julien-tierny: (or you can simply change the -1 into NULL_GRADIENT)

core/base/discreteGradient/DiscreteGradient.cpp
```diff
@@ -164,21 +164,27 @@ bool DiscreteGradient::isCellCritical(const int cellDim,
...
+           && ((*gradient_)[0][cellId] == NULL_GRADIENT);
```
julien-tierny: if I understand right:
if your gradient is equal to NULL_GRADIENT (second condition), then the first condition has to be false.
so in short, the first condition is not needed.
in the original line, I would simply change -1 to NULL_GRADIENT.

(careful, the discrete gradient is very very efficient and adding non-needed tests could slow it down)

core/base/discreteGradient/DiscreteGradient.cpp
```diff
@@ -164,21 +164,27 @@ bool DiscreteGradient::isCellCritical(const int cellDim,
...
   }
 
   if(cellDim == 1) {
-    return ((*gradient_)[1][cellId] == -1
-            && (dimensionality_ == 1 || (*gradient_)[2][cellId] == -1));
+    return ((*gradient_)[1][cellId] != GHOST_GRADIENT)
+           && ((*gradient_)[1][cellId] == NULL_GRADIENT
```
julien-tierny: same remark as before

core/base/discreteGradient/DiscreteGradient.cpp
```diff
@@ -164,21 +164,27 @@ bool DiscreteGradient::isCellCritical(const int cellDim,
...
+               && (dimensionality_ == 1
+                   || (*gradient_)[2][cellId] == NULL_GRADIENT));
   }
 
   if(cellDim == 2) {
-    return ((*gradient_)[3][cellId] == -1
-            && (dimensionality_ == 2 || (*gradient_)[4][cellId] == -1));
+    return ((*gradient_)[3][cellId] != GHOST_GRADIENT)
+           && ((*gradient_)[3][cellId] == NULL_GRADIENT
```
julien-tierny: same remark as before

core/base/discreteGradient/DiscreteGradient.cpp
```diff
@@ -164,21 +164,27 @@ bool DiscreteGradient::isCellCritical(const int cellDim,
...
+               && (dimensionality_ == 2
+                   || (*gradient_)[4][cellId] == NULL_GRADIENT));
   }
 
   if(cellDim == 3) {
-    return ((*gradient_)[5][cellId] == -1);
+    return ((*gradient_)[5][cellId] != GHOST_GRADIENT)
```
julien-tierny: same remark as before

core/base/discreteGradient/DiscreteGradient_Template.h
```diff
@@ -1515,6 +1537,33 @@ ttk::SimplexId DiscreteGradient::getCellLowerVertex(
   return vertexId;
 }
 
+#ifdef TTK_ENABLE_MPI
+template <typename triangulationType>
+int DiscreteGradient::getDistributedGlobalCellId(
```
julien-tierny: it looks to me that this function could be moved in the abstract triangulation.
I believe this feature may be used by all other filters, which at some point store on the output some identifiers (for instance scalarfield critical points, integral lines and all).

if you agree, thanks for moving this to the abstract triangulation and modify (in this PR directly) the other filters which could use this function.

core/base/discreteGradient/DiscreteGradient_Template.h
```diff
@@ -121,7 +121,12 @@ int DiscreteGradient::setCriticalPoints(
 
     triangulation.getCellIncenter(cell.id_, cell.dim_, points[i].data());
     cellDimensions[i] = cellDim;
+#ifdef TTK_ENABLE_MPI
```
julien-tierny, line 130: do we really need to protect this function by the TTK_ENABLE_MPI?

i.e. the function getDistributedGlobalCellId() will return localIds in case we have a single process, then it will provide the right result, even with only one process. so here, typically, just sticking to the use of getDistributedGlobalCellId() will make the code simpler to read and maintain, at the cost of an extra function call. 
I would argue we can probably afford that extra function call to make the code clearer.

core/base/discreteGradient/DiscreteGradient_Template.h
```diff
@@ -463,7 +468,9 @@ int DiscreteGradient::processLowerStars(
 
   // store lower star structure
   lowerStarType Lx;
-
+#ifdef TTK_ENABLE_MPI
+  const int *vertRankArray = triangulation.getVertRankArray();
```
julien-tierny: could you change (everywhere it is appropriate) the name of that function into getVertexRankArray() (to make it even more explicit)

core/vtk/ttkDiscreteGradient/ttkDiscreteGradient.cpp
```diff
@@ -101,7 +101,16 @@ int ttkDiscreteGradient::fillCriticalPoints(
     cellIds->SetTuple1(i, critPoints_cellIds[i]);
     cellScalars->SetTuple1(i, scalars[critPoints_PLVertexIdentifiers[i]]);
     isOnBoundary->SetTuple1(i, critPoints_isOnBoundary[i]);
+#ifdef TTK_ENABLE_MPI
+    if(ttk::hasInitializedMPI()) {
```
julien-tierny, line 105: I would argue that this test should happen in the base code.
i.e. when you fill the output vector critPoints_PLVertexIdentifiers, there you should check if you should use local or global ids.

the vtk layer can be completely independent from that.

julien-tierny (changes requested):
hi @eve-le-guillou 

thanks for the PR.
could you please consider my remarks.

@pierre-guillou 
could you also have a quick look?
thanks!

core/base/discreteGradient/DiscreteGradient_Template.h
pierre-guillou: Well, most of the MPI functions are under `TTK_ENABLE_MPI` anyway, so I would keep it for consistency. It also helps to highlight the MPI-dependent code parts.

core/base/discreteGradient/DiscreteGradient.cpp
```diff
@@ -233,3 +239,24 @@ int DiscreteGradient::setManifoldSize(
...
+
+#ifdef TTK_ENABLE_MPI
+void DiscreteGradient::setCellToGhost(const int cellDim,
+                                      const SimplexId cellId) {
+  if(cellDim == 0) {
+    (*gradient_)[0][cellId] = GHOST_GRADIENT;
+  }
+
+  if(cellDim == 1) {
+    (*gradient_)[1][cellId] = GHOST_GRADIENT;
```
pierre-guillou, line 247: If I remember correctly, you might also want to put GHOST_GRADIENT in (*gradient_)[2] for edges and (*gradient_)[4] for triangles (each pair of gradient_ vectors holds mapping between facets and co-facets of increasing dimension).

core/base/discreteGradient/DiscreteGradient_Template.h
julien-tierny: alright, let's keep it this way then. thanks for your input.

core/vtk/ttkDiscreteGradient/ttkDiscreteGradient.cpp
pierre-guillou: The corresponding code in the base layer is the `DiscreteGradient::setCriticalPoints` method (the first one inside DiscreteGradient_Template.h).

core/base/discreteGradient/DiscreteGradient_Template.h
eve-le-guillou: DiscreteGradient is the only filter that requires such a function for now. I could move it to AbstractTriangulation, but I don't see any other filter that needs it. In all other filter, the dimension of the simplex is always the same, therefore it is possible to directly call the right function.

core/vtk/ttkDiscreteGradient/ttkDiscreteGradient.cpp
eve-le-guillou: `critPoints_PLVertexIdentifiers` in local ids is needed to add elements to the `cellScalars` array. Moving the test to the base layer means adding a new array as argument to setCriticalPoints, as `critPoints_PLVertexIdentifiers` is still needed here. I thought it would be simpler to do this in the vtk layer, as adding arguments only useful in MPI is verbose and makes the code harder to read.
Do you still want me to do this change?

core/base/discreteGradient/DiscreteGradient.cpp
eve-le-guillou: Indeed, that is true, I will correct that

core/base/discreteGradient/DiscreteGradient.cpp
eve-le-guillou: Done!

core/base/discreteGradient/DiscreteGradient_Template.h
julien-tierny: I'd still suggest to move this function to the abstract triangulation class. 
There is nothing in it that is specific to the discrete gradient.
thanks!

julien-tierny (changes requested):
Thanks @eve-le-guillou 
Could you address that last comment?

core/base/discreteGradient/DiscreteGradient_Template.h
eve-le-guillou: Understood, changes done!

core/vtk/ttkDiscreteGradient/ttkDiscreteGradient.cpp
julien-tierny: understood, let's keep it this way then.
thanks for the clarification.

core/base/abstractTriangulation/AbstractTriangulation.h
```diff
@@ -2858,6 +2858,30 @@ namespace ttk {
       return this->hasPreconditionedDistributedCells_;
     }
 
+    ttk::SimplexId inline getDistributedGlobalCellId(ttk::SimplexId localCellId,
```
julien-tierny: I'd suggest the following change here (to be consistent with the other functions of the triangulations):
`inline int getDistributedGlobalCellId(const SimplexId &localCellId, const int &cellDim, SimplexId &globalCellId) const{`

julien-tierny: then the funcion returns 0 upon success and negative values otherwise

julien-tierny: the general idea here is to have input arguments as const and output arguments as non const

core/base/discreteGradient/DiscreteGradient_Template.h
```diff
@@ -1575,8 +1609,15 @@ int DiscreteGradient::setGradientGlyphs(
         points_pairOrigins[2 * nProcessedGlyphs] = 0;
         points_pairOrigins[2 * nProcessedGlyphs + 1] = 1;
         cells_pairTypes[nProcessedGlyphs] = i;
+#ifdef TTK_ENABLE_MPI
+        triangulation.getDistributedGlobalCellId(j, i, globalId);
```
pierre-guillou, line 1613: You can declare `globalId` inside the loop to make it thread-private (and factor/avoid `#ifdefs TTK_ENABLE_MPI`)

julien-tierny:
ok, that one looks ready to go. thanks!

PR #868, opened 2022-11-23
GitHub Actions: Execute test_build workflow on pushes on "dev" branch

pierre-guillou:
This PR extends the `test_build` workflow to be executed also on pushes on the "dev" branch, but only on the `topology-tool-kit/ttk` repository (forked repositories that have enabled actions will skip the execution of the workflow).

Now, there should be 3 ways to execute this workflow:
* by pushing a tag matching the specifications,
* on a pull request on the "dev" branch and
* on a push on the "dev" branch iff the repository is `topology-tool-kit/ttk`.

This should also enable the badge in the README.

Enjoy,
Pierre

PR #873, opened 2022-11-30
GitHub Actions: Generate ccache archives after each run

pierre-guillou:
This PR builds on #868 by generating ccache archives for all tested platforms and storing them each time the `test_build` workflow is executed on pushes in the `topology-tool-kit` repository (usually corresponding to pull request merges).

This ensures the ccache archives are up-to-date and should reduce the time spend building TTK in the CI.

Enjoy,
Pierre


julien-tierny:
looks great, thanks!

PR #874, opened 2022-11-30
FTMTree: Sort produced nodes and arcs

pierre-guillou:
This PR extends the `FTMTree_MT::normalizeIds` method by sorting produced nodes and arcs according to the vertex order field.

This should make the FTMTree filter output deterministic when used in parallel (previously, the nodes and arcs order was depending on the parallel scheduling).

Enjoy,
Pierre


pierre-guillou:
There seems to be some issues left, especially with the `mergeTreeFeatureTracking.pvsm` state file...

julien-tierny:
thanks for letting me know when this one is ready to be reviewed.

pierre-guillou:
After fixing some inconsistencies left after the sorting operations and disabling the sorting for Contour Trees, this should be ready to go.

julien-tierny:
alright, let's go for it.

PR #876, opened 2022-11-30
ImplicitTriangulation: Use meta grid to compute isOnGlobalBoundary

pierre-guillou:
This PR extends #863 by using the meta grid introduced in #871 to avoid preconditioning in the Implicit Triangulation. To do so, it introduces 3 new methods: `isVertexOnGlobalBoundaryInternal`, `isEdgeOnGlobalBoundaryInternal` and `isTriangleOnGlobalBoundaryInternal` that are resp. called by `isVertexOnBondary`, `isEdgeOnBoundary` and `isTriangleOnBoundary`. These new methods forward the call to the meta grid. Consequently, the corresponding preconditioning  step is rendered useless and has been removed. The new methods are not meant to be called by non-triangulation modules, so they have been made protected.

Additionally, the `AbstractTriangulation` methods that return the global/local simplices identifier were modified to work on one process only.

Finally, `#if` conditionals were replaced by `#ifdef`s which better fit their use case.

Enjoy,
Pierre


julien-tierny:
nice, looks good. thanks!

PR #877, opened 2022-12-05
GitHub Actions: Reduce workflows concurrency

pierre-guillou:
This PR uses [concurrency groups](https://docs.github.com/en/actions/using-jobs/using-concurrency) to reduce the number of workflows running in parallel. Pushing a new commit to an open PR (or to the `dev` branch of `topology-tool-kit/ttk` repository) will cancel already running workflows for this PR rather than letting older workflows run in parallel with newer workflows.

This should reduce the computational impact on GitHub's infrastructure and avoid concurrency issues with ccache when several pushes in the `dev` branch happen in a short time span.

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -17,6 +17,10 @@ on:
       - 'mpi'
 
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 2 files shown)


julien-tierny:
looks great, thanks!

PR #878, opened 2022-12-05
ImplicitTriangulation: Support on-the-fly global/local vertex identifiers

pierre-guillou:
This PR implements the `getVertexGlobalIdInternal` and `getVertexLocalIdInternal` methods for `ImplicitTriangulation`, using an on-the-fly computation scheme based on the meta grid introduced in #871: the offset of the local grid inside the meta grid is computed from the bounding box and stored inside `ImplicitTriangulation`.

The `getGridDimensions` method has been reworked on the occasion.

Uses of `vertGid_` and `vertexGidToLid_` were removed from `ImplicitTriangulation`. We still need `preconditionDistributedVertices` as it fills other class members (`ghostVerticesPerOwner_`, `remoteGhostVertices_`) that may be used somewhere else (?)

I have not touched `ttkAlgorithm` to remove the call to `Identifiers` as it also computes the vertex rank array (and other cell-related arrays).

Enjoy,
Pierre

### DIFF
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -2881,11 +2881,13 @@ namespace ttk {
     }
 
   protected:
-    inline SimplexId getVertexGlobalIdInternal(const SimplexId lvid) const {
+    virtual inline SimplexId
+      getVertexGlobalIdInternal(const SimplexId lvid) const {
       return this->vertGid_[lvid];
     }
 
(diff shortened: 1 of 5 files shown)


julien-tierny:
looks good, thanks!

@pierre-guillou could you PR the same mechanism for the cells (i.e. tetrahedra in 3D, triangles in 2D).

@eve-le-guillou could you test this PR and make the necessary arrangements in ttkAlgorithm (if needed)?

PR #880, opened 2022-12-08
TopologicalSimplification: Optimize if only one constraint type

pierre-guillou:
This PR checks the number of extrema in the TopologicalSimplification & LTS constraints. If there is only one type of constraint, the algorithms are only executed once (instead of twice).

Enjoy,
Pierre


julien-tierny:
awesome, thanks!

PR #881, opened 2022-12-08
MorseSmaleComplex: Use relationship between pairs to skip saddle connectors returning

pierre-guillou:
This PR fixes issues with the algorithm introduced in #857 to return saddle connectors. First, invalid saddle connectors are detected and skipped to avoid infinite loops or segfauls. Secondly, `DiscreteMorseSandwich` now optionally exports the parent relationship between 2-saddles (a 2-saddle points to its children, meaning the 2-saddles whose boundary has been merged into its one). This relationship is then used to avoid computing the 2-separatrix wall if the connector of one or more children has been skipped, leading to improved performance.

Enjoy,
Pierre


julien-tierny:
looks great, looking forward to trying it out :)

PR #883, opened 2022-12-13
Replace manual memory management

pierre-guillou:
This PR reworks the following modules: `webSocketIO`, `ftrGraph`, `boundingVolumeHierarchy`, `contourAroundPoint`, `contourTree`, `contourTreeAlignment`, `contourForests` and `ftmTree` to replace manual memory management (`new` & `delete`) with modern C++ alternatives (allocate on stack, `std::shared_ptr`, `std::unique_ptr`, storage with a large enough scope).

This might remove some memory leaks but also might improve the performance when indirections are removed. At least, the `ftmTree` performance seems to be the same.

Enjoy,
Pierre

### DIFF
--- a/core/base/boundingVolumeHierarchy/BoundingVolumeHierarchy.h
+++ b/core/base/boundingVolumeHierarchy/BoundingVolumeHierarchy.h
@@ -12,8 +12,10 @@
 
 #include "Ray.h"
 #include <Geometry.h>
+
 #include <algorithm>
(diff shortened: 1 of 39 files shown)


julien-tierny:
awesome, thanks!

julien-tierny:
quick question: is there any way to make the github actions check the presence of `new` and `delete` in PRs?

PR #884, opened 2022-12-14
MorseSmaleComplex: Improve segmentation performance

pierre-guillou:
This PR focuses on improving the performance of the Morse-Smale complex ascending and descending segmentations.

The main idea is to follow integral lines for every vertex or tetrahedron towards the corresponding local minimum or maximum. The previous approach used a BFS from the extrema to extract their basins.

For instance, here are some speedups on `ctBones.vti` (8 threads):
* critical points extraction: from 1.55s to 0.21s
* ascending manifold: from 2.68s to 1.13s
* descending manifold: from 5.33s to 0.18s

While the new approach is way more efficient for the descending manifold, the speedup observed on the ascending manifold is less consequent. This is in part due to the number of tetrahedron being around 6 times the number of vertices on a regular grid and the need to handle non-manifold datasets.

I tried another faster propagation scheme for the ascending segmentation, with no real success. It consisted in following integral lines on vertices by their greatest neighbor. This was two or three times slower than the computation of the descending manifold (mostly due to the extraction of the greatest neighbor) but failed to stop at some separatrices.

Another improvement is the rewrite of the `DiscreteGradient::getCriticalPoints` method using an optimized version that has been moved from `DiscreteMorseSandwich`. Previously, the critical cells were stored in a flat manner as `Cell`s. Now, only the simplex ids are stored, in separate vectors per dimension. This remove the need to extract back cells of specific dimensions inside `MorseSmaleComplex`. Additionally, the extraction process has been parallelized using a map-reduce scheme (at it was in `DiscreteMorseSandwich`).

Enjoy,
Pierre

### DIFF
--- a/core/base/discreteGradient/DiscreteGradient.cpp
+++ b/core/base/discreteGradient/DiscreteGradient.cpp
@@ -191,21 +191,19 @@ bool DiscreteGradient::isCellCritical(const Cell &cell) const {
 }
 
 int DiscreteGradient::setManifoldSize(
-  const size_t nCritPoints,
-  const std::vector<size_t> &nCriticalPointsByDim,
+  const std::array<std::vector<SimplexId>, 4> &criticalCellsByDim,
   const SimplexId *const ascendingManifold,
   const SimplexId *const descendingManifold,
   std::vector<SimplexId> &manifoldSize) const {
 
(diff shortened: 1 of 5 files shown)


julien-tierny:
awesome, thanks a lot!

PR #885, opened 2022-12-15
[MergeTree*] avoid casting unsigned int max value to int

MatPont:
This PR removes all casting of unsigned int max value to int in the merge tree related filters.

Instead of initializing some unsigned int to -1 it uses the `std` function `std::numeric_limits<ftm::idNode>::max()` and to test equality it now avoids casting to int and then comparing to -1 by comparing directly to this value.


julien-tierny:
alright, looks good. thanks!

PR #886, opened 2022-12-15
[MergeTreeDistance] rename misleading distanceSquared parameter name and fix display

MatPont:
This PR renames the misleading `distanceSquared` parameter of merge tree related filters to `distanceSquaredRoot` since setting this parameter to true will returns the squared root of the sum the costs (the real distance) and not the distance squared.

It also fixes the display of the distance computation to take into account this parameter.


julien-tierny:
looks good, thanks!

PR #887, opened 2022-12-15
[MergeTreeClustering] fix barycenter parameters xml logical expression

MatPont:
This PR fixes the logical expressions in the xml file of the `MergeTreeClustering` filter of some parameters related to the barycenter/clustering computation.

The problem was that the logical gates was not in the right order allowing for some parameters to not show after some very specific manipulation.

Example: select the custom backend, check the keep subtrees option, then select the Wasserstein backend, check the compute barycenter option, then the number of clusters option will not show.


julien-tierny:
ok, thanks!

PR #888, opened 2022-12-16
MPI: On-the-fly computation of rank for vertices & cells of Implicit triangulations

pierre-guillou:
This PR adds the support of on-the-fly computation of vertex and cell rank for Implicit triangulations, thus alleviating the need of storing them in memory. The buffers that previously held them were moved from `AbstractTriangulation` to `ExplicitTriangulation`. In `ttkAlgorithm`, "RankArray" PointData or CellData arrays are not needed anymore and code has been reworked to get rid of them. In particular, `MPIUtils` functions that required a pointer to a "RankArray" buffer were altered to take a Callable instead.

The on-the-fly computation of ranks relies on pre-computing bounding boxes of non-ghost vertices & cells. These bounding boxes are exchanged & stored as integer coordinates in the meta-grid. Then, a query quickly compares the global coordinates to the bounding boxes of the neighbors, to get the rank it fits in. The result should be identical to what was provided through the "RankArray" data arrays.

Other now less useful `AbstractTriangulation` members were moved to `ExplicitTriangulation`, such as the pointers to vertex and cell global identifiers or the map global to local vertex identifier.

Enjoy,
Pierre

### DIFF
--- a/core/base/abstractTriangulation/AbstractTriangulation.h
+++ b/core/base/abstractTriangulation/AbstractTriangulation.h
@@ -2816,24 +2816,24 @@ namespace ttk {
       return this->neighborRanks_;
     }
 
-    virtual inline const std::vector<std::vector<SimplexId>> *
+    virtual inline const std::vector<std::vector<SimplexId>> &
       getGhostCellsPerOwner() const {
-      return &(this->ghostCellsPerOwner_);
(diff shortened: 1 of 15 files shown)


julien-tierny:
looks good, thanks!

PR #889, opened 2022-12-16
DimensionReduction: Clean API

pierre-guillou:
This PR reworks the API of the `DimensionReduction` base module. It introduces a new enum class to improve the method selection, compact the setters and removes duplicate class members between the base and VTK layers.

This should make the base module easier to be called from another base module.

Enjoy,
Pierre


julien-tierny:
looks good, thanks!

PR #890, opened 2022-12-16
[FTMTree] fix scalars and params copy

MatPont:
This PR fixes the use of shared pointers of #883 in `ftmTree`. 

Indeed, during a copy of a `MergeTree` the now-used pointers were shared between the old and the new tree resulting in problem if scalars of the first tree were modified (the scalars of the copied tree would also be modified).

Especially with the `mergeTreeTemplateToDouble` function that changes the type of the scalars, giving weird results when trying to access the scalars of the template tree since they were casted in double.

This PR fixes that by creating a new pointer and copying the information.

Done with the agreement of @pierre-guillou.


julien-tierny:
alright, let's go for it.

PR #891, opened 2022-12-19
ParaView 5.11 patches

pierre-guillou:
This PR adds TTK patches for ParaView 5.11. It fixes the TTK Examples UI (there's apparently a new ParaView example, messing with the columns of the previous patch). A CMake component has been renamed to fix an error in the NSIS Windows packager. A mpi4py file has been regenerated using latest Cython to provide compatibility between ParaView, MPI and Python 3.11. The headless packages now use the default Python (the first one found by CMake, which should currently be 3.11) on macOS. A basic gitignore is also part of the patch.

The macOS package format (productbuild) has been replaced with a binary archive (`.tar.gz`) for practical concerns (the productbuild artifacts were not really functional). Binary archives were already in use for headless packages on macOS so this change reduces the diff between the two workflows.

The old workflow file `main.yml` has been removed as it has been deprecated by the `package.yml` workflow file.

Further patches should follow to adapt TTK's GitHub Actions workflows to the new generated binaries.

Enjoy,
Pierre


julien-tierny:
looks great, thanks!

PR #892, opened 2022-12-22
[MT-PGA] fix doxygen parameters and reference

MatPont:
This small PR fixes the doxygen parameters of the MT-PGA module and update the reference.


julien-tierny:
looks good, thanks!

PR #894, opened 2023-01-02
Forbid use of Variable-Length Arrays

pierre-guillou:
This PR adds the `-Wvla` compiler flag to detect the use of Variable-Length Arrays inside TTK. VLAs are a controversial C feature introduced in C99 but made optional in C11. They don't seem to be part of the C++ standard (but are supported anyway). A `std::vector` (although heap-allocated) is a natural replacement.

This PR removes the last VLA occurrence inside `MPIUtils.h`. With the compiler flag (and the CI), no new VLA should be introduced anymore in TTK.

Enjoy,
Pierre

### DIFF
--- a/core/base/common/MPIUtils.h
+++ b/core/base/common/MPIUtils.h
@@ -386,9 +386,10 @@ namespace ttk {
     }
     std::vector<int> sendVector(neighborSet.begin(), neighborSet.end());
     int localSize = neighborSet.size();
-    int sizes[ttk::MPIsize_];
-    int displacements[ttk::MPIsize_];
-    MPI_Gather(&localSize, 1, MPI_INT, sizes, 1, MPI_INT, 0, communicator);
+    std::vector<int> sizes(ttk::MPIsize_);
+    std::vector<int> displacements(ttk::MPIsize_);
+    MPI_Gather(
+      &localSize, 1, MPI_INT, sizes.data(), 1, MPI_INT, 0, communicator);
     int totalSize = 0;
     if(ttk::MPIrank_ == 0) {
       for(int i = 0; i < ttk::MPIsize_; i++) {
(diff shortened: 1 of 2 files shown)


core/base/common/MPIUtils.h
```diff
@@ -386,9 +386,10 @@ namespace ttk {
     }
     std::vector<int> sendVector(neighborSet.begin(), neighborSet.end());
     int localSize = neighborSet.size();
-    int sizes[ttk::MPIsize_];
-    int displacements[ttk::MPIsize_];
-    MPI_Gather(&localSize, 1, MPI_INT, sizes, 1, MPI_INT, 0, communicator);
+    std::vector<int> sizes(ttk::MPIsize_);
```
julien-tierny, line 389: why not using `std::arrray` here?

core/base/common/MPIUtils.h
```diff
@@ -386,9 +386,10 @@ namespace ttk {
...
+    std::vector<int> displacements(ttk::MPIsize_);
```
julien-tierny, line 390: same as above

julien-tierny (changes requested):
why not using `std::array` instead of `std::vector`?

core/base/common/MPIUtils.h
pierre-guillou: `ttk::MPIsize_` is not a compile-time constant, we can't use `std::array` here.

julien-tierny: Is that a requirement?
See `core/base/progressiveTopology/ProgressiveTopology.h:121`

julien-tierny: It is a requirement indeed.

julien-tierny (approved):

julien-tierny:
ok, looks good, thanks!

core/base/common/MPIUtils.h
pierre-guillou: `ttk::MPIsize` is a run-time variable set via `mpirun`. When enabling `-Wvla`, a warning appears on these lines. Replacing the C-array with a `std::array` (whose size should be a compile-time constant) returns a compiler error.

PR #895, opened 2023-01-04
GitHub Actions: Add support for ParaView 5.11

pierre-guillou:
This PR updates the TTK workflow files to use the v5.11.0 packages from ttk-paraview. The `ccache` workflow has been retired, since most of its functionality has been duplicated in the `test_build` workflow (and is a PITA to maintain).

The `package` workflow no resamples `ctBones.vti` when testing the `ttk-data` Python scripts. On macOS, binary archives are preferred to productbuild artifacts for packaging. Binary archives are installed under `/usr/local`, productbuild under `/Applications` and need more work with environment variables to make it work.

Note that some `ttk-data`  state files must be updated to fix some deleted ParaView filters. This will be done in another PR.

Enjoy,
Pierre


julien-tierny:
awesome, thanks!

PR #896, opened 2023-01-09
GaussianPointCloud: Avoid parallel random

pierre-guillou:
This PR removes the OpenMP parallel generation of Gaussian point cloud, as it interferes with the determinism introduced in https://github.com/topology-tool-kit/ttk/commit/a6c7c013aafbeb7772a47e4e1f3c95257e7644d0. 

Indeed, the previous pseudo-random determinist scheme was dependent on the number of threads on which the filter is run (one pseudo-random generator with a distinct seed per thread). To keep the determinism, the OpenMP support has been removed (the filter is fast anyway).

Enjoy,
Pierre


julien-tierny:
thanks!

PR #898, opened 2023-01-20
GitHub Actions: Update screenshot filter

pierre-guillou:
This PR updates the screenshot filter for the Ubuntu jobs of the `test_build` workflow to be in sync with https://github.com/topology-tool-kit/ttk-data/pull/147.

Also, the run-time dependencies of the Ubuntu ttk-data jobs have been reworked, the generated .deb artifacts are deleted after use and the ccache URLs now points to the `topology-tool-kit` repository (instead of the one that executes the workflow).

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -229,17 +229,10 @@ jobs:
         if ! python3 -u validate.py; then
 
           if [[ "$VERS" == *"22.04"* ]]; then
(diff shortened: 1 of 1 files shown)


julien-tierny:
can you please sync with the dev branch? (for the updated CI)

julien-tierny:
thanks a lot!

PR #899, opened 2023-01-20
PersistenceDiagram: Force PersistenceSimplexPairs on non-manifold datasets

pierre-guillou:
This PR extends the (`Explicit`)`Triangulation` by adding a new precondition to detect non-manifold datasets. It checks the number of cells in the stars of (d-1)-simplices.

This new precondition is then used in `PersistenceDiagram` to select the `PersistentSimplexPairs` (Zomorodian) backend by default, instead of `DiscreteMorseSandwich`, on non-manifold datasets. As a matter of fact, the saddle-max pairs generated by `DiscreteMorseSandwich` on non-manifold datasets are invalid (`PersistentGenerators` is not affected) and the simpler `PersistentSimplexPairs` is correct and may be faster.

The `PersistentSimplexPairs` backend has been cleaned a little: the output pairs format now matches `DiscreteMorseSandwich` and the progression is displayed in the terminal.

Enjoy,
Pierre

### DIFF
--- a/core/base/abstractTriangulation/AbstractTriangulation.cpp
+++ b/core/base/abstractTriangulation/AbstractTriangulation.cpp
@@ -33,6 +33,7 @@ int AbstractTriangulation::clear() {
   hasPreconditionedVertexNeighbors_ = false;
   hasPreconditionedVertexStars_ = false;
   hasPreconditionedVertexTriangles_ = false;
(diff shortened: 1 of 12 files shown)


julien-tierny:
can you please sync with the dev branch? (for the updated CI)

julien-tierny:
thanks!

PR #902, opened 2023-01-23
MPIUtils: Pass global identifiers array via a Callable

pierre-guillou:
This PR rewrites the MPIUtils functions to pass a Callable (a lambda) instead of a pointer for the global identifiers parameter. This modification ensures that the triangulation methods can be called to provide the global identifiers instead of relying on a VTK data array.

Enjoy,
Pierre


julien-tierny:
ok, looks good, thanks!

PR #903, opened 2023-01-24
ExplicitTriangulation: Move preconditionEdges,TrianglesRankArray

pierre-guillou:
This PR moves the call to precondition edges & triangles rank arrays from the preconditionBoundary* methods to the preconditionDistributed* ones.

This ensures that edges & triangles rank arrays are always computed in an MPI context.

Enjoy,
Pierre


julien-tierny:
thanks!

PR #904, opened 2023-01-26
[MPI] Add MPI support for IntegralLines

eve-le-guillou:
Hi all,

This PR adds MPI support for the IntegralLines filter as well as fixes a few bugs/problems for the MPI API.
Here is a list of all changes outside of IntegralLines:
- Addition of a setter for RankArray for vertices and cells, in case an explicit grid is created from an implicit grid in a filter (such as IntegralLines). This prevents TTK from recomputing the RankArray when it is already present in the Pipeline.
- A function `getVertexLocalIdIfExists` has been added, which returns the local id of a vertex if the vertex belongs to the current process, and -1 otherwise.
- The class ArrayLinkedList has been added in the core/base/common directory. This class is used in IntegralLines to store the data of integral lines while being thread safe.
- In the ExplicitTriangulation class, after the computation of the RankArray for vertices, neighbors are recomputed using the RankArray (as it is a more precise way than using the bounding box as firstly done). RankArray is only computed here if it is not empty.
- In the ImplicitTriangulation, when computing offsets and dimension of the grid, `std::round` was added before the `static_cast`, as the former always rounds down, causing false results for a handful of vertices.
- In ScalarFieldCriticalPoints, modifications were made to store upper and lower components when computing the critical type of a vertex. This will be used in IntegralLines.
- In ttkAlgorithm, the use of RankArray was added back to the method `checkGlobalIdValidty`, as in our development model, ghosts should be used as little as possible as they can't be trusted. The same was done in `GenerateGlobalIds` for the explicit grid. The call to `checkGlobalIdValidty` is now only done when the triangulation is an explicit triangulation.

In addition to the previous changes, extensive changes to integralLines were made. The filter now support MPI and OpenMP. The seeds given as inputs can either be already distributed across processes or only present on one process. The filter will know to distribute the seeds according to the domain of each process.
Regarding the core algorithm, several changes were made, regardless of MPI:
- The algorithm now stops only at maximum or minimum. When reaching a saddle point, the algorithm will fork and create as many new integral lines as there are upper (or lower) components on the current vertex.
- The next vertex is no longer chosen based on the highest value of gradient, as for grids of high resolution, the `float` precision was no longer enough to produce a correct result. The next vertex is now chosen using the highest value of the order array.
- For each integral line, the data stored is the following: the identifier of the seed, the identifier of the first vertex chosen after the last fork, the distance from the seed, the trajectory in local identifiers and the number of each vertex (called localVertexIdentifier).

The execution of the overall algorithm with MPI and OpenMP is the following:
- The seeds are distributed.
- OpenMP tasks containing the computation of several integral lines are created. When an integral line arrives at a ghost vertex, a message is created and added to the queue, to later send to the next process.
- When all computations of integral lines are finished, processes exchange the number of integral lines that have been fully finished on their domain (all lines except for those that arrived at a ghost).
- If there are still integral lines to compute, processes will exchange all their queued messages at once, and create new tasks to compute integral lines.
- The three previous points happen again and again until all integral lines are fully computed.
- Then new global identifiers are created for the output vertices and edges and the new output geometry is created.

Thanks for any feedback,
Eve


core/base/common/ArrayLinkedList.h
```diff
@@ -0,0 +1,37 @@
+/// \ingroup base
+/// \class ttk::ArrayLinkedList
```
julien-tierny, line 2: can you please add a short description of what the class is doing? thanks!

core/base/common/ArrayLinkedList.h
```diff
@@ -0,0 +1,37 @@
...
+  class ArrayLinkedList {
+  public:
+    std::list<std::array<datatype, size>> list;
+    int numberOfElement;
+    // In order to prevent false sharing when creating a
+    // std::vector of ArrayLinkedList objects (one element
+    // of the std::vector for each thread), it is necessary
+    // that one ArrayLinkedList object be bigger than one cache
+    // line. Here, we assume a cache line to be 64 bytes.
+    unsigned char padding[32];
```
julien-tierny: I guess it'd be preferrable to replace that with a modern variant:
`std::array<unsigned char, 32> padding_{};`

core/base/common/ArrayLinkedList.h
```diff
@@ -0,0 +1,37 @@
...
+#include <list>
+
+#pragma once
+
+namespace ttk {
+  template <typename datatype, int size>
+  class ArrayLinkedList {
+  public:
+    std::list<std::array<datatype, size>> list;
+    int numberOfElement;
```
julien-tierny: please use the prefix `_` for all member variables.
`list` -> `list_`
`numberOfElement` -> `numberOfElements_` (plural)
`padding` -> `padding_`

core/base/explicitTriangulation/ExplicitTriangulation.cpp
```diff
@@ -123,21 +123,29 @@ int ExplicitTriangulation::preconditionBoundaryEdgesInternal() {
 #ifdef TTK_ENABLE_MPI
 
 int ExplicitTriangulation::preconditionVertexRankArray() {
-  this->vertexRankArray_.resize(this->vertexNumber_);
-  if(ttk::isRunningWithMPI()) {
-    ttk::produceRankArray(this->vertexRankArray_, this->vertGid_,
-                          this->vertexGhost_, this->vertexNumber_,
-                          this->boundingBox_.data(), this->neighborRanks_);
+  if(vertexRankArray_.size() == 0) {
+    this->vertexRankArray_.resize(this->vertexNumber_);
```
julien-tierny: I believe you want to write `this->vertexRankArray_.resize(this->vertexNumber_, 0);`
(to init it with 0, useful if not running with MPI)

core/base/explicitTriangulation/ExplicitTriangulation.cpp
```diff
@@ -123,21 +123,29 @@ int ExplicitTriangulation::preconditionBoundaryEdgesInternal() {
...
 }
 
 int ExplicitTriangulation::preconditionCellRankArray() {
-  this->cellRankArray_.resize(this->cellNumber_);
-  if(ttk::isRunningWithMPI()) {
-    ttk::produceRankArray(this->cellRankArray_, this->cellGid_,
-                          this->cellGhost_, this->cellNumber_,
-                          this->boundingBox_.data(), this->neighborRanks_);
+  if(cellRankArray_.size() == 0) {
+    this->cellRankArray_.resize(this->cellNumber_);
```
julien-tierny: same as above `this->cellRankArray_.resize(this->cellNumber_, 0);`

core/base/scalarFieldCriticalPoints/ScalarFieldCriticalPoints.h
```diff
@@ -88,18 +88,27 @@ namespace ttk {
...
 
     template <class triangulationType = AbstractTriangulation>
-    int getNumberOfLowerUpperComponents(const SimplexId vertexId,
-                                        const SimplexId *const offsets,
-                                        const triangulationType *triangulation,
-                                        ttk::SimplexId &lowerComponentNumber,
-                                        ttk::SimplexId &upperComponentNumber,
-                                        bool &isLowerOnBoundary,
-                                        bool &isUpperOnBoundary) const;
+    int getNumberOfLowerUpperComponents(
```
julien-tierny: I'd suggest to re-model this function as follows:
- rename the function into `getLowerAndUpperComponents(`
- remove the parameters `ttk::SimplexId &lowerComponentNumber,` and `ttk::SimplexId &upperComponentNumber,` (in the using call, we'll simply call the size of the components, see below)
- force the usage of a reference of vectors for the output (without default value to `nullptr`): `std::vector<std::vector<ttk::SimplexId>> &upperComponents` and `std::vector<std::vector<ttk::SimplexId>> &lowerComponents`
- actually compute the lower and upper components even in case of extrema (this is easy, before lines 417).

core/base/scalarFieldCriticalPoints/ScalarFieldCriticalPoints.h
```diff
@@ -483,6 +494,42 @@ int ttk::ScalarFieldCriticalPoints::getNumberOfLowerUpperComponents(
...
   std::vector<UnionFind *>::iterator it;
+  // We retrieve the lower and upper components if we want them
+  if(upperComponents != nullptr && lowerComponents != nullptr) {
+    std::vector<UnionFind *> upperUnionFind;
+    upperComponents->push_back(
+      std::vector<ttk::SimplexId>(1, upperNeighbors[0]));
+    upperUnionFind.push_back(upperList[0]);
+    for(ttk::SimplexId i = 1; i < (SimplexId)upperNeighbors.size(); i++) {
+      it
+        = std::find(upperUnionFind.begin(), upperUnionFind.end(), upperList[i]);
```
julien-tierny: I believe this piece of code can be much simpler.
after line 494, you just need to iterate on `upperNeighbors` (each entry is a local vertex identifier). For the entry i (i.e. for the vertex `upperNeighbors[i]`), its corresponding union-find seed is given by `upperList[i]->find()`. at this point you may want to store the current vertex (`upperNeighbors[i]`) in a map of vectors (`std::unordered_map<std::vector<SimplexId>, SimplexId>`) which stores for each component id, its list of vertices.

julien-tierny: the suggested usage of a map will trade the quadratic loop (find after insertions) by something faster in practice.

core/base/scalarFieldCriticalPoints/ScalarFieldCriticalPoints.h
```diff
@@ -508,13 +555,16 @@ template <class triangulationType>
 char ttk::ScalarFieldCriticalPoints::getCriticalType(
   const SimplexId &vertexId,
   const SimplexId *const offsets,
-  const triangulationType *triangulation) const {
+  const triangulationType *triangulation,
+  std::vector<std::vector<ttk::SimplexId>> *upperComponents,
```
julien-tierny, line 530: here set a default value `upperComponents = nullptr` (same for `lowerComponents`)

julien-tierny: then, in the code (before the call to `getLowerAndUpperComponents`), instantiate a `std::vector<std::vector<ttk::SimplexId>> localUpperComponents;`
then:
```
if(upperComponents == nullptr){
  upperComponents= &localUpperComponents;
}
```

same for `lowerComponents`.

julien-tierny (changes requested):
hi @eve-le-guillou ,
thanks for the PR.
here's a first batch of comments.
thanks for pinging me on slack if you need to further discuss them.
cheers,

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -683,32 +708,41 @@ void ttkAlgorithm::MPIPipelinePreconditioning(
 
   bool pointValidity{false};
   bool cellValidity{false};
-  if(globalPointIds != nullptr) {
-    unsigned char *ghostPoints = ttkUtils::GetPointer<unsigned char>(
-      input->GetPointData()->GetArray("vtkGhostType"));
-    pointValidity
-      = checkGlobalIdValidity(globalPointIds, vertexNumber, ghostPoints);
+  if((triangulation != nullptr
+      && triangulation->getType() == ttk::Triangulation::Type::EXPLICIT)
```
julien-tierny: thanks for also considering the case `ttk::Triangulation::Type::Compact` everywhere there is a special treatment for the explicit version (the compact version is a variant of the explicit one).

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -683,32 +708,41 @@ void ttkAlgorithm::MPIPipelinePreconditioning(
...
-  }
   // If the global ids are not valid, they are computed again
   if(!pointValidity || !cellValidity) {
-    bool flag{false};
-    if(triangulation != nullptr
-       && triangulation->getType() == ttk::Triangulation::Type::EXPLICIT) {
-      flag = this->GenerateGlobalIds(
-        input, triangulation->getVertexGlobalIdMap(), neighborRanks);
+    if(triangulation != nullptr) {
+      if(triangulation->getType() == ttk::Triangulation::Type::EXPLICIT) {
```
julien-tierny: cf Compact comment above.

core/base/integralLines/IntegralLines.h
```diff
@@ -72,9 +173,157 @@ namespace ttk {
...
+     * @param trajectories linked list of identifiers of forks, one for each
+     * integral line
+     * @param localVertexIdentifiers linked list of identifiers, one for each
+     * vertex of an integral lines. It represents the number of the vertex in
+     * the integral line.
+     * @param triangulation Triangulation
+     * @return int
+     */
+    template <typename triangulationType>
+    int getGlobalIdentifiers(
```
julien-tierny, line 213: are all these parameters required to be non-const?
please declare in `const` all the parameters which are read-only (that helps the interpretation of the code).
thanks for double-checking all function prototypes in the base code.

core/base/integralLines/IntegralLines.h
```diff
@@ -72,9 +173,157 @@ namespace ttk {
...
+     * @brief Sorts ghosts and exchanges their global identifiers between
+     * processes.
+     *
+     * @param unmatchedGhosts Ghosts vertices and edges for which the global
+     * identifier is to be determined
+     * @param globalVertexId vector of global identifiers for vertices
+     * @param globalCellId vector of global identifiers for cells
+     * @return int 0 for success
+     */
+    int exchangeGhosts(std::vector<std::vector<ttk::intgl::GhostElementsToSort>>
```
julien-tierny: can't that function (or significant parts of it) go to `core/base/common/MPIUtils.h`?

julien-tierny (changes requested):
here's a mini-batch (more to come)

core/base/integralLines/IntegralLines.h
eve-le-guillou: Everything in this function is very specific to integral lines. It can be moved to MPIUtils.h but for now no other filter uses it. Only a filter that has an output that require to create new geometry with points and cells that can be located in the same coordinates while at the same time being able to sort the GhostElementToSort exactly the same way on two different processes. This feels so specific that I don't think any other filter fits these conditions. Please correct me if I'm wrong.

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
```diff
@@ -1,11 +1,16 @@
+#include "BaseClass.h"
```
julien-tierny: this line should not be needed.
please remove it.

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
```diff
@@ -1,11 +1,16 @@
...
+#include "DataTypes.h"
```
julien-tierny: same as above.

core/base/integralLines/IntegralLines.h
```diff
@@ -14,14 +16,89 @@
...
 #include <Triangulation.h>
-
 // std includes
+#include <algorithm>
+#include <iterator>
 #include <limits>
+#include <numeric>
 #include <unordered_set>
 
+#define TABULAR_SIZE 50
```
julien-tierny: can you please rename this constant with a more explicit name (with the name of the class where it is defined being apparent), for example:
```
INTEGRAL_LINE_TABULAR_SIZE
```

core/base/integralLines/IntegralLines.h
```diff
@@ -14,14 +16,89 @@
...
+#ifdef TTK_ENABLE_MPI
+#define IS_ELEMENT_TO_PROCESS 0
```
julien-tierny: same as above, please make the name more explicit with `INTEGRAL_LINE` in the name

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
```diff
@@ -40,25 +45,62 @@ int ttkIntegralLines::FillOutputPortInformation(int port,
...
-  ttk::Triangulation *triangulation,
-  std::vector<std::vector<ttk::SimplexId>> &trajectories,
+  triangulationType *triangulation,
+  std::vector<ttk::ArrayLinkedList<std::vector<ttk::SimplexId>, TABULAR_SIZE>>
+    &trajectories,
+  std::vector<ttk::ArrayLinkedList<ttk::SimplexId, TABULAR_SIZE>>
+    &forkIdentifiers,
+  std::vector<ttk::ArrayLinkedList<std::vector<double>, TABULAR_SIZE>>
+    &distancesFromSeed,
+  std::vector<ttk::ArrayLinkedList<ttk::SimplexId, TABULAR_SIZE>>
```
julien-tierny: I guess it would be more readable to define (in the base code) a single structure `integralLine` which would have several members, including:
- a `std::vector<ttk::SimplexId>` trajectory
- a `ttk::SimplexId` forkIdentifier
- a `std::vector<double>` distance from field
- a `ttk::SimplexId` seed identifier.

once this is done, you could manipulate objects of the type
`std::vector<ttk::ArrayLinkedList<integralLine, TABULAR_SIZE>`

that'd be much more readable than manipulating multiple vectors.

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
```diff
@@ -40,25 +45,62 @@ int ttkIntegralLines::FillOutputPortInformation(int port,
...
+#endif
   vtkUnstructuredGrid *output) {
-
   if(input == nullptr || output == nullptr
      || input->GetPointData() == nullptr) {
     this->printErr("Null pointers in getTrajectories parameters");
     return 0;
   }
-
+  ttk::SimplexId threadNumber = trajectories.size();
```
julien-tierny: why not using the purposely-introduced variable `threadNumber_`?

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
```diff
@@ -40,25 +45,62 @@ int ttkIntegralLines::FillOutputPortInformation(int port,
   return 1;
 }
 
+template <typename triangulationType>
 int ttkIntegralLines::getTrajectories(
   vtkDataSet *input,
-  ttk::Triangulation *triangulation,
-  std::vector<std::vector<ttk::SimplexId>> &trajectories,
+  triangulationType *triangulation,
```
julien-tierny: I believe this parameter can be `const`.
please double-check each function parameter and declare these which are read-only as `const`

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
```diff
@@ -40,25 +45,62 @@ int ttkIntegralLines::FillOutputPortInformation(int port,
   return 1;
 }
 
+template <typename triangulationType>
```
julien-tierny, line 46: in the future, thanks for providing a quick comment right before each function to explain what it does.

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
```diff
@@ -139,14 +239,167 @@ int ttkIntegralLines::RequestData(vtkInformation *ttkNotUsed(request),
...
+  int totalSeeds;
+#else
+#ifdef TTK_ENABLE_MPI
+  int totalSeeds;
+#endif
+#endif
+
+#ifdef TTK_ENABLE_MPI_TIME
+  ttk::Timer t_mpi;
+  ttk::startMPITimer(t_mpi, ttk::MPIrank_, ttk::MPIsize_);
```
julien-tierny, line 238: what is the purpose of that timer?
in general, time performance should only be measured within the base code (at least for the performance numbers to report in a paper).

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
julien-tierny: I have just come to realize that such a struct already exists in the base code (`ttk::intgl::IntegralLine`). Good!
Please use it here too.

core/base/integralLines/IntegralLines.h
```diff
@@ -14,14 +16,89 @@
...
+#define IS_MESSAGE_SIZE 1
```
julien-tierny: same as above

core/base/integralLines/IntegralLines.h
```diff
@@ -72,9 +173,158 @@ namespace ttk {
...
+     * vertex of an integral lines. It represents the number of the vertex in
+     * the integral line.
+     * @param triangulation Triangulation
+     * @return int
+     */
+    template <typename triangulationType>
+    int getGlobalIdentifiers(
+      std::vector<ttk::SimplexId> &globalVertexId,
+      std::vector<ttk::SimplexId> &globalCellId,
+      const std::vector<ttk::ArrayLinkedList<ttk::SimplexId, TABULAR_SIZE>>
```
julien-tierny: same remark as before (in the vtk layer, where I started): why not manipulating a `std::vector<ttk::ArrayLinkedList<ttk::intgl::IntegralLine>, TABULAR_SIZE>`?

that'd be so much more readable.

julien-tierny:
hi eve, 
here's a global remark as I'm progressing through my review.
integral line forking should be optional, as a checkbox in the ParaView property panel, set to false by default (both in the paraview xml file and in the declaration of the corresponding variable).

core/base/integralLines/IntegralLines.h
julien-tierny: let's leave it here for now then and reconsider later if we bump into a similar setting.

core/base/integralLines/IntegralLines.h
```diff
@@ -72,9 +173,158 @@ namespace ttk {
...
+        *localVertexIdentifiers) {
+      outputLocalVertexIdentifiers_ = localVertexIdentifiers;
+    }
+
+    int preconditionTriangulation(ttk::AbstractTriangulation *triangulation) {
+      int status = triangulation->preconditionVertexNeighbors();
+      // For critical points computation
+      status += triangulation->preconditionVertexStars();
+      status += triangulation->preconditionBoundaryVertices();
+      this->scalarFieldCriticalPoints_.setDomainDimension(2);
```
julien-tierny: this is incorrect.
at minima, this should be `this->scalarFieldCriticalPoints_.setDomainDimension(triangulation->getDimensionality());` (this could be 1, 2, 3...)

what would be best here is to simply call
`this->scalarFieldCriticalPoints_.preconditionTriangulation(triangulation);`
that will take care of everything.

core/base/integralLines/IntegralLines.h
```diff
@@ -93,145 +343,792 @@ namespace ttk {
...
+    ttk::SimplexId seedNumber_;
+    ttk::SimplexId chunkSize_;
+    ttk::SimplexId direction_;
     void *inputScalarField_;
-    const SimplexId *inputOffsets_;
-    SimplexId *vertexIdentifierScalarField_;
-    std::vector<std::vector<SimplexId>> *outputTrajectories_;
+    const ttk::SimplexId *inputOffsets_;
+    std::vector<ttk::SimplexId> *vertexIdentifierScalarField_;
+    std::vector<ttk::ArrayLinkedList<std::vector<ttk::SimplexId>, TABULAR_SIZE>>
```
julien-tierny: same remark as before with the `ttk::intgl::IntegralLine`

core/base/integralLines/IntegralLines.h
```diff
@@ -72,9 +173,158 @@ namespace ttk {
       direction_ = direction;
     }
 
-    int preconditionTriangulation(
-      ttk::AbstractTriangulation *triangulation) const {
-      return triangulation->preconditionVertexNeighbors();
+    void findNextVertex(ttk::SimplexId &vnext,
```
julien-tierny, line 195: could you provide a brief documentation about what this function is doing (as you did above). thanks!

core/base/integralLines/IntegralLines.h
```diff
@@ -93,145 +343,792 @@ namespace ttk {
...
+        ttk::intgl::finishedElement_++;
+      } else {
+        this->storeToSendIfNecessary<triangulationType>(
+          triangulation, integralLine, isMax);
+      }
+#endif
+    } else {
+      // If the integral line continues, the computation continues using lower
+      // and upper components
+      components = &(lowerComponents);
```
julien-tierny, line 543: right below this line, I believe the following lines need to be added (and they should be removed from the code below):
```
if(direction_ == static_cast<int>(Direction::Forward)) {
          components = &(upperComponents);
}
```

core/base/integralLines/IntegralLines.h
```diff
@@ -93,145 +343,792 @@ namespace ttk {
...
+      if(criticalType == (char)CriticalType::Saddle1
+         || criticalType == (char)CriticalType::Saddle2
+         || criticalType == (char)CriticalType::Degenerate) {
+        if(direction_ == static_cast<int>(Direction::Forward)) {
```
julien-tierny: these 3 lines should be removed from this block, see above.

core/base/integralLines/IntegralLines.h
```diff
@@ -93,145 +343,792 @@ namespace ttk {
...
+  }
+  return 0;
+}
+
+#ifdef TTK_ENABLE_MPI
+template <typename triangulationType>
+int ttk::IntegralLines::getGlobalIdentifiers(
+  std::vector<ttk::SimplexId> &globalVertexId,
+  std::vector<ttk::SimplexId> &globalCellId,
+  const std::vector<ttk::ArrayLinkedList<ttk::SimplexId, TABULAR_SIZE>>
```
julien-tierny: same remark as before with `ttk::intgl::IntegralLine`

julien-tierny (changes requested):
hi eve,
thanks for the PR. 
the base code is clean in general, but its readability can still be improved a bit.
I haven't looked so closely at the functions `getGlobalIdentifiers` and `exchangeGhosts` but they look ok at first sight.
also, don't forget to make the forking optional (just a boolean check before processing the saddle case).
thanks a lot!

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
eve-le-guillou: This timer is used to compute the preparation time, and later on the execution of the algorithm. For the execution of the algorithm, the timer only measures two functions calls: `execute` and `getGlobalIdentifiers`. 
The other timers for MPI performances in all other filters have been implemented this way.
Do you want me to delete the preparation timer?

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
eve-le-guillou: The changes were done, leading to quite extensive changes overall, but indeed the result is more readable.

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
eve-le-guillou: This was a mistake, thank you for spotting it.

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
eve-le-guillou: Yes, sorry about that, it should be good for all methods now.

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
eve-le-guillou: This was done in the ttkIntegralLines.h file.

eve-le-guillou:
Thank you for this batch of comments.
They led to quite some extensive changes therefore I checked precisely that the results were still correct on my laptop.
Please let me know if there are any other changes that you might want.

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
julien-tierny: my bad! thanks!

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
julien-tierny: awesome, thanks!

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
julien-tierny: thanks for the clarification.
you can keep the preparation timer, but I don't believe we will include these numbers in individual performance numbers (only `execute` and maybe `getGlobalIdentifiers`)

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
julien-tierny: thanks!

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
```diff
@@ -40,12 +43,18 @@ int ttkIntegralLines::FillOutputPortInformation(int port,
...
 int ttkIntegralLines::getTrajectories(
   vtkDataSet *input,
-  ttk::Triangulation *triangulation,
-  std::vector<std::vector<ttk::SimplexId>> &trajectories,
+  const triangulationType *triangulation,
+  const std::vector<
+    ttk::ArrayLinkedList<ttk::intgl::IntegralLine, INTEGRAL_LINE_TABULAR_SIZE>>
+    &integralLines,
+#ifdef TTK_ENABLE_MPI
+  const std::vector<ttk::SimplexId> &globalVertexId,
```
julien-tierny, line 54: quick question: why no moving the `std::vector<ttk::SimplexId> &globalVertexId` and the `std::vector<ttk::SimplexId> &globalCellId` to the `ttk::intgl::IntegralLine`?

julien-tierny (changes requested):
thanks a lot @eve-le-guillou for this careful review.
I had one last suggestion/question

core/vtk/ttkIntegralLines/ttkIntegralLines.cpp
eve-le-guillou: The `ttk::intgl::IntegralLine` is used all along the computation of the algorithm, while the `globalVertexId` and `globalCellId` vectors are only used afterwards during the computation of the global ids. It wouldn't be efficiently possible to pre-compute the global ids during the computation of the integral lines, therefore they don't need to go together.
Also, currently the `exchangeGhosts` function makes use of accessing to the elements of `globalVertexId` and `globalCellId` using indexes. It could be possible to do it without that in a vector of `ArrayLinkedList`, but it would be a lot less readable.

julien-tierny: alright, thanks for the clarification.

julien-tierny (approved):

julien-tierny:
ok, let's go for it!

PR #907, opened 2023-01-31
[MPI] Add support for processes with empty data set

eve-le-guillou:
Hi all,

This Pull Request adds support in the case where one or more process has an empty data set. Up until now, TTK stops the computation when the data set contains no vertices or cells. However, in the case of MPI, it is possible that a process has an empty data set but the overall domain is not empty. 

In its current state, TTK will hang during preconditioning, as empty processes will exit prematurely, but the others will enter collective communications that require all processes to execute.

I have made the choice to create new MPI communicators rather than ensure that all processes, even empty ones, reach collective communications, because the former would force processes to check if the data set is empty at every turn, increasing greatly the number of tests and modifying significantly the TTK code. 

In `GetTriangulation`, a call to `updateMPICommunicator` has been added. In this method, the MPI communicator is split in two communicators: one that contains all processes with no vertices or cells and one that contains processes with data sets that are not empty.

If there are indeed empty data sets, `MPIrank_`, `MPIsize_` and the RankArrays (in explicite) need to be computed again. For the RankArray, it is done quickly by using the previous values and simply replacing them with the new. The debug prefix message needs also to be updated as `MPIRank_` has been modified. This is done is the method `updateDebugPrefix`.

In case the triangulation is empty, `GetTriangulation` returns a `nullptr`. Filters with MPI support have been modified so that when a triangulation is equal to `nullptr`, an error is returned only in case MPI is not running. If MPI is running, we exit the filter, but without any error signal.

Similarly, the Explicit Triangulation has been modified to print error messages in case the data set is empty only in the case that the execution is not using MPI.

Any feedback is welcome,

Eve

### DIFF
--- a/core/base/common/Debug.h
+++ b/core/base/common/Debug.h
@@ -360,29 +360,18 @@ namespace ttk {
      * debug message.
      */
     inline void setDebugMsgPrefix(const std::string &prefix) {
+      this->debugMsgNamePrefix_ = prefix;
 #if TTK_ENABLE_MPI
       this->debugMsgPrefix_
-        = prefix.length() > 0
-            ? "[" + prefix + "-" + std::to_string(MPIrank_) + "] "
+        = debugMsgNamePrefix_.length() > 0
+            ? "[" + debugMsgNamePrefix_ + "-" + std::to_string(MPIrank_) + "] "
             : "";
 #else
-      this->debugMsgPrefix_ = prefix.length() > 0 ? "[" + prefix + "] " : "";
+      this->debugMsgPrefix_ = debugMsgNamePrefix_.length() > 0
+                                ? "[" + debugMsgNamePrefix_ + "] "
+                                : "";
 #endif
     }
-
-#ifdef TTK_ENABLE_MPI
-    /**
-     * This method should be overloaded in all classes that have MPI support.
-     *
-     */
-    inline void updateDebugPrefix() {
-      this->debugMsgPrefix_ = this->debugMsgPrefix_.length() > 0
-                                ? this->debugMsgPrefix_.substr(
-                                    0, this->debugMsgPrefix_.length() - 3)
-                                    + std::to_string(MPIrank_) + "] "
-                                : "";
-    };
-#endif
 
   protected:
     mutable int debugLevel_;
@@ -390,6 +379,7 @@ namespace ttk {
     COMMON_EXPORTS static debug::LineMode lastLineMode;
 
     std::string debugMsgPrefix_;
+    std::string debugMsgNamePrefix_;
 
     /**
      * Internal debug method that formats debug messages.
--- a/core/base/common/Debug.h
+++ b/core/base/common/Debug.h
@@ -375,9 +375,12 @@ namespace ttk {
      * This method should be overloaded in all classes that have MPI support.
      *
      */
-    inline virtual void updateDebugPrefix() {
-      this->printWrn("The method updateDebugPrefix has not been implemented in "
-                     "the current class");
+    inline void updateDebugPrefix() {
+      this->debugMsgPrefix_ = this->debugMsgPrefix_.length() > 0
+                                ? this->debugMsgPrefix_.substr(
+                                    0, this->debugMsgPrefix_.length() - 3)
+                                    + std::to_string(MPIrank_) + "] "
+                                : "";
     };
 #endif
 
--- a/core/base/common/Debug.h
+++ b/core/base/common/Debug.h
@@ -370,6 +370,17 @@ namespace ttk {
 #endif
     }
 
+#ifdef TTK_ENABLE_MPI
+    /**
+     * This method should be overloaded in all classes that have MPI support.
+     *
+     */
+    inline virtual void updateDebugPrefix() {
+      this->printWrn("The method updateDebugPrefix has not been implemented in "
+                     "the current class");
+    };
+#endif
+
   protected:
     mutable int debugLevel_;
 
--- a/core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
+++ b/core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
@@ -471,15 +471,27 @@ int ttkAlgorithm::updateMPICommunicator(vtkDataSet *input) {
 
 bool ttkAlgorithm::checkGlobalIdValidity(ttk::LongSimplexId *globalIds,
                                          ttk::SimplexId simplexNumber,
-                                         unsigned char *ghost) {
+                                         unsigned char *ghost,
+                                         int *rankArray) {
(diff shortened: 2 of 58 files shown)


core/base/common/Debug.h
```diff
@@ -370,6 +370,20 @@ namespace ttk {
...
 
+#ifdef TTK_ENABLE_MPI
+    /**
+     * This method should be overloaded in all classes that have MPI support.
+     *
+     */
+    inline void updateDebugPrefix() {
+      this->debugMsgPrefix_ = this->debugMsgPrefix_.length() > 0
+                                ? this->debugMsgPrefix_.substr(
+                                    0, this->debugMsgPrefix_.length() - 3)
```
julien-tierny: I don't get where the `3` is coming from (not sure if this should not be different with a larger number of processes)

julien-tierny: I am not sure I understand the rationale behind that function.
what is the problem again?

if it's just the module name which gets incorrect (and still, this is not clear for me why this would be the case), then I'd suggest to add an internal variable `debugMsgNamePrefix_` which would simply copy the content of the parameter `prefix` upon the first call to `setDebugMsgPrefix`.
then, for the function `updateDebugPrefix`, you could simply call `setDebugMsgPrefix(debugMsgNamePrefix_);`

core/vtk/ttkAlgorithm/ttkAlgorithm.cpp
```diff
@@ -431,6 +434,41 @@ int ttkAlgorithm::RequestDataObject(vtkInformation *ttkNotUsed(request),
...
+    }
+    int *cellRankArray
+      = ttkUtils::GetPointer<int>(input->GetCellData()->GetArray("RankArray"));
+    if(cellRankArray != nullptr) {
+      for(int i = 0; i < input->GetNumberOfCells(); i++) {
+        cellRankArray[i] = oldToNewRanks[cellRankArray[i]];
+      }
+    }
+  }
+  this->updateDebugPrefix();
```
julien-tierny: thanks for adding a comment to explain why this call is necessary.

core/vtk/ttkDiscreteGradient/ttkDiscreteGradient.cpp
```diff
@@ -242,6 +242,18 @@ int ttkDiscreteGradient::RequestData(vtkInformation *ttkNotUsed(request),
   auto outputCriticalPoints = vtkPolyData::GetData(outputVector, 0);
   auto outputGradientGlyphs = vtkPolyData::GetData(outputVector, 1);
 
+  auto triangulation = ttkAlgorithm::GetTriangulation(input);
+  if(!triangulation) {
```
julien-tierny: could you wrap these tests into one test function in `core/base/common/MPIUtils.h`?
that way, for the calling module, that would be only a one-liner to call each time.

core/vtk/ttkScalarFieldCriticalPoints/ttkScalarFieldCriticalPoints.cpp
```diff
@@ -56,8 +56,17 @@ int ttkScalarFieldCriticalPoints::RequestData(
   vtkPolyData *output = vtkPolyData::GetData(outputVector, 0);
 
   ttk::Triangulation *triangulation = ttkAlgorithm::GetTriangulation(input);
-  if(!triangulation)
-    return 0;
+  if(!triangulation) {
+#ifdef TTK_ENABLE_MPI
```
julien-tierny: see above comment

core/vtk/ttkScalarFieldNormalizer/ttkScalarFieldNormalizer.cpp
```diff
@@ -97,10 +97,16 @@ int ttkScalarFieldNormalizer::RequestData(vtkInformation *ttkNotUsed(request),
   // get input scalar field
   vtkDataArray *inputArray = this->GetInputArrayToProcess(0, inputVector);
   if(inputArray == nullptr) {
-    this->printErr("No such input scalar field");
-    return 0;
+#ifdef TTK_ENABLE_MPI
+    if(ttk::isRunningWithMPI()) {
```
julien-tierny: see above comment

core/vtk/ttkScalarFieldSmoother/ttkScalarFieldSmoother.cpp
```diff
@@ -48,8 +48,17 @@ int ttkScalarFieldSmoother::RequestData(vtkInformation *ttkNotUsed(request),
 
   Triangulation *triangulation = ttkAlgorithm::GetTriangulation(input);
 
-  if(!triangulation)
-    return 0;
+  if(!triangulation) {
+#ifdef TTK_ENABLE_MPI
+    if(ttk::isRunningWithMPI()) {
```
julien-tierny: see above comment

julien-tierny (changes requested):
thanks!
I have a few remarks. thanks for addressing them!

core/base/common/Debug.h
eve-le-guillou: The module name becomes wrong for the following reason:
In this PR, we add the possibility of having processes that are empty. We allow this by computing new communicators, one that holds all the empty processes, and one that holds all the non-empty processes. Therefore, when an empty process appears, the size and rank of the processes are modified. The debug prefix variable contains the rank of the process and the check is done after the creation of the debug prefix, therefore the debug prefix contains the wrong rank.

The use of a debugMsgNamePrefix_ deletes the need for that `3`

core/vtk/ttkDiscreteGradient/ttkDiscreteGradient.cpp
eve-le-guillou: I wrote this in the method `continueComputation` in ttkAlgorithm, as `MPIUtils.h` is for MPI related code.

julien-tierny: that's great @eve-le-guillou. that does the trick exactly.
one thing though: could you change the name of the function such that it becomes more obvious for other developers that this function is MPI-related (I'm thinking of those who will want to port their module to MPI).
maybe something like:
`checkMPIinput`

eve-le-guillou:
I've changed the name of the function.

Following the merge of PR #904, I've been able to perform additional testing, which brought to light some more bugs: mainly, when the pipeline is recomputed, the communicator's value taken to recompute the pipeline is the value of the communicator at the end of the previous execution.
This was solved by always applying `MPI_Comm_split` to `MPI_COMM_WORLD` instead of `ttk::MPIcomm_` and moving `updateMPICommunicator` to the `ProcessRequest` function of `ttkAlgorithm`, so it is always executed only once by all processes.

julien-tierny:
it's all good, thanks!

PR #901, opened 2023-01-20
GitHub Actions: Workaround for windows job

pierre-guillou:
This PR contains a small workaround to make the windows job of the `test_ci` workflow functional again.

Enjoy,
Pierre

PR #908, opened 2023-01-31
GitHub Actions: More robust fix for windows job

pierre-guillou:
This PR offers a more robust fix to the Windows job in the `test_build` CI workflow (with a first workaround in #901) by pinning a better Python version in conda (which does not seem to be the one used to build TTK-ParaView...).

This also reverts most of #901, only the pinning of the Python version persists (for now).

Enjoy,
Pierre

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -422,7 +422,7 @@ jobs:
       shell: bash
       run: |
         conda install -c conda-forge boost glew eigen spectralib zfp \
-          scikit-learn openmp graphviz ninja websocketpp sccache python=3.9.12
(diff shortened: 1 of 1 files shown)


julien-tierny:
looks great.
should ttk-paraview be updated to use the same version of python?

pierre-guillou:
> should ttk-paraview be updated to use the same version of python?

Good question. For now I would keep the ttk-paraview packages as they are already (kind-of) working. I need to test more the interactions between the Python version used to build ttk-paraview and the one used to build & run with TTK (they can differ depending on the `conda` internal solver since we use a superset of ttk-paraview dependencies for ttk).

PR #909, opened 2023-02-09
[Common] Add memory usage tool

eve-le-guillou:
Hi all,

This PR adds a memory measuring function in the common functionalities.

It uses `getrusage` to find the memory used during the execution of the process. The output quantity is in Kilo Bytes.

If TTK is compiled and executed with MPI, the maximum use of memory is computed and can be found on process 0.

Here is a snippet of example code:

    ttk::Memory m;
    float usedMemory = m.getTotalUsage();


julien-tierny:
looks good, thanks!

PR #910, opened 2023-02-13
[VTK Layer] The ttkLDistance module now writes the Lp-distance in the…

atalon-lip6:
Small addition in the VTK Layer so that the ttkLDistance module stores  the Lp-distance in the field data of its output.


julien-tierny:
great, thanks!

PR #911, opened 2023-02-14
The switch between the legacy and the lts versions of the topological simplification are now done in the ttk layer. Also, fixing the lts module so that each function returns 0 if successful, 1 on errors.

atalon-lip6:
This PR contains two commits.

The first one modifies both the VTK and the TTK layer for the topological simplification. The TopologicalSimplificaiton class is now a container with two members : one object from the legacy implementation (formerly in the TopologicalSimplification class) and one from the lts version. We can set the backend for the new TopologicalSimplification object, which then runs the appropriate algorithm, all in the TTK Layer.

The second was prompted by a note in the code requiring to make lts use the same status return convention as the other ttk modules: return 0 in case of success and something else on errors. This commit fixes this problem for all functions and subfunctions of this module, updating the few places in the other modules which explicitely used lts.

### DIFF
--- a/core/base/legacyTopologicalSimplification/LegacyTopologicalSimplification.h
+++ b/core/base/legacyTopologicalSimplification/LegacyTopologicalSimplification.h
@@ -178,10 +178,6 @@ namespace ttk {
       return 0;
     }
 
-    inline void setNumberOfVertices(size_t n) {
-      vertexNumber_ = n;
-    }
-
     inline void setConsiderIdentifierAsBlackList(bool onOff) {
       considerIdentifierAsBlackList_ = onOff;
     }
--- a/core/base/legacyTopologicalSimplification/LegacyTopologicalSimplification.h
+++ b/core/base/legacyTopologicalSimplification/LegacyTopologicalSimplification.h
@@ -1,0 +1,503 @@
+/// \ingroup base
+/// \class ttk::LegacyTopologicalSimplification
+/// \author Julien Tierny <julien.tierny@lip6.fr>
+/// \author Guillaume Favelier <guillaume.favelier@lip6.fr>
+/// \date February 2016
+///
+/// \brief TTK processing package for the topological simplification of scalar
+/// data.
+///
+/// Given an input scalar field and a list of critical points to remove, this
+/// class minimally edits the scalar field such that the listed critical points
+/// disappear. This procedure is useful to speedup subsequent topological data
+/// analysis when outlier critical points can be easily identified. It is
+/// also useful for data simplification.
+///
+/// \b Related \b publications \n
+/// "Generalized Topological Simplification of Scalar Fields on Surfaces" \n
+/// Julien Tierny, Valerio Pascucci \n
+/// Proc. of IEEE VIS 2012.\n
+/// IEEE Transactions on Visualization and Computer Graphics, 2012.
+///
+/// "Localized Topological Simplification of Scalar Data"
+/// Jonas Lukasczyk, Christoph Garth, Ross Maciejewski, Julien Tierny
+/// Proc. of IEEE VIS 2020.
+/// IEEE Transactions on Visualization and Computer Graphics
+///
+/// \sa ttkLegacyTopologicalSimplification.cpp %for a usage example.
+///
+/// \b Online \b examples: \n
+///   - <a
+///   href="https://topology-tool-kit.github.io/examples/1manifoldLearning/">1-Manifold
+///   Learning example</a> \n
+///   - <a
+///   href="https://topology-tool-kit.github.io/examples/1manifoldLearningCircles/">1-Manifold
+///   Learning Circles example</a> \n
+///   - <a
+///   href="https://topology-tool-kit.github.io/examples/2manifoldLearning/">
+///   2-Manifold Learning example</a> \n
+///   - <a
+///   href="https://topology-tool-kit.github.io/examples/BuiltInExample1/">BuiltInExample1
+///   example</a> \n
+///   - <a
+///   href="https://topology-tool-kit.github.io/examples/contourTreeAlignment/">Contour
+///   Tree Alignment example</a> \n
+///   - <a href="https://topology-tool-kit.github.io/examples/ctBones/">CT Bones
+///   example</a> \n
+///   - <a href="https://topology-tool-kit.github.io/examples/dragon/">Dragon
+///   example</a> \n
+///   - <a
+///   href="https://topology-tool-kit.github.io/examples/harmonicSkeleton/">
+///   Harmonic Skeleton example</a> \n
+///   - <a
+///   href="https://topology-tool-kit.github.io/examples/imageProcessing/">Image
+///   Processing example</a> \n
+///   - <a
+///   href="https://topology-tool-kit.github.io/examples/interactionSites/">
+///   Interaction sites</a> \n
+///   - <a
+///   href="https://topology-tool-kit.github.io/examples/karhunenLoveDigits64Dimensions/">Karhunen-Love
+///   Digits 64-Dimensions example</a> \n
+///   - <a
+///   href="https://topology-tool-kit.github.io/examples/morsePersistence/">Morse
+///   Persistence example</a> \n
+///   - <a
+///   href="https://topology-tool-kit.github.io/examples/morseSmaleQuadrangulation/">Morse-Smale
+///   Quadrangulation example</a> \n
+///   - <a
+///   href="https://topology-tool-kit.github.io/examples/persistenceClustering0/">Persistence
+///   clustering 0 example</a> \n
+///   - <a
+///   href="https://topology-tool-kit.github.io/examples/persistenceClustering0/">Persistence
+///   clustering 1 example</a> \n
+///   - <a
+///   href="https://topology-tool-kit.github.io/examples/persistenceClustering0/">Persistence
+///   clustering 2 example</a> \n
+///   - <a
+///   href="https://topology-tool-kit.github.io/examples/persistenceClustering0/">Persistence
+///   clustering 3 example</a> \n
+///   - <a
+///   href="https://topology-tool-kit.github.io/examples/persistenceClustering0/">Persistence
+///   clustering 4 example</a> \n
+///   - <a
+///   href="https://topology-tool-kit.github.io/examples/tectonicPuzzle/">Tectonic
+///   Puzzle example</a> \n
+///   - <a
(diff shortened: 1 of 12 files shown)


julien-tierny:
can you please investigate the CI fail?

atalon-lip6:
> can you please investigate the CI fail?

I am doing it. Apparently it's a false positive (one uploaded and needed artifact not found), so I am rerunning the tests.

atalon-lip6:
> > can you please investigate the CI fail?
> 
> I am doing it. Apparently it's a false positive (one uploaded and needed artifact not found), so I am rerunning the tests.

But there was also something more, the tests are taking ages now. I will have a look at it, some ContourTree computations aborted and done again a lot of times.

julien-tierny:
> > > can you please investigate the CI fail?
> > 
> > 
> > I am doing it. Apparently it's a false positive (one uploaded and needed artifact not found), so I am rerunning the tests.
> 
> But there was also something more, the tests are taking ages now. I will have a look at it, some ContourTree computations aborted and done again a lot of times.

it seems that this PR is actually breaking the example `states/contourTreeAlignment.pvsm`.
thanks for double-checking this (for instance by running it locally, I suspect it may go in an infinite loop, this is what it looks like it's doing in the CI).

atalon-lip6:
> > > > can you please investigate the CI fail?
> > > 
> > > 
> > > I am doing it. Apparently it's a false positive (one uploaded and needed artifact not found), so I am rerunning the tests.
> > 
> > 
> > But there was also something more, the tests are taking ages now. I will have a look at it, some ContourTree computations aborted and done again a lot of times.
> 
> it seems that this PR is actually breaking the example `states/contourTreeAlignment.pvsm`. thanks for double-checking this (for instance by running it locally, I suspect it may go in an infinite loop, this is what it looks like it's doing in the CI).

Indeed, I noticed that. In the end it was a small mistake in the second commit: I missed one replacement to do.
It's all clean now :-)

core/base/legacyTopologicalSimplification/LegacyTopologicalSimplification.h
```diff
@@ -0,0 +1,503 @@
+/// \ingroup base
```
julien-tierny, line 1: to add in core/base/topologicalSimplification.

core/base/legacyTopologicalSimplification/LegacyTopologicalSimplification.h
```diff
@@ -0,0 +1,503 @@
...
+
+    inline int preconditionTriangulation(AbstractTriangulation *triangulation) {
+      if(triangulation) {
+        vertexNumber_ = triangulation->getNumberOfVertices();
+        triangulation->preconditionVertexNeighbors();
+      }
+      return 0;
+    }
+
+    inline void setNumberOfVertices(size_t n) {
```
julien-tierny: to remove.

core/base/topologicalCompression/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ ttk_add_base_library(topologicalCompression
     OtherCompression.h
   DEPENDS
     triangulation
-    topologicalSimplification
+    legacyTopologicalSimplification
```
julien-tierny, line 10: topologicalSimplification

core/base/topologicalCompression/TopologicalCompression.h
```diff
@@ -15,7 +15,7 @@
 
 // base code includes
 #include <FTMTreePP.h>
-#include <TopologicalSimplification.h>
+#include <LegacyTopologicalSimplification.h>
```
julien-tierny, line 18: TopologicalSimplification

core/base/topologicalCompression/TopologicalCompression.h
```diff
@@ -359,7 +359,7 @@ namespace ttk {
 
   protected:
     // General.
-    TopologicalSimplification topologicalSimplification{};
+    LegacyTopologicalSimplification topologicalSimplification{};
```
julien-tierny, line 362: TopologicalSimplification

core/base/topologicalSimplification/TopologicalSimplification.h
```diff
@@ -93,6 +93,8 @@
```
julien-tierny, line 93: add doxygen documentation

core/base/topologicalSimplification/TopologicalSimplification.h
```diff
@@ -168,149 +120,38 @@ namespace ttk {
...
-
-    inline void setAddPerturbation(bool onOff) {
-      addPerturbation_ = onOff;
-    }
-
   protected:
-    SimplexId vertexNumber_{};
-    bool considerIdentifierAsBlackList_{false};
-    bool addPerturbation_{false};
+    enum TOPO_BACKEND { TOPO_LTS, TOPO_LEGACY };
```
julien-tierny: create a dedicated namespace within ttk, before the class declaration (replicate the mechanism for the Priority levels of the debug class).

```
namespace ttk{
  namespace ts{
     enum class Backend{
        LEGACY,
        LTS
     }
  }
}
```

then, to use: ttk::ts::Backend::LTS (for example)

core/base/topologicalSimplification/TopologicalSimplification.h
```diff
@@ -168,149 +120,38 @@ namespace ttk {
...
+                       "'legacy', but got: "
+                       + arg);
+    }
 
     inline int preconditionTriangulation(AbstractTriangulation *triangulation) {
       if(triangulation) {
-        vertexNumber_ = triangulation->getNumberOfVertices();
         triangulation->preconditionVertexNeighbors();
+        if(backend_ == TOPO_LEGACY)
+          legacyObject_.setNumberOfVertices(
```
julien-tierny: call legacyObject_.preconditionTriangulation() or ltsObject_.preconditionTriangulation()

julien-tierny: do the setDebugLevel + setThreadNumber here

core/base/topologicalSimplification/TopologicalSimplification.h
```diff
@@ -168,149 +120,38 @@ namespace ttk {
                 const SimplexId *const inputOffsets,
                 SimplexId *const offsets,
                 const SimplexId constraintNumber,
-                const triangulationType &triangulation) const;
+                const bool addPerturbation,
+                const triangulationType &triangulation);
+
+    inline void setBackend(const std::string &arg) {
```
julien-tierny: not with a string but enum

core/vtk/ttkTopologicalSimplification/ttkTopologicalSimplification.cpp
```diff
@@ -113,37 +111,20 @@ int ttkTopologicalSimplification::RequestData(
                                                  ttk::VertexScalarFieldName,
                                                  constraints, idSpareStorage);
 
-  // NOTE it'd be better if the two backends were inheriting from the same API
-  // (the switch would then happen in the base code)
+  if(!this->UseLTS) {
+    this->setBackend("legacy");
```
julien-tierny: not with a string but enum

julien-tierny (changes requested):
thanks!
please consider my remarks.

core/base/topologicalSimplification/TopologicalSimplification.h
```diff
@@ -319,181 +160,26 @@ int ttk::TopologicalSimplification::execute(
...
-    if(addPerturbation_)
-      addPerturbation<dataType>(outputScalars, offsets);
-
-    if(!needForMoreIterations)
-      break;
+  const bool addPerturbation,
+  const triangulationType &triangulation) {
+  switch(backend_) {
+    case TOPO_LTS:
+      ltsObject_.setDebugLevel(debugLevel_);
```
julien-tierny: move to preconditioning

core/base/topologicalSimplification/TopologicalSimplification.h
```diff
@@ -319,181 +160,26 @@ int ttk::TopologicalSimplification::execute(
...
+      ltsObject_.setThreadNumber(threadNumber_);
```
julien-tierny: move to preconditioning

core/base/topologicalSimplification/TopologicalSimplification.h
```diff
@@ -319,181 +160,26 @@ int ttk::TopologicalSimplification::execute(
...
+      return ltsObject_
+        .removeUnauthorizedExtrema<dataType, SimplexId, triangulationType>(
+          outputScalars, offsets, &triangulation, identifiers, constraintNumber,
+          addPerturbation);
+    case TOPO_LEGACY:
+      legacyObject_.setDebugLevel(debugLevel_);
```
julien-tierny: move to preconditioning

core/base/topologicalSimplification/TopologicalSimplification.h
```diff
@@ -319,181 +160,26 @@ int ttk::TopologicalSimplification::execute(
...
+      legacyObject_.setThreadNumber(threadNumber_);
```
julien-tierny: move to preconditioning

julien-tierny (changes requested):
(added further clarifications)

core/base/topologicalCompression/TopologicalCompression.h
atalon-lip6: It seems to me that this variable is never used.

core/base/topologicalSimplification/TopologicalSimplification.h
julien-tierny: maybe use the PersistenceDiagram style instead (enum declared within the class, simpler)

core/base/topologicalCompression/CMakeLists.txt
julien-tierny: use the legacy instead (based on discussions) because the compression is using specific features.

core/base/topologicalCompression/TopologicalCompression.h
julien-tierny: use the legacy instead (based on discussions) because the compression is using specific features.

core/base/topologicalCompression/TopologicalCompression.h
julien-tierny: use the legacy instead (based on discussions) because the compression is using specific features.

julien-tierny:
alright, looks good. thanks!

PR #912, opened 2023-02-16
Small CI bugfix (parallelisation of python tests)

atalon-lip6:
Commit 576473427759399daf07005e5b24ce068d148b2a added the deletion of a temporary artifact (.deb package of ttk for ubuntus).
If the other python test job is run much later, the artifact was already deleted and it cannot find it.

The deletion is now at the end of one test job, hence only if the other one is not run during 30 minutes will it fail.


julien-tierny:
thanks!

PR #913, opened 2023-03-03
[MPI] Multiple bug corrections (identifiers, scalarFieldCriticalPoints, preconditioning)

eve-le-guillou:
Hi all,

In this Pull Request, multiple fixes are done:

- In `ExplicitTriangulation`, the function `preconditionDistributedVertices` sets its flag `hasPreconditionedDistributedVertices_` to true at the very end of the function. However, inside this function, a call to `preconditionVertexRankArray` is done and requires to use `getVertexRank`, which will print an error message if the flag `hasPreconditionedDistributedVertices_` is not equal to true. I moved the line `hasPreconditionedDistributedVertices_ = true` to the line before  `preconditionVertexRankArray` so it no longer prints out error messages.
- In Identifiers, the type of RankArray was set to ttk::SimplexId, which is false when the identifiers are 64 bits, as RankArray will always be int.
- In `Identifiers`, the dimension of the kd-tree should always be 3 as it represents the number of coordinates of the points in space, and not the dimension of the domain. VTK will always output 3 coordinates for a point.
- In `ScalarFieldCriticalPoints`, a special case was added for dimension 1, as the current code doesn't handle properly points in dimension 1 that have two upper (or lower) connected components.

Thanks for any feedback,

### DIFF
--- a/MPI.md
+++ b/MPI.md
@@ -14,6 +14,8 @@ To specify the number of threads to use during execution, the environment variab
 
 To use Paraview in distributed mode using MPI, one has to use either `pvserver` or `pvbatch`. 
 
+Please note that TTK only supports a number of processes equal to $2^n$, with $n\in \mathbb{N}$. There are no constraints on the number of threads.
+
 #### pvserver
 The following command allows one to start `pvserver` with 4 MPI processes and 8 threads (per process) with OpenMPI:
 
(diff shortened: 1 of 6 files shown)


julien-tierny:
ok, looks good (although I believe the code edit for scalar field critical points can be made more generic/compact. todo later)

PR #914, opened 2023-03-16
NoCopy arguments passing to the nmi-ari, distanceMatrixDistorsion, and LDistanceMatrix/ttktableDistanceMatrix modules

atalon-lip6:
Before this commits, the vtk layers of these modules would call the ttk layers passing for instance vector<int>& or vector<vector<double>>& variables arguments. Then the output arguments would be copied to the vtk structures. Now we pass vector<int*> and vector<vector<double*>> arguments, which are shallow copy from the vtk structures.
When a vector<vector>> is passed, we provide a small function allowing to use a vector<vector<>>, with a negligible time cost.

There is no big time improvement, but a quite big memory consumption reduction.

### DIFF
--- a/core/base/distanceMatrixDistorsion/DistanceMatrixDistorsion.h
+++ b/core/base/distanceMatrixDistorsion/DistanceMatrixDistorsion.h
@@ -33,10 +33,10 @@ namespace ttk {
   public:
     DistanceMatrixDistorsion();
 
-    int execute(const std::vector<std::vector<double>> &highDistMatrix,
-                const std::vector<std::vector<double>> &lowDistMatrix,
+    int execute(const std::vector<double *> &highDistMatrix,
+                const std::vector<double *> &lowDistMatrix,
                 double &distorsionValue,
-                std::vector<double> &distorsionVerticesValues) const;
+                double *distorsionVerticesValues) const;
   };
 
 } // namespace ttk
--- a/core/base/lDistanceMatrix/LDistanceMatrix.h
+++ b/core/base/lDistanceMatrix/LDistanceMatrix.h
@@ -25,9 +25,18 @@ namespace ttk {
       DistanceType = val;
     }
 
-    template <typename T>
-    int execute(std::vector<std::vector<double>> &output,
-                const std::vector<const T *> &inputs,
+    template <typename TIn, typename TOut>
+    int execute(std::vector<TOut *> &output,
+                const std::vector<const TIn *> &inputs,
+                const size_t nPoints) const;
+
+    /// \warning This version of the execute function allocates, in addition to
+    /// the output vector (of size inputs.size()^2) output, an extra vector of
+    /// pointers of size output.size(). Given the size of the output 2D vector,
+    /// this should be negligible.
+    template <typename TIn, typename TOut>
+    int execute(std::vector<std::vector<TOut>> &output,
+                const std::vector<const TIn *> &inputs,
                 const size_t nPoints) const;
 
   protected:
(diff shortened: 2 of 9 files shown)


core/base/lDistanceMatrix/LDistanceMatrix.h
```diff
@@ -25,37 +25,47 @@ namespace ttk {
...
+                                  const std::vector<const TIn *> &inputs,
                                   const size_t nPoints) const {
 
-  const auto nInputs = inputs.size();
-  output.resize(nInputs);
+  const size_t nInputs = inputs.size();
+
+  if(output.size() != nInputs)
+    this->printErr(
+      " When using execute_raw in LDistanceMatrix module, the output must be "
```
julien-tierny: `execute_raw` --> `execute` ?

core/base/lDistanceMatrix/LDistanceMatrix.h
```diff
@@ -25,37 +25,47 @@ namespace ttk {
...
+      "initialized, with the same number of lines as the number of inputs.");
+  for(size_t i = 0; i < nInputs; i++)
+    if(output[i] == NULL)
+      this->printErr(
+        " When using execute_raw in LDistanceMatrix module, the output must be "
```
julien-tierny: same as above.

core/base/lDistanceMatrix/LDistanceMatrix.h
```diff
@@ -25,37 +25,47 @@ namespace ttk {
...
 
-    template <typename T>
-    int execute(std::vector<std::vector<double>> &output,
-                const std::vector<const T *> &inputs,
+    template <typename TIn, typename TOut>
+    int execute(std::vector<TOut *> &output,
+                const std::vector<const TIn *> &inputs,
+                const size_t nPoints) const;
+    template <typename TIn, typename TOut>
+    int execute(std::vector<std::vector<TOut>> &output,
```
julien-tierny, line 38: please add some doxgen documentation (in particular `\warning`) to warn developers that there is a non-negligible memory overhead when using the second version of that function. In particular, the second version is still allocating a vector of `std::vector<TOut *>` of size `output.size()`.

core/base/distanceMatrixDistorsion/DistanceMatrixDistorsion.h
```diff
@@ -33,10 +33,10 @@ namespace ttk {
   public:
     DistanceMatrixDistorsion();
 
-    int execute(const std::vector<std::vector<double>> &highDistMatrix,
-                const std::vector<std::vector<double>> &lowDistMatrix,
+    int execute(const std::vector<double *> &highDistMatrix,
```
julien-tierny, line 36: should the `double` type made more generic (i.e. `THigh` and `TLow`, similarly to LDistanceMatrix)

core/base/distanceMatrixDistorsion/DistanceMatrixDistorsion.h
```diff
@@ -33,10 +33,10 @@ namespace ttk {
...
+                const std::vector<double *> &lowDistMatrix,
```
julien-tierny, line 37: same remark as above.

core/base/distanceMatrixDistorsion/DistanceMatrixDistorsion.h
```diff
@@ -33,10 +33,10 @@ namespace ttk {
...
                 double &distorsionValue,
-                std::vector<double> &distorsionVerticesValues) const;
+                double *distorsionVerticesValues) const;
```
julien-tierny, line 39: same remark as above

core/base/lDistanceMatrix/LDistanceMatrix.h
```diff
@@ -67,3 +77,19 @@ int ttk::LDistanceMatrix::execute(std::vector<std::vector<double>> &output,
...
+
+template <typename TIn, typename TOut>
+int ttk::LDistanceMatrix::execute(std::vector<std::vector<TOut>> &output,
+                                  const std::vector<const TIn *> &inputs,
+                                  const size_t nPoints) const {
+
+  const auto nInputs = inputs.size();
+  output.resize(nInputs);
+
+  std::vector<double *> outputRaw(nInputs);
```
julien-tierny, line 95: `double` -> `TOut`

julien-tierny (changes requested):
Thanks @atalon-lip6 
That looks good.
Thanks for addressing a few comments.

core/base/lDistanceMatrix/LDistanceMatrix.h
atalon-lip6: I understand the request, but I wanted to discuss it a little before. output is a 2D array, so if I have n points, it is of size n^2, so 4n^2 or 8n^2 bytes (float vs double).
I am allocating a new vector of size n, hence of size 4n or 8n (depending if 4-byte pointers or 8-byte ones).
If n <= 1000, then the overhead is at most 8kB, which is negligible to me. If n >= 1000, the new vector is at most 1/1000  or 1/500 of the size of the original output vector, hence negligible.

If that being said you maintain your remark, no problems, I'll add it.

core/base/distanceMatrixDistorsion/DistanceMatrixDistorsion.h
atalon-lip6: Small question: is the overhead of templating (compilation time, maybe executable size) worth it?
Same as above, if you believe it, I'll do the changes.

core/base/lDistanceMatrix/LDistanceMatrix.h
julien-tierny: thanks for adding the warning still (this is not a zero-allocation function)

julien-tierny (changes requested):
.

core/base/distanceMatrixDistorsion/DistanceMatrixDistorsion.h
julien-tierny: at the moment ttkLDistanceMatrix and others (ttkPersistenceDiagramDistanceMatrix, ttkMergeTreeDistanceMatrix) all seem to produce distance matrices encoded with double. so ok for keeping doubles then.

core/base/distanceMatrixDistorsion/DistanceMatrixDistorsion.h
julien-tierny: let's skip that remark for now.

core/base/lDistanceMatrix/LDistanceMatrix.h
julien-tierny: let's skip that remark for now.

core/base/distanceMatrixDistorsion/DistanceMatrixDistorsion.h
julien-tierny: let's skip that remark for now.

julien-tierny:
ok!

PR #915, opened 2023-03-20
[MPI] Array preconditioning performance fix

eve-le-guillou:
Hi all,

While working on a pipeline using a quite substantial dataset, I noticed an unreasonable growth in the memory footprint of the `ArrayPreconditioning` filter.
I looked into the code and made three changes, two small and one that has a big impact:

- I prevented the recomputation of the `gidToLidMap` variable using lambdas as it had been done for RankArrays and GlobalIds
- I prevented the recomputation of neighbors by passing it as an optionnal argument in `ProcessScalarArray`
- Here is the big one: in `produceOrdering`, the vector `finalValues` was reserved a size of `totalSize`, the size of all the points on all the processes, which defeated the purpose of the `burstSize` variable, as it would make the memory footprint on the root process grow enormously. `finalValues` was only used for its size, the elements inserted in it were never actually accessed. I therefore replaced it with a simple counter. This doesn't really change how the whole function works, but it greatly improves the size of the memory footprint, as well as the execution time. 

For a dataset of $512^3$, on 16 nodes, the memory footprint is reduced by 38% compared to before these changes.

Thanks for any feedback,

### DIFF
--- a/core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
+++ b/core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
@@ -1,4 +1,3 @@
-#include "DataTypes.h"
 #include <OrderDisambiguation.h>
 #include <Triangulation.h>
 #include <ttkArrayPreconditioning.h>
--- a/core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
+++ b/core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
@@ -1,3 +1,4 @@
+#include "DataTypes.h"
 #include <OrderDisambiguation.h>
 #include <Triangulation.h>
 #include <ttkArrayPreconditioning.h>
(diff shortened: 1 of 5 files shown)


core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
```diff
@@ -1,3 +1,4 @@
+#include "DataTypes.h"
```
julien-tierny: please use brackets (i.e. `<` and `>` instead of quotes). this would yield missing header errors if that code is directly called by an external program.

julien-tierny (changes requested):
Ok, this looks good to me modulo the minor change in header inclusion.

@m-s-will 
can you double check this is OK for you? thanks!

m-s-will:
Looks good, the values of finalValues were probably used at some point and it evaded the removal when changing to the burst approach 👍

core/vtk/ttkArrayPreconditioning/ttkArrayPreconditioning.cpp
eve-le-guillou: This line was added automatically by my IDE, but is actually useless here.
I deleted it.

julien-tierny:
awesome, thanks everyone!

PR #916, opened 2023-03-24
[MPI] Fix preconditionDistributedCells

eve-le-guillou:
Hi all,

While going through the code, I noticed the attribute `ghostCellsPerOwner_` was created and later exchanged between processes to create `remoteGhostCells_`, but never actually populated, though I knew that it was at some point in the history of the code.

I traced the disappearance back to PR #888 (line 3097 in ImplicitTriangulation.cpp). The line was probably forgotten as the PR was quite substantial.

The goal of `ghostCellsPerOwner_` is to store TTK ghost cells. There is no longer need to translate from VTK to TTK cell identifiers (as it was done in #888) since the method `getCellRankInternal` does it for us.

This wasn't noticed earlier as right now, no filter uses functions that use  `ghostCellsPerOwner_`.

Thanks for any feedback,

### DIFF
--- a/core/base/implicitTriangulation/ImplicitTriangulation.cpp
+++ b/core/base/implicitTriangulation/ImplicitTriangulation.cpp
@@ -3154,6 +3154,16 @@ int ttk::ImplicitTriangulation::preconditionDistributedCells() {
                  MPI_STATUS_IGNORE);
   }
 
+  int cellRank = 0;
+  for(LongSimplexId lcid = 0; lcid < nLocCells; ++lcid) {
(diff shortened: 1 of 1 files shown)


julien-tierny:
looks good thanks!

@pierre-guillou would you agree with that?

pierre-guillou:
I don't remember exactly if what I did for `ghostCellsPerOwner_` in #888 was intentional or not (maybe a case of "if we don't need, we can slash it"). I don't have any problem concerning adding it back if you need to use it in other filters.

Pierre

julien-tierny:
alright, thanks a lot for your feedback!

PR #917, opened 2023-04-01
GitHub Actions: Use curl on Windows 

pierre-guillou:
This PR replaces the PowerShell command `Invoke-WebRequest` in Windows virtual machines with the `curl` program, which is easier to use and lead to more readable commands. Sadly, `wget` is not available on Windows VMs.

Enjoy,
Pierre


julien-tierny:
great stuff! thanks a lot @pierre-guillou !!!

PR #918, opened 2023-04-11
[AssignmentSolver] Auction and Exhaustive clean and optimization

MatPont:
This PR cleans and optimizes the code of the Auction and the Exhaustive assignment solver.

- Changes for Auction solver:
  - Return the best assignment (giving the lowest distance) instead of the assignment of the last iteration.
  - Add some mechanisms of `PersistenceDiagramAuction` like setting the second best value to the best value if there is only one good to bid for (very special case) ; and avoid price explosion if the new price begins to be too high.
  - Add a mechanism to converge faster (especially for close diagrams) by computing a lower bound of the cost. If the cost at the current iteration is below this lower bound times _1 + delta_lim_ then the algorithm stops. This criteria allows to keep the theoretical properties regarding the error to the real distance while converging faster.

- Changes for Exhaustive solver:
  - Add some hard-coded assignments for small problems, these hard-coded assignments can be used to avoid the call for Auction (PR soon for `MergeTreeDistance`).


core/base/assignmentSolver/AssignmentExhaustive.h
```diff
@@ -48,155 +52,9 @@ namespace ttk {
...
-
-      return allAsgn;
-    }
-
-    // std::vector<std::vector<int>>
-    // AssignmentExhaustive::constructAssignments3(int min_dim, int max_dim){
-    std::vector<std::vector<int>> constructAssignments3(unsigned int min_dim,
-                                                        unsigned int max_dim) {
-      std::vector<std::vector<int>> allAsgn;
+    void constructAssignments3(unsigned int min_dim,
```
julien-tierny: are you sure about the name of that function? (is there a `constructAssignment2`?)
could you please rename it with a more explicit name?

julien-tierny (changes requested):
looks great, thanks!
can you address my small remark regarding a function name?

core/base/assignmentSolver/AssignmentExhaustive.h
MatPont: There was a `constructAssignments2` but it is removed by this PR since it is not used anywhere (and is a "worse" version than `constructAssignments3`). I will rename the function to `enumerateAsssignments` since its goal is to enumerate all possible assignments given the dimensions of the cost matrix.

julien-tierny (approved):

julien-tierny:
awesome, thanks!

PR #919, opened 2023-04-11
Cleanup ttk plugin cmake install logic

mwestphal:
- Since only the singlePlugin logic is supported now, simplyify the cmake logic
 - Remove a duplicate installation of the plugin in the bin directory


julien-tierny:
hi @mwestphal 
thanks a lot for your contribution!
however, it seems that this PR is breaking the continuous integration (see for instance https://github.com/topology-tool-kit/ttk/actions/runs/4668519453/jobs/8265713183?pr=919)

julien-tierny:
Also, I had a quick look into this and on my systems, the plugin seems to be installed only once (here in `/usr/local/bin/plugins/`, which I believe is the default path where ParaView will search for plugins, if ParaView was installed in `/usr/local`). So I'm not so sure about the double install remark. Could you please clarify? Thanks!

paraview/CMakeLists.txt
```diff
@@ -43,14 +43,21 @@ paraview_plugin_build(
   INSTALL_HEADERS
     OFF
   RUNTIME_DESTINATION
-    "${CMAKE_INSTALL_BINDIR}/plugins"
+    "${TTK_INSTALL_PLUGIN_DIR}/${TTK_PLUGIN_SUBDIR}"
   LIBRARY_DESTINATION
-    "${CMAKE_INSTALL_BINDIR}/plugins"
+    "${TTK_INSTALL_PLUGIN_DIR}/${TTK_PLUGIN_SUBDIR}"
```
julien-tierny, line 48: Agreed with these two lines.

mwestphal:
I definitelly misunderstood something but I dont know what yet. However it looks like there multiple `install` cmake commands where there should be only one or none at all if handled correctly by paraview plugin mechanism.

> here in /usr/local/bin/plugins/

Expected location is `INSTALL_PREFIX/lib/paraview-5.11/plugins/` not bin.

The reason why I'm trying to do that is that in the paraview superbuild, ttk is installed in bin (and lib) but it should be only in lib.

I need to spend a bit more time on this and understand what it is supposed to do before I can try to fix it.

PR #921, opened 2023-04-14
[integralLines] Fix integralLine Mask with MPI OFF

eve-le-guillou:
Hi all,

This PR aims at fixing the following bug: the Mask value is not correct for the critical point of the end of an integral line.
This is only true when `TTK_ENABLE_MPI` is OFF. This was due to the two deleted lines of this Pull Request. Smoothing will no longer shorten the integral line.

Thanks for any feedback,
Eve


julien-tierny:
awesome, thanks for the hot fix!

PR #922, opened 2023-04-14
[MPI] Simplify order array computation

eve-le-guillou:
Hi all,

The current `ArrayPreconditioning` filter computes an order array for all process. This means that if vertex `v` has an order value of $n$, then this vertex is the $n^{th}$ vertex for all processes. This computation requires to do a parallel sort, which is quite costly and often unnecessary in our use cases. 

The choice is made in this PR to switch to local sorting: each process will compute a local order array. By default, `ttkAlgorithm` will trigger the computation of local order arrays.

Similarly, the filter `ArrayPreconditioning` will now by default compute order arrays locally. However, the code for global order array computation is kept in TTK and can be called in this filter by checking the option `Global Order`.

A small fix for a bug in the `ImplicitTriangulation` on one process is also made here.

Thanks for any feedback,
Eve

### DIFF
--- a/core/base/arrayPreconditioning/ArrayPreconditioning.h
+++ b/core/base/arrayPreconditioning/ArrayPreconditioning.h
@@ -80,6 +80,8 @@ namespace ttk {
       return 1; // return success
     }
 
+  protected:
+    bool GlobalOrder{false};
   }; // ArrayPreconditioning class
 
 } // namespace ttk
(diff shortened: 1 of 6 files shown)


eve-le-guillou:
@m-s-will Does this look good to you?

m-s-will:
Looks good, a local order array with one layer of ghost cells should be sufficient for most filters and the GlobalOrder checkbox will do it for the rest.

julien-tierny:
wonderful, thanks a lot @eve-le-guillou !

PR #924, opened 2023-04-19
[PersistenceDiagramClustering] Auction lower bound trick

MatPont:
This PR adds the auction lower bound trick (previously added in the cost-matrix-based Auction in #918 ) to the `PersistenceDiagramAuction` class. It also modifies classes related to persistence diagram distance, barycenter and clustering to use this feature. For now it is only added in the "Classical Auction Approach" backend and not in the progressive one.


julien-tierny:
awesome, thanks a lot @MatPont !!!

PR #926, opened 2023-04-24
[MergeTreeFeatureTracking] doxygen example links

MatPont:
This PR adds links to the documentation of the related filters of the MergeTreeFeatureTracking example.


julien-tierny:
awesome, thanks a lot!

PR #927, opened 2023-04-24
[PersistenceDiagramClustering] lower bound trick for progressive algo

MatPont:
This PR adds the lower bound trick for the Auction algorithm of persistence diagrams for the progressive algorithm of the `PersistenceDiagramClustering` module.

For now it is only for 2 inputs (without progressivity) and when the option to force the precision is enabled.


julien-tierny:
looks good, thanks!

PR #928, opened 2023-04-27
[PersistenceDiagramClustering] homogenize distance accumulation between classic and progressive

MatPont:
This PR modifies how the classical approach of `PersistenceDiagramClustering` accumulates distances to be like the progressive one. Without this, both backend could display different distance values whereas having same matchings.

The problem was that the classical approach in the case of 2 inputs returned the square root of the sum of the squared distances instead of just the sum of distances.

Some minor modifications in this PR also include:
- Use of OpenMP reduction to accumulate distances, before this PR the different threads computing distances were adding to the `total_cost` variable without any safety. It seems to me that the reduction is necessary here to make these functions thread-safe.
- Change of the *meaning* of the return value of `runMatching` and `runMatchingAuction`. Before this PR, these functions returned the squared distance or the distance to the power of 4 (depending on an input parameter), then the code calling these functions square root the result (to have the distance or the squared distance). From my point of view it is better to have these functions returning directly the distance or the squared distance without having to square root the result outside these functions.

### DIFF
--- a/core/base/persistenceDiagramClustering/PDBarycenter.cpp
+++ b/core/base/persistenceDiagramClustering/PDBarycenter.cpp
@@ -33,11 +33,12 @@ void ttk::PDBarycenter::runMatching(
   std::vector<double> *min_price,
   std::vector<std::vector<MatchingType>> *all_matchings,
   bool use_kdt,
-  int actual_distance) {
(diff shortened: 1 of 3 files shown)


julien-tierny:
ok, thanks!

PR #930, opened 2023-05-05
Fix GCC13 warnings

pierre-guillou:
GCC 13 has arrived this week in ArchLinux repositories and when rebuilding TTK with it, two new warnings popped up. These warnings are linked to the use of `std::array` that are only valid up to 3D datasets (and not more).

This PR fixes these warnings by adding checks to please GCC 13.

Enjoy,
Pierre


julien-tierny:
looks good, thanks!

PR #931, opened 2023-05-05
Patch script for ParaView 5.11.1

pierre-guillou:
ParaView 5.11.1 was released a month ago. This PR provides an updated patch script based on the 5.11.0 version with two small modifications/corrections:

* The Python 3.11/MPI4Py workaround introduced in #891 does not apply anymore since it has been fixed upstream. The corresponding patch is not applied for 5.11.1.
* A missing `#include <cstdint>` has been added in `libproj` to fix a GCC 13 compilation error (I have not tried with other compilers).

Enjoy,
Pierre


julien-tierny:
looks great, thanks!

PR #933, opened 2023-05-09
ImplicitTriangulation: Fix MPI 2D segfault

### LINKED ISSUE #929: [MPI] Crashes when using ImplicitTriangulation in 2D
**Describe the bug**
Paraview crashes when trying to use the ImplicitTriangulation on 2-dimensional data.

**To Reproduce**
1. Run Paraview with mpirun with more than 1 rank
2. Use the 'Wavelet' filter to generate a 2D dataset
3. Use any filter accessing the triangulation e.g. ScalarFieldSmoother
4. Paraview will crash during PreconditionDistributedCells (see the attached screenshot)

**Expected behavior**
It should behave as in the 3D case, by e.g. smoothing the scalar field without crashing

**Screenshots**
Bildschirm­foto 2023-05-04 um 17 13 12

 - OS: Ubuntu 22.04
 - Paraview 5.10.1, current ttk-dev

**Additional context**
@emmwel encountered this problem a while ago and had some fixes, however different problems arose when trying to further work with the 2D data. She can give further context in the comments

pierre-guillou:
This PR adds a workaround when computing the 2D bounding box of an implicit triangulation. It fixes #929.

Also added are two `nullptr` checks inside `ttkAlgorithm.cpp` . Without them, I cannot open the Python state files inside `ttk-data/states` (e.g. `states/cinemaDarkroom.py`) without a segfault.

Enjoy,
Pierre

### DIFF
--- a/core/base/implicitTriangulation/ImplicitTriangulation.cpp
+++ b/core/base/implicitTriangulation/ImplicitTriangulation.cpp
@@ -3115,6 +3115,9 @@ int ttk::ImplicitTriangulation::preconditionDistributedCells() {
       this->tetrahedronToPosition(lcid, p.data());
     } else if(this->dimensionality_ == 2) {
       this->triangleToPosition2d(lcid, p.data());
+      // compatibility with tetrahedronToPosition; fix a bounding box
+      // error in the first axis
+      p[0] /= 2;
     }
 
     // global vertex coordinates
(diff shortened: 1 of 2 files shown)


julien-tierny:
great stuff, thanks!

m-s-will:
triangleToPosition2D is called two other times in `Implicit Triangulation.cpp` when changing  getting the global / local ids for cells. I haven't tested it yet, but could the error appear there too and would it be incorrext to add the division directly into `triangleToPosition2D `?

m-s-will:
When testing this by transforming from the local id to the global and back again it seems like the division should be added everywhere. The first screenshot shows the current state, the second one shows the division added to `triangleToPosition2d` itself, `Local2 = triangulation->getCellLocalId(triangulation->getCellGlobalId(Local))`.

Bildschirm­foto 2023-05-10 um 09 46 23Bildschirm­foto 2023-05-10 um 09 47 25

`triangleToPosition2d` is also implemented in `PeriodicImplicitTriangulation.h`, but afaik there is currently no MPI support for the PeriodicImplicitTriangulation, so this problem will no appear (yet).
I can prepare a PR if you'd like.

julien-tierny:
> I can prepare a PR if you'd like.

That'd be great, thanks!

PR #934, opened 2023-05-09
Remove Windows-specific OpenMP directives

pierre-guillou:
This PR removes old Windows-specific OpenMP directives in the `ContinuousScatterPlot` and `UncertainDataEstimator` modules. These directives should be useless now since with the `clang-cl` frontend, OpenMP is fully supported on Windows. Removing these directives unifies the OpenMP usage in these modules and reduces the Windows-specific code.

In the `ttk-data` repository, the `python/builtInExample2.py` Python script should be the only one affected by this change. The ValidPointMask array values should now be identical between Windows and Linux/macOS. I will do a PR on ttk-data that updates the hashes later.

Enjoy,
Pierre


julien-tierny:
great, thanks.

PR #935, opened 2023-05-09
GitHub Actions: Fix scikit-learn segfault on Windows

pierre-guillou:
After tinkering with a Windows VM, I managed to find a missing Conda directory in the PYTHONPATH environment variable responsible for the segfaults when calling the `DimensionReduction` filter on Windows.

Sadly, this does not fixes (yet) the other Python issue: importing the `topologytoolkit` Python package still ends with an error (similarly when trying to import `paraview.simple` with `python`).

Enjoy,
Pierre


julien-tierny:
thanks!

PR #936, opened 2023-05-10
[MPI] ImplicitTriangulation: Fixes for Cell IDs in 2D

m-s-will:
As an extension of #933 , this PR moves the division to `triangleToPosition2d` itself, to fix some lingering issues with global / local cell id computation.

### DIFF
--- a/core/base/implicitTriangulation/ImplicitTriangulation.cpp
+++ b/core/base/implicitTriangulation/ImplicitTriangulation.cpp
@@ -3115,9 +3115,6 @@ int ttk::ImplicitTriangulation::preconditionDistributedCells() {
       this->tetrahedronToPosition(lcid, p.data());
     } else if(this->dimensionality_ == 2) {
       this->triangleToPosition2d(lcid, p.data());
-      // compatibility with tetrahedronToPosition; fix a bounding box
-      // error in the first axis
-      p[0] /= 2;
     }
 
     // global vertex coordinates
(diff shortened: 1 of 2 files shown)


julien-tierny:
looks good to me.

@pierre-guillou is this OK with you? (cf https://github.com/topology-tool-kit/ttk/pull/933)

julien-tierny:
@pierre-guillou some examples are failing under windows. I suspect this might be a consequence of https://github.com/topology-tool-kit/ttk/pull/934 or https://github.com/topology-tool-kit/ttk/pull/935

what do you think?
(I can simply roll these back to give it a try)

pierre-guillou:
I don't think there is an issue with the CI.
If we look at what failed, it was either the `imageProcessing` and/or `timeTracking` Python scripts, both have 2D regular grids as input.

julien-tierny:
good catch. the CI is running with MPI support with only one process if I'm not mistaken.
some pre-conditioning is enabled while it should not?

m-s-will:
I was checking for this and saw that sometimes preconditioning stuff is encapsulated with `isRunningWithMPI`, some with `hasInitializedMPI` and some not at all. Is this due to functional differences depending on the amount of ranks or was this an oversight?

Bildschirm­foto 2023-05-10 um 14 06 46
Bildschirm­foto 2023-05-10 um 14 07 16
Bildschirm­foto 2023-05-10 um 14 09 14
Bildschirm­foto 2023-05-10 um 14 09 11

pierre-guillou:
With these changes but without the MPI support, my laptop tries to overflow my RAM + swap when launching ParaView on the two offending state files (everything works without).

pierre-guillou:
I don't think we should modify `triangleToPosition2d`. I fear this breaks a lot of assumptions in the `ImplicitTriangulation` module (the method is used in `ImplicitPreconditions` then the result coordinates are accessed through `getTriangleCoords`).

m-s-will:
Yes I reverted it back and added the missing divisions to the correct places, the state files should not fail anymore and the Cell id conversions are correct now too.

pierre-guillou:
You might want to remove the leftover comment in `ImplicitTriangulation.h` to reduce the time spent in recompiling all the dependent modules.

julien-tierny:
great stuff. thanks a lot to the both of you.

PR #937, opened 2023-05-12
[ProjectionFromTable] fix array size

MatPont:
This PR fixes an _out of bound_ access in `ProjectionFromTable` that can lead to non-determinism (due to accessing non-initialized values in the memory).

The problem was that the array `tableValues` had a size of 2 and then the pointer to its data was given to `computeTriangleArea` from `Geometry` that assume size 3, hence accessing something out of the array for the last index.


julien-tierny:
wonderful!

PR #938, opened 2023-05-15
Restrict input domain to point data for persistence diagram filters

CharlesGueunet:
Dear Julien,

This PR restrict the input domain of persistence diagram filters to point data.
Without this change, we can call these filters with unsupported cell data input, leading to undefined behavior.

Best,  
Charles


CharlesGueunet:
@julien-tierny failures do not seems related to these changes

julien-tierny:
looked like a temporary issue with github's VMs.
I re-launched the failed jobs. 
:crossed_fingers:

julien-tierny:
thanks, I'll go ahead and propagate that to the other filters.

PR #939, opened 2023-05-16
Path compression

rgcmaack:
The path compression filter allows to take an input scalar field (vtkDataSet) and compute the ascending and descending Morse segmentations, as well as a Morse-Smale segmentation hash.

### DIFF
--- a/core/base/pathCompression/PathCompression.h
+++ b/core/base/pathCompression/PathCompression.h
@@ -164,7 +164,7 @@ namespace ttk {
     bool ComputeDescendingSegmentation{true};
 
     // Compute Morse-Smale segmentation hash?
-    bool computeMSSegmentationHash{true};
+    bool ComputeMSSegmentationHash{true};
   };
 } // namespace ttk
 
@@ -181,7 +181,7 @@ int ttk::PathCompression::execute(OutputSegmentation &outSegmentation,
                  this->threadNumber_);
 
   if((ComputeAscendingSegmentation && ComputeDescendingSegmentation)
-     || computeMSSegmentationHash) {
+     || ComputeMSSegmentationHash) {
     computePathCompression(outSegmentation.ascending_,
                            outSegmentation.descending_, orderArray,
                            triangulation);
@@ -193,7 +193,7 @@ int ttk::PathCompression::execute(OutputSegmentation &outSegmentation,
       outSegmentation.descending_, false, orderArray, triangulation);
   }
 
-  if(computeMSSegmentationHash) {
+  if(ComputeMSSegmentationHash) {
     computeMSHash(outSegmentation.morseSmale_, outSegmentation.ascending_,
                   outSegmentation.descending_, triangulation);
   }
--- a/core/base/pathCompression/PathCompression.h
+++ b/core/base/pathCompression/PathCompression.h
@@ -159,7 +159,7 @@ namespace ttk {
     template <typename triangulationType>
     int computeMSHash(SimplexId *const morseSmaleSegmentation,
                       const SimplexId *const ascSegmentation,
-                      const SimplexId *const desSegmentation,
+                      const SimplexId *const dscSegmentation,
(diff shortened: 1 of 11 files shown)


julien-tierny:
Thanks a lot @rgcmaack for this PR!
It seems that it still does not go through the entire set of CI tests (for instance doxygen errors).
Could you address that before we move forward?
Thanks!

rgcmaack:
It now failed at installing Ubuntu dependencies. Is there something I can do about that? @julien-tierny 
`E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/universe/w/websocketpp/libwebsocketpp-dev_0.8.1-7_amd64.deb  Unable to connect to azure.archive.ubuntu.com:http: [IP: 40.119.46.219 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?`

julien-tierny:
that's a github issue (sometimes their VMs have network issues). I re-launched the failed jobs.

core/base/pathCompression/PathCompression.h
```diff
@@ -0,0 +1,552 @@
...
+          activeVertices[currentIndex++] = v;
+        }
+      }
+
+      nActiveVertices = currentIndex;
+      currentIndex = 0;
+    }
+
+#ifdef TTK_ENABLE_OPENMP
+  } else {
```
julien-tierny: I don't understand why you have a specialized implementation for 1 thread and one for multiple threads.
Is it to improve the performance of the mono-thread version? 

From a maintenance standpoint, it would be really preferable to have only one implementation (i.e. the multi-threaded one). I don't see a case where optimizing mono-thread execution is relevant anymore.

core/base/pathCompression/PathCompression.h
```diff
@@ -0,0 +1,552 @@
...
+
+#pragma omp parallel num_threads(threadNumber_)
```
julien-tierny: same remark as above.

julien-tierny (changes requested):
Alright, thanks a lot @rgcmaack for this great PR.
It looks good to me and ready to be merged, except one tiny detail that bugs me (below, about the double implementation, mono-thread VS multi-thread). Thanks for your feedback on that. In principle, I'd argue for only one implementation.
@JonasLukasczyk your thoughts?

core/base/pathCompression/PathCompression.h
rgcmaack: The double implementation comes from the issue of guarding everything related to OpenMP/multithreading with "ifdefs". Previous versions needed even more of those guards.

As removing the TTK_ENABLE_OPENMP flag will be another discussion, I will push a single implementation version for now.

julien-tierny: hi @rgcmaack, thanks for your feedback.
I believe you can still have only one implementation and still use `#ifdef TTK_ENABLE_OPENMP`.
See for instance the line 254 of `ScalarFieldCriticalPoints.h`:
https://github.com/topology-tool-kit/ttk/blob/dev/core/base/scalarFieldCriticalPoints/ScalarFieldCriticalPoints.h

julien-tierny: Maybe there is something I do not see which prevents you from using the `#ifdef` as in the above example?

rgcmaack:
> Alright, thanks a lot @rgcmaack for this great PR. It looks good to me and ready to be merged, except one tiny detail that bugs me (below, about the double implementation, mono-thread VS multi-thread). Thanks for your feedback on that. In principle, I'd argue for only one implementation. @JonasLukasczyk your thoughts?

This came from the time where the implementation was changing rapidly. I just pushed a single implementation version.

julien-tierny:
that looks like it, yes. thanks.
I'm trying to fix the windows CI now. I'll get back to you when it's done.

julien-tierny:
alright, looks good!
let's go!

PR #940, opened 2023-05-16
Marching Tetrahedra

rgcmaack:
This plugin adds the Marching tetrahedra algorithm with three options. You can either extract a separating geometry between labels or extract a hull around each label, separating it from the rest of the dataset (fast and detailed version).


julien-tierny:
Thanks a lot @rgcmaack for this PR!
It seems that it still does not go through the entire set of CI tests.
Could you address that before we move forward?
Thanks!

rgcmaack:
@julien-tierny Please restart the jobs.
Thank you :)

julien-tierny:
done. 
thanks for addressing the next issue :)

CharlesGueunet:
Hello,
I am not sure to exactly get what this module compute. In my understanding, marching tet is an algorithm that can compute contours, but here the description mention labels. Can you explain quickly the use case ?

paraview/xmls/MarchingTetrahedra.xml
```diff
@@ -0,0 +1,95 @@
...
+        name="Input"
+        command="SetInputConnection">
+        <ProxyGroupDomain name="groups">
+          <Group name="sources"/>
+          <Group name="filters"/>
+        </ProxyGroupDomain>
+        <DataTypeDomain name="input_type">
+          <DataType value="vtkDataSet"/>
+        </DataTypeDomain>
+        <InputArrayDomain name="input_scalars" number_of_components="1">
```
vollia: maybe restrict to point data as seen in #938  ?

core/base/marchingTetrahedra/MarchingTetrahedra.h
```diff
@@ -0,0 +1,4047 @@
...
+ * (e.g. [0,1] = edge 0, [0,2] = edge 1, [1,2] = edge 3, [0,1,3] = triangle 7).
+ * Every 3 entries represent one generated triangle by using the middle of the
+ * edges and triangles respectively.
+ * The comments next to each line first indicate the number of unique vertex
+ * labels on the tetrahedron, the 2nd label, 3rd label, 4th label, and its index
+ * (the first label is alway considered to be 0).
+ * This allows to directly retrieve the triangulation from this lookup table by
+ * supplying a bit code, where the 4th label supplies the two LSB
+ */
+constexpr int tetLookupWall[28][15] = {
```
vollia: These kind of data files are often added in `.inl`  files. This would keep data / implemetation asides

core/base/marchingTetrahedra/MarchingTetrahedra.h
```diff
@@ -0,0 +1,4047 @@
...
+  const SimplexId nV = triangulation.getNumberOfVertices();
+  const SimplexId nC = triangulation.getNumberOfCells();
+  const int dim = triangulation.getDimensionality();
+
+  unsigned char *tetCases = nullptr;
+  size_t *numberOfTetCases = nullptr;
+  unsigned long long *cScalars = nullptr;
+
+  tetCases = (unsigned char *)malloc(nC * sizeof(unsigned char));
+  cScalars = (unsigned long long *)malloc(nV * sizeof(unsigned long long));
```
vollia: Is there a good reason to use `malloc` instead of an smart container like `vector` ?
I assume this related to the free of the memory, this still seems really dangerous. (Where are those `free`d ?)

core/base/marchingTetrahedra/MarchingTetrahedra.h
```diff
@@ -0,0 +1,4047 @@
...
+
+      if(triangleLookupIs2Label[tetCases[tet]]) {
+        float eC[2][3];
+        getCenter(vertPos[edgeVerts[0]], vertPos[edgeVerts[1]], eC[0]);
+        getCenter(vertPos[edgeVerts[2]], vertPos[edgeVerts[3]], eC[1]);
+
+        const unsigned long long sparseID
+          = getHash(msm[edgeVerts[0]], msm[edgeVerts[1]]);
+
+        p[0] = eC[0][0];
```
vollia, line 527: a small comment regarding these patterns ?

core/vtk/ttkMarchingTetrahedra/ttkMarchingTetrahedra.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include <vtkSignedCharArray.h>
+#include <vtkUnsignedCharArray.h>
+#include <vtkUnsignedLongLongArray.h>
+
+vtkStandardNewMacro(ttkMarchingTetrahedra);
+
+ttkMarchingTetrahedra::ttkMarchingTetrahedra() {
+  this->setDebugMsgPrefix("MarchingTetrahedra");
+  SetNumberOfInputPorts(1);
+  SetNumberOfOutputPorts(1);
```
vollia: this->

(these two line are optional as these are the default)

core/vtk/ttkMarchingTetrahedra/ttkMarchingTetrahedra.cpp
```diff
@@ -0,0 +1,170 @@
...
+                                                     vtkInformation *info) {
+  if(port == 0) {
+    info->Set(vtkDataObject::DATA_TYPE_NAME(), "vtkPolyData");
+    return 1;
+  }
+  return 0;
+}
+
+template <typename vtkArrayType, typename vectorType>
+void setArray(vtkArrayType &vtkArray, vectorType &vector) {
```
vollia: if this is a helper, this should go into an anonymous namespace, at the top of the file not to be confused with method classes.

core/vtk/ttkMarchingTetrahedra/ttkMarchingTetrahedra.cpp
```diff
@@ -0,0 +1,170 @@
...
+  ttkVtkTemplateMacro(inputScalars->GetDataType(), triangulation->getType(),
+                      (ret = dispatch<VTK_TT, TTK_TT>(
+                         inputScalars, outputSeparators,
+                         *static_cast<TTK_TT *>(triangulation->getData()))));
+
+  if(ret != 0) {
+    return -1;
+  }
+
+  return !ret;
```
vollia: This seems wrong. You want to return 0 (false) if something went wrong. Here this will always return `true`.

rgcmaack:
> 

The use case is to utilize the input scalar field as labels. This means that the same scalar values are grouped together and sourounded by a separating geometry. One way to use this is to compute the ascending segmentation with the path compression module, giving you the minima each vertex converges to in gradient diretion, and create the geometry separating each equivalence class of vertices.
Contours usually have an iso value and first "creates" labels that tell if the vertex has a higher or lower value than the iso value, then creating the separating geometry. However, contours are restricted to two labels only (depending on the implementation).

rgcmaack:
@CharlesGueunet @julien-tierny Adding the MarchingTetrahedraLookupTables.inl to extract the Lookuptables from the code seems to cause issues with some of the checks, as they are not able to locate the file. I only found one occurance of .inl files being used (core/base/common/) and tried to mimic what you did to get the file found (Last two commits). Still, the cheks fail.

Do you have any ideas on how to fix this? It works perfectly fine on my system...

julien-tierny:
you should call your `inl` from your cpp files, not the headers.

julien-tierny:
alternatively, if you want them to be accessible from the headers, you need at least to include them in the list of header files, i.e. in the CMake file of the directory `core/base/marchingTetrahedra` (such that the install install these `inl` as headers)

core/base/marchingTetrahedra/MarchingTetrahedra.h
```diff
@@ -0,0 +1,3603 @@
...
+    unsigned long long *m = cellLabels.data();
+    SimplexId cellIndex = 0;
+    const SimplexId numTets = triangulation.getNumberOfCells();
+
+    for(SimplexId tet = 0; tet < numTets; ++tet) {
+      if(!tetLookupIsMultiLabel[tetCases[tet]]) {
+        continue;
+      }
+
+      SimplexId vertices[4];
```
pierre-guillou: Using zero-initialized `std::array` here instead of a C array should fix the remaining static analysis warning.

core/vtk/ttkMarchingTetrahedra/ttkMarchingTetrahedra.h
```diff
@@ -0,0 +1,92 @@
...
+
+  int FillInputPortInformation(int port, vtkInformation *info) override;
+  int FillOutputPortInformation(int port, vtkInformation *info) override;
+  int RequestData(vtkInformation *request,
+                  vtkInformationVector **inputVector,
+                  vtkInformationVector *outputVector) override;
+
+private:
+  // Output data
+  std::vector<float> output_points{};
```
julien-tierny: for the member variables which are used in `core/base`, please add a `_` suffix to their names such that readers can immediately see we're dealing with a member variable (i.e. `output_points` -> `output_points_`).
this convention is used throughout ttk.

core/base/marchingTetrahedra/MarchingTetrahedra.h
```diff
@@ -0,0 +1,3605 @@
...
+
+#ifdef TTK_ENABLE_OPENMP
+  } else {
+#pragma omp parallel num_threads(threadNumber_)
+    {
+      SimplexId threadTriangles = 0;
+      const int tid = omp_get_thread_num();
+
+#pragma omp for schedule(static)
+      for(SimplexId tet = 0; tet < numTetra; ++tet) {
```
julien-tierny: Same remark as in PR #939.
why have two implementations (1 thread vs multiple threads) when it would be convenient to have only one implementation.

core/base/marchingTetrahedra/MarchingTetrahedra.h
```diff
@@ -0,0 +1,3605 @@
...
+  const unsigned long long *const scalars,
+  const triangulationType &triangulation) const {
+
+  ttk::Timer localTimer;
+
+  this->printMsg("Writing Boundaries", 0, 0, this->threadNumber_,
+                 ttk::debug::LineMode::REPLACE);
+
+#ifdef TTK_ENABLE_OPENMP
+  if(threadNumber_ == 1) {
```
julien-tierny: same remark as above.

core/vtk/ttkMarchingTetrahedra/ttkMarchingTetrahedra.h
julien-tierny: now I understand what you did :)
I believe you can simply declare the actual vector in the base code, with a `_` suffix at the end.
then since the vtk layer inherits from the base code, both will be able to access this vector directly (no need for the pointer passing you're currently doing).
the same goes for the vectors below.

core/base/marchingTetrahedra/MarchingTetrahedra.h
```diff
@@ -0,0 +1,3605 @@
...
+    }
+
+  protected:
+    // Output options
+    SURFACE_MODE SurfaceMode{SURFACE_MODE::SM_SEPARATORS};
+
+  private:
+    // Output data
+    SimplexId *output_numberOfPoints_{};
+    std::vector<float> *output_points_{};
```
julien-tierny: cf my remark in the vtk layer.
(just use vectors and not pointers, these will be readily available from the vtk layer which inherits from the base code)

julien-tierny (changes requested):
alright @rgcmaack 
thanks a lot!
I haven't checked completely but I have a couple remarks (about member variables to share between the base and vtk layers) and again the double-implementation.

can you comment on these remarks?

thanks!

core/base/marchingTetrahedra/MarchingTetrahedra.h
```diff
@@ -0,0 +1,2155 @@
...
+
+  std::vector<unsigned char> tetCases;
+  std::vector<size_t> numberOfTetCases;
+  std::vector<unsigned long long> cScalars;
+
+  tetCases.resize(nC);
+  cScalars.resize(nV);
+
+#ifdef TTK_ENABLE_OPENMP
+  numberOfTetCases.resize(omp_get_max_threads());
```
julien-tierny: I believe you want to use `threadNumber_` instead of `omp_get_max_threads()`

julien-tierny (changes requested):
alright @rgcmaack ,
thanks a lot for the quick fix. this will make the maintenance much easier.

I have one last question (regarding `omp_get_max_threads`, see my review. what do you think?)

core/base/marchingTetrahedra/MarchingTetrahedra.h
rgcmaack: You are totally right. The code now also consistently uses `this->threadNumber_`.

julien-tierny: wonderful.
it seems that your latest commits don't build under the CI.
could you have a look into that please?
thanks!

julien-tierny:
alright, looks good. let's go!

PR #942, opened 2023-05-19
Enable & fix compiler flags

pierre-guillou:
This PR enables 3 new compiler flags:

* `-Wcast-qual` enforces the use of `const_cast` when modifying the constness of a variable (this makes these workarounds easier to detect);
* `-Wmissing-declarations` ensures that all (non-`static`) free functions have been declared in a header file;
* `-Wundef` makes sure that `#ifdef` is used instead of `#if` when checking if a macro-definition has been defined (`#if` only checks if the value of the macro is `!= 0`).

Enjoy,
Pierre

### DIFF
--- a/CMake/CompilerFlags.cmake
+++ b/CMake/CompilerFlags.cmake
@@ -9,6 +9,7 @@ if (NOT MSVC) # GCC and Clang
     -Wtype-limits
     -Wshadow
     -Wvla
+    -Wcast-qual
     )
 
   # performance and debug flags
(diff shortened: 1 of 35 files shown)


julien-tierny:
the CI is failing when testing the screenshots for the latest example (which I have just added).
everything runs fine on my end, but that example requires a lot of memory (about 20 GB if I'm not mistaken).
that may be the issue. do you remember how much memory the VMs ship with?

maybe I could simply skip that pvsm and restrict the test to the python test (which is much simpler than the pvsm).
what do you think?

pierre-guillou:
If I remember correctly, the GitHub VMs have between 7 and 15GB of RAM. There was no chance they would run this state file…

julien-tierny:
Thanks a lot!!!

PR #945, opened 2023-06-07
[MPI] Delete GhostCellPreconditioning and split preconditionDistributed[Cells/Vertices]

eve-le-guillou:
Hi all,

In this PR, we deleted the module `GhostCellPreconditioning`. Indeed, the actions performed by this filter are now triggered automatically when needed, therefore there is no use for this module.

The methods `preconditionDistributed[Cells/Vertices]` are split up in two methods: `preconditionDistributed[Cells/Vertices]` and `preconditionExchangeGhost[Cells/Vertices]`. The first will always be called, and for the second, it may depend on the type of triangulation.
`preconditionExchangeGhost[Cells/Vertices]` should be called when `getGhostVerticesPerOwner()`, `getRemoteGhostVertices()`, `getGhostCellsPerOwner()` and `getRemoteGhostCells()` are called.

The filter `ScalarFieldSmoother` was modified to adapt its preconditioning to these changes.

Thanks for any feedback,

Eve


julien-tierny:
alright, looks good.
thanks a lot!

PR #946, opened 2023-06-19
[MPI] Add MPI support to periodic triangulation

eve-le-guillou:
Hi all,

This PR adds MPI support to the periodic triangulation.
This is in two fold:
- Implementation of a periodic ghost generator, to handle the periodicity at the boundaries.
- Implementation of all the MPI related methods of `ImplicitTriangulation` in `PeriodicImplicitTriangulation`.

The periodic ghost generator will add a layer of ghost vertices at the global boundaries, by copying data at the opposite side of the data set. The ghosts can then be treated like the ghosts generated by Paraview\`s `GhostCellsGenerator`.  The algorithm is implemented in the `ttkPeriodicGhostsGeneration` filter, however the filter will be called automatically when switching to a periodic grid using the `ttkTriangulationManager`. Therefore no additional step needs to be taken to execute this filter with MPI.

Both the meta grid and the local grid for the `PeriodicImplicitTriangulation` are `PeriodicImplicitTriangulation` objects. This is useful when the number of processes of the execution is low, and some opposite global boundaries are present on the same process. In this case, periodic ghosts are not created, and the local and meta grids are used to have the proper periodic behavior.

Some MPI methods in `ImplicitTriangulation` and `PeriodicImplicitTriangulation` were factorized in a new abstract  class called `RegularGridTriangulation`.

Here is an example that creates a Wavelet dataset, makes it periodic and then smoothes the scalar field:

    data = Wavelet(registrationName="wavelet1")

    tTKTriangulationManager1 = TTKTriangulationManager(registrationName='TTKTriangulationManager1', Input=data)
    tTKTriangulationManager1.PeriodicityinAllDimensions = 1
    tTKTriangulationManager1.DataArrays = ['RTData', 'vtkGhostType']

    tTKScalarFieldSmoother1 = TTKScalarFieldSmoother(registrationName='TTKScalarFieldSmoother1', Input=tTKTriangulationManager1)
    tTKScalarFieldSmoother1.ScalarField = ['POINTS', 'RTData']
    tTKScalarFieldSmoother1.IterationNumber = 5
    tTKScalarFieldSmoother1.MaskField = ['POINTS', 'RTData']


core/base/regularGridTriangulation/RegularGridTriangulation.cpp
```diff
@@ -0,0 +1,520 @@
+#include "BaseClass.h"
```
julien-tierny: I don't believe this include is necessary.

core/base/regularGridTriangulation/RegularGridTriangulation.h
```diff
@@ -0,0 +1,105 @@
+/// \ingroup base
+/// \class ttk::RegularGridTriangulation
+/// \author Eve Le Guillou <eve.le-guillou@lip6.fr>
+/// \date June 2023.
+///
+/// \brief RegularGridTriangulation is an abstract subclass of
+/// ttk::AbstractTriangulation that exposes a common MPI API for triangulation
+/// on regular grids. The virtual methods in this class are meant to be
+/// implemented in ttk::ImplicitTriangulation and
+/// ttk::PeriodicImplicitTriangulation. \sa Triangulation
```
julien-tierny: can you please clarify in the documentation that for the moment, this meta-class only holds an interface for MPI-related features? thanks!

julien-tierny: (i.e. maybe this interface will be enlarged in the future)

core/base/periodicImplicitTriangulation/PeriodicImplicitTriangulation.cpp
```diff
@@ -561,6 +560,310 @@ const vector<vector<SimplexId>> *
...
+    static_cast<SimplexId>(
+      std::round((this->origin_[2] - globalBounds[4]) / this->spacing_[2])),
+  };
+
+  this->metaGrid_ = std::make_shared<PeriodicNoPreconditions>();
+  this->metaGrid_->setInputGrid(globalBounds[0], globalBounds[1],
+                                globalBounds[2], this->spacing_[0],
+                                this->spacing_[1], this->spacing_[2],
+                                dimensions[0], dimensions[1], dimensions[2]);
+  this->metaGrid_->preconditionBoundaryVertices();
```
julien-tierny: I don't believe these 3 lines are necessary.
(if they are, I'd recommand to move them over to the actual `preconditionBoundaryVerticesInternal()`)

core/base/triangulation/Triangulation.h
```diff
@@ -2939,9 +2950,19 @@ namespace ttk {
...
+    }
+
+    inline void setIsValid(bool flag) {
+      isValid = flag;
+    }
 #endif // TTK_ENABLE_MPI
 
   protected:
+#ifdef TTK_ENABLE_MPI
+    bool isValid{true};
```
julien-tierny: Could you rename this into `isMPIValid_` (and rename the associated setters/getters)?

core/vtk/ttkIdentifiers/ttkIdentifiers.cpp
```diff
@@ -53,6 +54,19 @@ int ttkIdentifiers::RequestData(vtkInformation *ttkNotUsed(request),
     this->printErr("Triangulation is NULL");
     return 0;
   }
+
+#ifdef TTK_ENABLE_MPI
```
julien-tierny: what if this filter is run without MPI? 
I couldn't find a non-MPI implementation (maybe I didn't see it)

julien-tierny (changes requested):
ok, thanks a lot @eve-le-guillou 
the code for the periodic ghost did not look so straighforward in the end but you managed to make this work. congratulations!
I have a few remarks below. Thanks for addressing them

core/base/regularGridTriangulation/RegularGridTriangulation.cpp
eve-le-guillou: Indeed, it isn't

core/base/periodicImplicitTriangulation/PeriodicImplicitTriangulation.cpp
eve-le-guillou: They are not, so I deleted them.

core/vtk/ttkIdentifiers/ttkIdentifiers.cpp
eve-le-guillou: I added a non-MPI implementation that simply sets the global ids to the local id. Before that, it would simply do nothing.

julien-tierny (approved):

julien-tierny:
hi @eve-le-guillou , thanks a lot for the fixes.

while reading again `ttkIdentifiers.cpp`, I realized that it only generated vertex identifiers. 
I believe we should also generate *cell* identifiers (to make sure we can re-construct unambiguously the global identifiers of simplices of intermediate dimensions in explicit mode).
could you add that up real quick?
thanks!

julien-tierny (changes requested):
hi @eve-le-guillou , thanks a lot for the fixes.

while reading again ttkIdentifiers.cpp, I realized that it only generated vertex identifiers.
I believe we should also generate cell identifiers (to make sure we can re-construct unambiguously the global identifiers of simplices of intermediate dimensions in explicit mode). could you add that up real quick please?
thanks!

julien-tierny:
awesome, thanks. let's go!

julien-tierny (approved):

PR #947, opened 2023-06-28
Feature/vs build

martinfalk:

Two fixes for building the dev version with Visual Studio 2022:
- `ULONG_LONG_MAX` is not part of the C/C++ standard, see https://en.cppreference.com/w/cpp/types/climits
- the nested macros in IntegralLines.cpp caused a compile error with `TTK_ENABLE_MPI` not being enabled


julien-tierny:
awesome, thanks a lot!

PR #956, opened 2023-07-29
[CI] Fix for docker build timeout

atalon-lip6:
The docker script builds ttk containers for two architectures (linux/amd64 and linux/arm64). The two were done sequentially, causing a fail because of Github timeout.
Now the two containers are built in parallel,, hence dividing by two the total time, now under the time limit.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -15,6 +15,9 @@ on:
 jobs:
   docker:
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        arch: [linux/amd64, linux/arm64]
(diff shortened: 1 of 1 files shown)


julien-tierny:
great, thanks!

atalon-lip6:
From what I understand, some docker/ actions are not up to date, and use some deprecated version of Node.js.
I fixed this bug I think, I will create a pull request, maybe it is best to wait for tomorrow to see the results on my fork.

But I think the bug does not always occur (some runs I did worked for three hours). With the future fix, no more warnings, so I think it will solve the issue for good.

PR #957, opened 2023-07-30
[CI] [Bugfix] Upgraded deprecated docker/ actions

atalon-lip6:
The previous PR should have fixed the timelimit problem but there was another issue: the version of Nodejs used by the docker/ actions was no longer supported. This PR uses the last version of these actions.


julien-tierny:
ok, thanks

PR #959, opened 2023-08-06
Misc MPI fixes

julien-tierny:
This PR fixes various little details in the MPI support:
- filters now run without warnings when used with only 1 rank
- the critical points use the legacy backend by default if used with more than 1 rank
- the compact triangulation runs without warnings when used with only 1 rank
- adjusted warnings

@eve-le-guillou thanks for double-checking everything's ok.


eve-le-guillou:
Hi Julien,

Everything looks good to me.

Thank you for this!

julien-tierny:
awesome, thanks a lot for the quick review!

PR #961, opened 2023-08-24
Fix TTK_ENABLE_MPI_TIME creation

eve-le-guillou:
Hi all,

This PR fixes the creation of `TTK_ENABLE_MPI_TIME` and ensures it is automatically visible in `ccmake` or `cmake-gui`.

Now the variable will be automatically created when `TTK_ENABLE_MPI` is set to ON and will be set to ON by default.

Thanks for any feedback,

### DIFF
--- a/CMake/config.cmake
+++ b/CMake/config.cmake
@@ -44,6 +44,7 @@ option(TTK_ENABLE_MPI "Enable MPI support" FALSE)
 if (TTK_ENABLE_MPI)
   find_package(MPI REQUIRED)
   option(TTK_ENABLE_MPI_TIME "Enable time measuring for MPI computation" TRUE)
+  mark_as_advanced(TTK_ENABLE_MPI_TIME)
 endif()
 
 if(TTK_BUILD_PARAVIEW_PLUGINS OR TTK_BUILD_VTK_WRAPPERS)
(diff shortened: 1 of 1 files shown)


CMake/config.cmake
```diff
@@ -43,6 +43,7 @@ mark_as_advanced(TTK_CELL_ARRAY_LAYOUT)
 option(TTK_ENABLE_MPI "Enable MPI support" FALSE)
 if (TTK_ENABLE_MPI)
   find_package(MPI REQUIRED)
+  option(TTK_ENABLE_MPI_TIME "Enable time measuring for MPI computation" TRUE)
```
julien-tierny: thanks @eve-le-guillou 
I'd recommend to mark this option as advanced (see `mark_as_advanced` in the same file for examples). thanks!

julien-tierny (changes requested):
thanks for marking the option as advanced.

julien-tierny:
wonderful, thanks!

julien-tierny (changes requested):
one last thing, I would set this option to FALSE by default.
(you don't want to measure time with barriers by default)

julien-tierny:
perfect, thanks!

PR #964, opened 2023-08-28
New option for distanceMatrixDistortion: do not normalize the result

atalon-lip6:
We propose here two modifications to the distanceMatrixDistortion class. First at a few locations we correct a spelling mistake: distorTion insted of distorsion.

Secondly, we provide the possibility to compute a non-normalized similarity score. Before, we divided everything by the maximum values for each row such that the final result would lie in [0,1]. With the new options enabled, it lies in [-inf, 1].

### DIFF
--- a/paraview/xmls/DistanceMatrixDistortion.xml
+++ b/paraview/xmls/DistanceMatrixDistortion.xml
@@ -2,7 +2,7 @@
 <ServerManagerConfiguration>
   <ProxyGroup name="filters">
     <SourceProxy name="ttkDistanceMatrixDistortion" class="ttkDistanceMatrixDistortion" label="TTK DistanceMatrixDistortion">
-      <Documentation long_help="DistanceMatrixDistortion plugin" short_help="DistanceMatrixDistortion plugin">This plugin, given two distance matrices representing the same points, computes the distortion between the two, according the SIM formula. It also provides, for each point, the distorsion for its own distances to the other points.</Documentation>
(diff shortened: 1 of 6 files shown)


paraview/xmls/DistanceMatrixDistortion.xml
```diff
@@ -24,7 +24,18 @@
           High dimension distance matrix.
         </Documentation>
       </InputProperty>
-      
+
+      <IntVectorProperty
+        name="DoNotNormalize"
+        label="Do not normalize computations"
+        command="SetDoNotNormalize"
+        number_of_elements="1"
```
julien-tierny, line 33: thanks for adding the option `panel_visibility="advanced"` (this will make the checkbox appear only if the advanced view is triggered, i.e. button with the gear icon)

paraview/xmls/DistanceMatrixDistortion.xml
```diff
@@ -161,6 +172,7 @@
       ${DEBUG_WIDGETS}
 
       <PropertyGroup panel_widget="Line" label="Input options">
+        <Property name="DoNotNormalize" />
```
julien-tierny: This is more of an output option I would say.
You may want to create a new, similar `PropertyGoup` with label "Output options" and put this option in it.

julien-tierny (changes requested):
Thanks a lot Alexandre!
Could you address the two minor points of the review? thanks!

atalon-lip6:
> Thanks a lot Alexandre! Could you address the two minor points of the review? thanks!

I made the changes you asked!

julien-tierny:
wonderful, thanks a lot!

PR #965, opened 2023-08-28
output cleanup

JonasLukasczyk:
Hi Julien,
this PR only effects the `ttkUtils::CellVertexFromPoints` method which generates a cell array for a set of input points. Previously, this method only derived a single cell containing all vertices, but now it generates one cell for each point. This has the advantage that one can now apply a threshold filter directly on the output points without first having to derive `ttkIcoSphereFromPoints`.

Please also run the state files from ttk-data. I have some problems with ttk-data even without my changes.


julien-tierny:
right on, thanks a lot!

PR #967, opened 2023-09-04
Export declaration fixes

petersteneteg:
Added missing exports in ttkPersistenceDiagramUtils. Disabled the global use of CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS since it will lead to duplicated exports in the vtk layer that has explicit exports. Enabled WINDOWS_EXPORT_ALL_SYMBOLS only for the base modules since they are missing export declaration.


julien-tierny:
thanks a lot peter!!!

PR #968, opened 2023-09-04
OpenMP for MSVC

petersteneteg:
MSVC has for a long time only supported OpenMP2 but with in the last years they have added a new flag `/openmp:llvm`that enabled more modern OpenMP using the llvm runtime library (Note this is without compiling with clang-cl).

Documentation of what is supported is a bit sparse, these blog posts are the best I have found https://devblogs.microsoft.com/cppblog/category/openmp/

In this PR I enable some support for OpenMP when compiling with MSVC.
This is not meant as a done PR but more of a discussion since I'm not familiar enough with OpenMP to have a clear understanding of all the aspects.

I have renamed some TTK_ENABLE_OPENMP to TTK_ENABLE_OPENMP4 with the goal of having all the TTK_ENABLE_OPENMP ones being supported by the MSVC with the new /openmp:llvm flag.
In total I had to modify 11 files to be able to use the openmp support in MSVC.

In general the MSVC supports OpenMP 3.1 more or less. So it does support atomics and task. But with some restrictions that were removed in later versions. The main uses that I had to disable here were:
* The use of references as arguments to `firstprivate`
* The atomic "modifier" `seq_cst` is not supported 
* The atomic capture block can not use references

Most of the disabled code is in the FTMTree and FTRGraph. I have tried to disable all OpenMP in the affected call graphs but Im not 100% sure I got that right. This probably needs more check by someone familiar with the code.

Some of the disabled OpenMP functionality could probably be enabled again with smallish code changes. But I avoided doing that here.  

I've updated the cmake config with a new option for OpenMP4.
By default TTK_ENABLE_OPENMP and TTK_ENABLE_OPENMP4 is enabled on non MSVC and 
on MSVC TTK_ENABLE_OPENMP is now enabled and TTK_ENABLE_OPENMP4 disabled.
I've also added the needed MSVC flags for using openml:llvm


core/base/integralLines/IntegralLines.h
```diff
@@ -403,9 +403,9 @@ void ttk::IntegralLines::receiveElement(
 
   // Create integral line object on this process
   int threadNum{0};
-#ifdef TTK_ENABLE_OPENMP
+#ifdef TTK_ENABLE_OPENMP4
   threadNum = omp_get_thread_num();
```
julien-tierny, line 407: isn't this call supported by openmp3??

core/base/mergeTreeClustering/MergeTreeBarycenter.h
```diff
@@ -57,7 +57,7 @@ namespace ttk {
       this->setDebugMsgPrefix(
         "MergeTreeBarycenter"); // inherited from Debug: prefix will be printed
                                 // at the beginning of every msg
-#ifdef TTK_ENABLE_OPENMP
+#ifdef TTK_ENABLE_OPENMP4
       omp_set_nested(1);
```
julien-tierny, line 61: same here.

julien-tierny:
Dear Peter,

thanks a lot for this great initiative :)

Overall, I'm OK with the edits (I was surprised to read that a few calls that I believed were basic, such as `omp_get_thread_num` or `omp_set_nested`, were actually not supported by MSVC).
*Edit* I take that back (see above comments)

I have two remarks:

1. I believe that all your edits make sense. However, I'm afraid things are more complicated for FTM  and FTR in the sense that I'm not sure the code will compute correctly if only parts of its OpenMP support are disabled. I believe this is an all-in or all-out kind of strategy that should be followed here. @CharlesGueunet is the original author of these classes. Charles, what's your take?

2. The Windows CI is based on clang, so at the moment, we have no way to check that future commits involving OpenMP will break or not under MSVC. I guess we'd need to add a second Windows port to the CI (to explicitly use MSVC). I'm not sure how easy this would be (I am not familiar at all with MSVC). @pierre-guillou What do you think?

core/base/integralLines/IntegralLines.h
petersteneteg: Yes, that call is supported. But the since some operations later done the call chain, like on line 531 are not supported, I wanted to disable OpenMP for the entire call to be defensive. One is still using OpenMP at the top level, but not at the bottom of a call stack that might lead to errors right? 
But I have not studied the code very carefully here. Maybe openmp could still be enabled for some parts?

julien-tierny: good catch :)

petersteneteg:
Hi Julien
I agree with you. 
About number 1. I have tried to disable openmp in the whole affected call chain, but it is a bit hard do know if I have done that correctly. Which is is why I have just put this up here as a draft. I have compiled and ran it like this in inviwo and from our (limited) testing it seems to be working well. 

So point 1 here is also why I have disabled some of the  `omp_get_thread_num` calls.

About number 2
I'm building ttk using on MSVC using vcpkg and that seems to work well. I could potentially make that in to a github actions if there are interest. Although if you enable the paraview plugins it will also build paraview in vcpkg which is very time consuming.

julien-tierny:
Hi Peter,

1. I think one quick/safe way to proceed would be to replace all the occurrences of `TTK_ENABLE_OPENMP` by `TTK_ENABLE_OPENMP4` in the following directories:
- `core/base/ftmTree`
- `core/base/ftrGraph`
That would effectively force FTM and FTR to be completely sequential (maybe that's what you did in the end).
@CharlesGueunet What do you think?

If you are building the TTK ParaView plugin, one quick way to double check that everything is still working as expected is to run all the examples from https://github.com/topology-tool-kit/ttk-data whose state file includes the string "FTM" and double-check that their output is compatible with the reference output (https://github.com/topology-tool-kit/ttk-data/tree/dev/tests/reference_screenshots). The CI actually performs this comparison automatically for the configured actions.

> build paraview in vcpkg which is very time consuming
2. Have you ever run that in a github action? If it takes less than 6 hours, we should be good to go. We'd need to add the ParaView MSVC build to the actions of the https://github.com/topology-tool-kit/ttk-paraview directory. Once we have that setup, we just need to trigger one build upon each ParaView release. Then we could use that binary (more frequently) when building TTK in the TTK CI.

Best,

CharlesGueunet:
>     I think one quick/safe way to proceed would be to replace all the occurrences of TTK_ENABLE_OPENMP by TTK_ENABLE_OPENMP4 in the following directories:
> 
>     core/base/ftmTree
>     core/base/ftrGraph

Unfortunatly, as long as `seq_cst` is not supported I agree, we need to disable OpenMP on those modules. `seq_cst` date back to 2013, maybe Microsoft will "jump" to the 4.0 standard one day :)

petersteneteg:
I will double check that all OpenMP is disabled in core/base/ftmTree and core/base/ftrGraph.
And then try out the paraview plugin and the state files.
Then I will play with some CI actions, we use vcpkg for inviwo on GitHub Actions and that works so, I think it should be possible. 

It seems that from reading the MSVC blog posts linked above that they plan on eventually supporting all of OpenMP. But they have not promised any timeline. At least the progress has been fairly good since they announced the use of the llvm runtime. So Im at least somewhat hopeful.

julien-tierny:
Hi Peter,

I have tested your branch with `TTK_ENABLE_OPENMP4` set to `FALSE` on my linux system (Gentoo).
I have run the CI tests (btw you can do all of them in batch mode, have a look at `ttk-data/python/run.py`, that test is a little too aggressive, so it's worth running the examples reported as incorrect with the `pvsm` state file, for a visual check). Long story short, I believe that this version of your branch is actually working properly, so no need to replace `TTK_ENABLE_OPENMP` throughout the directories I mentioned. Based on the test examples, I observed no regression, so we should be good.

Can you please mark `TTK_ENABLE_OPENMP4` as `advanced`? i.e.
```
diff --git a/CMake/config.cmake b/CMake/config.cmake
index eec383ce3..00b175e74 100644
--- a/CMake/config.cmake
+++ b/CMake/config.cmake
@@ -252,6 +252,7 @@ if(MSVC)
 else()
   option(TTK_ENABLE_OPENMP "Enable OpenMP support" TRUE)
   option(TTK_ENABLE_OPENMP4 "Enable OpenMP support" TRUE)
+  mark_as_advanced(TTK_ENABLE_OPENMP4)
 endif()
 if(TTK_ENABLE_OPENMP)
   find_package(OpenMP REQUIRED)
```

So for point 1, I believe we're good.
How would you like to proceed from there?
Should I merge this PR and leave the MSVC CI for future work?

petersteneteg:
@julien-tierny 

Sorry for the slow response. From my point of view I would be happy with merging this as is. I will still try and add a CI test, but that can very will be an a different PR.

julien-tierny:
alright, let's merge it :)
thanks a lot for your help peter!!!!

PR #969, opened 2023-09-07
[MPI] Improve process identifier preconditioning performance

eve-le-guillou:
Hi,

This PR uses OpenMP to parallelize the Bounding box computation occuring during the preconditioning for process identifiers for implicit grids (and implicit periodic), for both vertices and cells. This should accelerate the preconditioning step significantly.

The parallelization is done through min and max reductions on the for loop.

The output bounding boxes are identical for 2, 4 and 8 processes, for dimensions of $200^3$, $256^3$ and $512^3$.

This should accelerate the preconditioning step significantly.

Thanks for any feedback,

Eve


julien-tierny:
awesome, thanks!

PR #973, opened 2023-09-29
Merge Tree Computation via ExTreeM

m-s-will:
Hello,

this PR adds a new merge tree algorithm based on our upcoming VIS paper "ExTreeM: Scalable Augmented Merge Tree Computation via Extremum Graphs". We have reported speedups over FTM-Tree of up to two orders of magnitude, however, some of this speedup comes from triangulation optimizations which should also come into baseline TTK. 

The filter does the following:

- check if the ascending and descending segmentation exist, if not compute them via [PathCompression](https://github.com/topology-tool-kit/ttk/pull/939)
- efficiently compute the critical points using the ascending and descending segmentation
- compute the Extremum Graph, connecting the saddles with the maxima reachable by their link
- compute the merge tree on the extremum graph

While the geometry of the output of ExTreeM is identical to FTM-Tree (for split and join tree respectively), they do **not** the strictly equal output. ExTreeM has some output arrays which FTM-Tree does not have, and vice versa. Additionally, ExTreeM only computes the selected merge tree, it currently is not able to merge both trees into the contour tree.

A PR to [ttk-data](https://github.com/topology-tool-kit/ttk-data) is in the works.


julien-tierny:
Thanks a lot @m-s-will.
I'll come back to you soon (today or tomorrow)

julien-tierny:
hi @m-s-will ,
thanks again for your PR.
as discussed with @JonasLukasczyk a few days ago, I believe we should unify the entry point to merge tree features.

we should rename the FTM filter into a "Merge and contour trees" filter which would expose mutliple backends to the user (FTM or ExtreeM). Ideally the switch between the two backends should happen in the base code but we can already  put something together quickly by making the switch happen in the VTK layer. I'll draft a PR for that this week and keep you posted.

cheers,

CharlesGueunet:
hi @julien-tierny,
 would it make sense to generalize at the Reeb Graph level ? The output structure should be compatible and the topological concepts behind are related.

julien-tierny:
@CharlesGueunet, thanks for the suggestion!
I don't know, that's a good call.
what do you guys think?

maybe, we can start by merging FTM and ExtreeM and then see to merge the Reeb graph with that later.

JonasLukasczyk:
Hi everyone. I see several issues here:
1. FTM-tree produces skeleton nodes, arcs, and the segmentation, whereas ExTreeM produces a merge tree (nodes+edges) and a segmentation.
2. FTM-tree assigns an increasing integer as a segmentation id, whereas ExTreeM uses vertex identifiers.
3. ExTreeM  does not support arc-refinement.
4. ExTreeM only computes merge trees, no contour trees.

Making the outputs of the merge tree filters compatible is already difficult, so I would even suggest we have a separate merge and contour tree filter in the vtk-layer as opposed to a combined merge/contour/Reeb Graph filter.

Of course I'm in favor of our design, as you guys are in favor of yours xD. So the question is how the output of the merge tree filter should look like. is there still a good reason to separate the nodes and the arcs in the output? I think this stems primarily from the arc-refinement, but I think we can now safely provide all point data arrays on the node output on the arc output, e.g., if someone wants to render only the critical points of a refined merge tree that user would have to first run a threshold on the arcs based on the point type.

julien-tierny:
hi jonas,
thanks for joining the discussion.
for various reasons, some (at least minimal) uniformization has to happen.
let's discuss this tomorrow.

julien-tierny:
hi Michael, 
I have re-arranged the classes as discussed on Friday (this has been merged with PR #979).
Could you please insert your code into this new class hierarchy? (check the comment `insert ExTreeM vtk execution flow here.` line 307 of `core/vtk/ttkMergeTree/ttkMergeTree.cpp`).
I'd recommend to implement the helper functions (non inherited from `vtkAlgorithm`, e.g. `getMergeTreePoints`, `getMergeTree`) in a separated class (`ttkMergeTreeBase`) if possible (otherwise, simply drop them all in `ttkMergeTree.cpp`).
Thanks!

m-s-will:
Hey Julien, 
when moving the helper functions to ttkMergeTreeBase, I get an "undefined symbols" error for each Triangulationtype, except AbstractTriangulation. When moving it into the ttkMergeTree class it works. I assume it is because ttkMergeTree inherits from ttkAlgorithm while ttkMergeTreeBase does not. I'm currently calling e.g. getMergeTree through ttkTypeMacroT. I'm not that well versed in the Triangulations, do you know what I'm doing wrong?

Michael
Bildschirmfoto 2023-10-10 um 17 20 58

julien-tierny:
at first glance, I'd say nothing (i.e. you don't do anything wrong here).

that's what I meant by "if possible". 
for charles code, I only moved these helper functions where I did not have such errors.
those which triggered errors, I left in `ttkMergeTree`.
not optimal but I guess that will work for the moment.

I guess you can do the same (only keep in `ttkMergeTreeBase` these helper functions which don't trigger errors, if any)

core/base/exTreeM/ExTreeM.h
```diff
@@ -0,0 +1,1743 @@
...
+        );
+        this->printMsg(ttk::debug::Separator::L1); // horizontal '=' separator
+      }
+
+      return 1; // return success
+    }
+
+  }; // ExTreeM class
+
+#else
```
julien-tierny: I'm unsure about this big #ifdef TTK_ENABLE_OPEN.
I'd recommend not to duplicate code, but rather, to put every #pragma omp in between these #ifdefs

Plus, I have the impression that the implementation below (without OpenMP) is not computing anything at all. Am I right?

julien-tierny (changes requested):
hi @m-s-will ,
thanks a lot for this PR.
I read the code and there's one thing I'm not quite sure about, regarding the ExTreeM class when OpenMP is disabled. 
I would recommend to avoid duplicated implementations (in particular, I have the impression that the non-OpenMP one is not computing anything). This can be done by wrapping each "#pragma omp" call in the main implementation with #ifdef.
Do you think could you could do that really quick? 
Thanks for your feedback.

m-s-will:
Hello Julien,
@JonasLukasczyk and my intention was to mostly limit the usage of ExTreeM to situations where OpenMP with high number of parallel threads is available, as that is where most of the performance improvements lie. I can change it such that you can theoretically use it without any parallelism at all, but I would give the user a warning then, recommending the use if OpenMP.

julien-tierny:
sounds good, thanks!
(it's just that I believe code duplicates should be avoided as much as possible)

JonasLukasczyk:
I'm sorry for the late response and I think we should now keep the current version, but I still want to point out that I think this is bad practice. Now we obfuscated the code with imho zero benefit. Who wants to run ExTreeM (an algorithm build for parallelization at scale) without the parallelization backend? If one really wants to compute merge trees without OpenMP we already have backends for that.

julien-tierny:
ok, looks good. thanks a lot!

PR #977, opened 2023-10-04
[MergeTreeDistance] non-matching weight

MatPont:
This PR adds a weight in the ParaView GUI for the deletion and insertion costs for merge tree distances.


julien-tierny:
wonderful, thanks!

PR #980, opened 2023-10-09
[PersistenceDiagramDistance] non-matching weight

MatPont:
This PR adds a weight in the ParaView GUI for the deletion and insertion costs for persistence diagram distances.


julien-tierny:
Thanks a lot!

PR #983, opened 2023-10-16
[CMAKE] Standalone apps and examples now require cmake version 3.21 i…

atalon-lip6:
Updating the minimum required CMake version to 3.21 for the files which required an older version.


julien-tierny:
ok, thanks!

PR #984, opened 2023-10-18
[MPI] Improve exchangeGhostVertices function

eve-le-guillou:
Hi all,

In this Pull Request, the function `exchangeGhostVertices` was modified to improve its performance. 

In the new version, the processes only interact with their neighbors. The communications are done in a non-blocking manner. Using the function `MPI_Waitsome`, the post-processing of communications is performed while the rest of the MPI messages are received.
Preprocessing and postprocessing steps are parallelized using OpenMP.

For datasets of $512^3$ vertices, on 16 processes, this new version is about 10 times faster than the previous one. It shows to scale much better than the previous version, as its execution time barely increases when the number of processes increases.

Similar changes were made to `exchangeGhostCells`.

A new attribute was added to the triangulation, `neighborsToId_`, as it is used at every execution of `exchangeGhostVertices`. It is a map of ints with ranks of processes as keys and their identifier in the `neighbors_` vector as values. Several changes were made in the MPI preconditioning to account for this change.

Thanks for any feedback,


julien-tierny:
can you sync with ttk:dev please?

julien-tierny:
OK, looks great. Thanks a lot!

PR #989, opened 2023-11-13
[CI/MacOS] Bugfix Pythons 3.12 + 3.11

atalon-lip6:
The Github MacOS runner seems to ship Python 3.12 binaries/libraries along with Python 3.11.6 being installed. This cause some problems which we solve here by removing th 3.12 files.


julien-tierny:
awesome, thanks!

PR #992, opened 2023-11-20
Added CritType and RegionType for ExTreeM

m-s-will:
Hello,

this PR adds the CriticalType array to the nodes and the RegionType array to the segmentation of the merge tree. The types follow the implementation of FTMTree for compatibility, see https://github.com/topology-tool-kit/ttk-data/pull/152


julien-tierny:
that's awesome @m-s-will 
thanks for the hot fix!

PR #993, opened 2023-11-27
[ttkMergeTree] Changed criticalType computation for ExTreeM based on treetype

m-s-will:
This PR fixes an oversight in the criticalType computation for splitTrees. Changes between join- and splitTrees were done by inverting the order of the data, the rest of the computation was treeType agnostic, which led to the wrong criticalType (the right one for the inverted data). This fixes it based on the treeType supplied in parameters.


julien-tierny:
awesome, thanks!

PR #994, opened 2023-11-28
  [MPI] Distributed global order array computation

eve-le-guillou:
Hi all, 

In this PR, I implemented another method for the computation of the global order array in a distributed memory context.

Here is how the method works: first, a new vector is created, composed of structs with the following attributes: the scalar value of a vertex, its global identifier and  the rank of its owner. This vector is sorted using the distributed sort psort (see the file psort.h for more details). This means that the struct value will move accross processes, until all the smallest values are located on process 0, then process 1 and so on and so forth. Then the global order of the sorted vertices can be computed, using its local offset (equal to the number of vertices on all the processes of rank lower than the current process) and the local order of the vertex. The global identifiers and the global order of the vertices are then sent back to the owner process. Finally, the global order for the ghosts are exchanged between processes using the method `exchangeGhostVertices`.

This method can become costly in terms of memory when each process owns a few hundred million vertices. To ensure the computation will run, an option `ChunkSize` was introduced. This will force the computation of the global order and the following communications to be performed on segments of the sorted vertices, limiting memory use. This causes a slow down in the execution but ensures the code will work with limited memory. `ChunkSize` is the size of the segment of sorted vertices to be processed.

A new compilation variable was added: `TTK_ENABLE_MPI_RANK_ID_TIME`. When set to true, the type of the variable storing ranks of processes in the algorithm `ArrayPreconditioning` is `integer`, otherwise it is `char`.

The time performance of this method is good: for the Backpack, Wavelet, Elevation, Isabel and Random datasets (resampled to $512^3$), on 16 processes with 24 threads eac    h, the execution takes less than a second, accelerating the computation up to 50 times, with most of the time being spent in the distributed sort as expected. 

When executing on the [AT dataset](https://github.com/topology-tool-kit/ttk-data/blob/dev/at.vti) resampled to $2048^3$, this method takes 13.4 seconds on 64 processes with 24 threads each. The previous method took 21 minutes on 32 processes to compute for the same dataset. 

### Example of use: 

There are to ways to trigger a global order computation: from within a TTK filter, using the usual `GetOrderArray` and setting its argument `getGlobalOrder` to `true`, or by calling the `ArrayPreconditioning` filter within a pipeline, as shown in the snippet below, for the dataset wavelet:

    data = Wavelet(registrationName="wavelet")

    tTKArrayPreconditioning1 = TTKArrayPreconditioning(registrationName='TTKArrayPreconditioning1', Input=data)
    tTKArrayPreconditioning1.GlobalOrderArray = 1
    tTKArrayPreconditioning1.ChunkSize = 1000000000

The MPI documentation has also been updated to show which level of thread support is required (instead of just its value). Two OpenMP statements were added to the `ScalarFieldSmoother` filter for better efficiency.

Thanks for any feedback,
Eve


core/base/abstractTriangulation/AbstractTriangulation.h
```diff
@@ -3853,6 +3861,7 @@ namespace ttk {
     bool hasPreconditionedDistributedVertices_{false};
     bool hasPreconditionedExchangeGhostVertices_{false};
     bool hasPreconditionedGlobalBoundary_{false};
+    bool isOrderArrayGlobal_{false};
```
julien-tierny: I remember we discussed this and we concluded to use this flag but now I'm having second thoughts.
for instance, consider two scalar fields defined on a common triangulation. 
how will things be handled with only one flag? (although there are two fields, with possibly distinct preferences). this is probably something we should discuss next time.

core/base/scalarFieldSmoother/ScalarFieldSmoother.h
```diff
@@ -116,6 +116,9 @@ int ttk::ScalarFieldSmoother::smooth(const triangulationType *triangulation,
   dataType *outputData = (dataType *)outputData_;
   dataType *inputData = (dataType *)inputData_;
   // init the output
+#ifdef TTK_ENABLE_MPI
```
julien-tierny: shouldn't it be `TTK_ENABLE_OPENMP`?

core/base/scalarFieldSmoother/ScalarFieldSmoother.h
```diff
@@ -156,6 +159,9 @@ int ttk::ScalarFieldSmoother::smooth(const triangulationType *triangulation,
 
     if(numberOfIterations) {
       // assign the tmpData back to the output
+#ifdef TTK_ENABLE_MPI
```
julien-tierny: same as above.

core/vtk/ttkAlgorithm/ttkAlgorithm.h
```diff
@@ -113,13 +113,31 @@ class TTKALGORITHM_EXPORT ttkAlgorithm : public vtkAlgorithm,
...
    * generated sorted offset field is then attached to the input
-   * vtkDataset \p inputData.
+   * vtkDataset \p inputData. The argument getGlobalOrder, when set to true,
+   * triggers a computation of a global order array, over all MPI processes.
+   * The triangulation argument is only required when getGlobalOrder is set
+   * to true.
    */
   vtkDataArray *GetOrderArray(vtkDataSet *const inputData,
                               const int scalarArrayIdx,
+                              ttk::Triangulation *triangulation,
```
julien-tierny, line 123: I believe you should provide the default value `nullptr` for the `triangulation` argument.

core/vtk/ttkAlgorithm/ttkAlgorithm.h
```diff
@@ -113,13 +113,31 @@ class TTKALGORITHM_EXPORT ttkAlgorithm : public vtkAlgorithm,
...
+                              const bool getGlobalOrder = false,
                               const int orderArrayIdx = 0,
                               const bool enforceOrderArrayIdx = false);
 
+  /**
+   * Initializes and computes the order array. When using MPI processes,
+   * the order will be local when getGlobalOrder is set to true, and global
```
julien-tierny: to false???

core/vtk/ttkAlgorithm/ttkAlgorithm.h
```diff
@@ -113,13 +113,31 @@ class TTKALGORITHM_EXPORT ttkAlgorithm : public vtkAlgorithm,
...
+   * otherwise. This function is called in GetOrderArray and should not be
+   * called directly in other functions or algorithms.
+   */
+
+  vtkDataArray *ComputeOrderArray(vtkDataSet *const inputData,
+                                  vtkDataArray *scalarArray,
+                                  const int scalarArrayIdx,
+                                  const bool getGlobalOrder,
+                                  vtkDataArray *oldOrderArray,
+                                  ttk::Triangulation *triangulation);
```
julien-tierny, line 153: same things with default value of `triangulation` to `nullptr`.

julien-tierny (changes requested):
hi eve,
thanks a lot for this great PR.
here's a first batch of comments. I'll provide more later.
thanks!

core/vtk/ttkAlgorithm/ttkAlgorithm.h
julien-tierny: Actually, I remember now a previous discussion when we said that parameter should *not* have a default value (like `nullptr`). The reason was that we wanted the developer to systematically call with the triangulation variable, such that, when switching to MPI, things would be taken care of already.

julien-tierny: please ignore that comment then.

core/base/scalarFieldSmoother/ScalarFieldSmoother.h
eve-le-guillou: Indeed, that was a mistake on my part.

core/vtk/ttkAlgorithm/ttkAlgorithm.h
eve-le-guillou: It is corrected.

julien-tierny (commented):
alright, the rest of the code looks good.
you may need to walk me through the logic of `ttkAlgorithm::ComputeOrderArray()` one more time (but we can do this next time we chat). Thanks!

core/base/abstractTriangulation/AbstractTriangulation.h
eve-le-guillou: Indeed, with a single flag, only one preference for all scalar fields is chosen. 

Maybe this can be avoided using a `std::map<string,bool>` with one element per order array, and the name of the order array as key?

julien-tierny: I remember there was a similar issue with the `DiscreteGradient` class. 
You may want to check it out to see how it was handled (I remember something like a map indeed).

eve-le-guillou: This is now solved by switching from a `bool` to a `std::map<const void*, bool>` with the pointer of the scalar field as key and the global order array flag as value.

julien-tierny:
it's all good, thanks!

PR #995, opened 2023-11-28
From c++14 to c++17 standard

atalon-lip6:
This changes is prompted by the use of if constexpr expressions in the near future, unsuported with std=c++14.

### DIFF
--- a/CMake/config.cmake
+++ b/CMake/config.cmake
@@ -1,6 +1,6 @@
 # --- Prerequisites
 
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
 
 # --- Global Options
 


julien-tierny:
preliminary experiments:
- ctBones (upsampled to 512³)
- [DiscreteGradient] Built discrete gradient ..................
- 24 cores
- averages over 5 runs

- c++-14: 36.927s
- c++-17: 38.7794

atalon-lip6:
So 5% more time :-(
I experienced the 10% faster on scalarFieldSmoother, I wonder if it still persists with ctBones.

julien-tierny:
PR #1006 requires migrating to C++17.
I've run again the performance tests and couldn't reproduce the 5% perf loss (on discrete gradient computation, the performances between C++14 and C++17 were pretty much on par).
I've checked visually the example data base and everything's OK.
let's go!

PR #997, opened 2023-12-06
[DimensionReduction] Fixing wrong default value for se

atalon-lip6:
Unlike the other sklearn modules, Spectral Embedding does not accept "auto" as a value for the eigen solver, but None instead.


julien-tierny:
thanks for the fix!

PR #998, opened 2023-12-07
[WIP] [standalone] Dimension reduction standalone module, and a few dimension reduction changes

atalon-lip6:
This PR introduces the dimension reduction standalone command-line module. It also adds some documentations for options of dimension reduction in the xml file. It finally reflects a change in a default options for tsne embedding in the base code.

### DIFF
--- a/standalone/DimensionReduction/main.cpp
+++ b/standalone/DimensionReduction/main.cpp
@@ -196,7 +196,7 @@ int main(int argc, char **argv) {
       "compressed in a dense cloud with few outliers. If the cost function "
       "gets stuck in a bad local minimum increasing the learning rate may "
       "help. Usually: [10.0, 1000.0]",
-      true);
+      true); // TODO allow 'auto'?
     parser.setArgument("tsne_maxIter", &tsne_MaxIteration,
                        "t-sne: Maximum number of iterations for the "
                        "optimization. Should be at least 250.",
--- a/standalone/DimensionReduction/main.cpp
+++ b/standalone/DimensionReduction/main.cpp
@@ -25,7 +25,7 @@ int main(int argc, char **argv) {
   bool listArrays{false};
 
   // Variables for the module options.
-  int backend{0};
+  int backend{2}; // MDS
   bool isInputDistMat{false};
   int nbComp{2}, nbNeighb{5};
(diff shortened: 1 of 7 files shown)


standalone/DimensionReduction/main.cpp
```diff
@@ -0,0 +1,437 @@
...
+      std::string outputFileName = outputPathPrefix + "_port_"
+                                   + std::to_string(i) + ".csv";
+      msg.printMsg("Writing output file `" + outputFileName + "'...");
+      writer->SetInputConnection(output);
+      writer->SetFileName(outputFileName.data());
+      writer->Update();
+    }
+  }
+
+  return 0;
```
julien-tierny, line 436: note for later: there's still a segfault upon the program termination.

standalone/DimensionReduction/main.cpp
```diff
@@ -0,0 +1,437 @@
...
+
+int main(int argc, char **argv) {
+  std::vector<std::string> inputFilePaths;
+  std::vector<std::string> inputArrayNames;
+  std::string outputArrayName{"2d-projection"};
+  std::string outputPathPrefix{"output"};
+  bool listArrays{false};
+
+  // Variables for the module options.
+  int backend{0};
```
julien-tierny: I'd argue to use a different default value. 0 is for Spectral Embedding. I guess MDS would be more standard (check the defaults in the ParaView *xml file).

julien-tierny (changes requested):
thanks a lot
here's a first batch of remarks.

standalone/DimensionReduction/main.cpp
atalon-lip6: Fixed, MDS it is indead.

standalone/DimensionReduction/main.cpp
atalon-lip6: I fixed this by running Py_Finalize in the destruction of dimensionReduction.
In the previous version we would not run it if Python was already initialized, I mimick this behaviour with an additional boolean member of dimensionReduction.

Note: not sure if it is normal, but valgrind says some things initialised by Python are not freed. Yet, Py_Finalized is called.

PR #999, opened 2023-12-12
FTMTree: Remove TTK_ENABLE_FTM_TREE_PROCESS_SPEED

### LINKED ISSUE #960: TTK_ENABLE_FTM_TREE_PROCESS_SPEED set to ON leads to error
**Describe the bug**

The `TTK_ENABLE_FTM_TREE_PROCESS_SPEED` Cmake variable has not been maintained and should be removed.
Setting this variable to `ON`  leads to a compiler error (when a **const** variable is updated)

**To Reproduce**

1. Set `TTK_ENABLE_FTM_TREE_PROCESS_SPEED` to `ON` at configure time.
1. Try to compile

**Expected behavior**

* [] Remove the deprecated variable and associated code OR
* [] Update the code so the process speed works AND add a CI build with this option

pierre-guillou:
This PR should fix #960 by removing the TTK_ENABLE_FTM_TREE_PROCESS_SPEED CMake key and associated code, as suggested by @CharlesGueunet.

Enjoy,
Pierre

### DIFF
--- a/core/base/ftmTree/FTMTree_MT.cpp
+++ b/core/base/ftmTree/FTMTree_MT.cpp
@@ -735,12 +735,6 @@ int FTMTree_MT::printTime(Timer &t,
       st << "-";
     st << s;
 
-#ifdef TTK_ENABLE_FTM_TREE_PROCESS_SPEED
-    const auto nbScalars = scalars_->size;
-    int speed = nbScalars / t.getElapsedTime();
-    st << " at " << speed << " vert/s";
-#endif
-
     this->printMsg(st.str(), 1, t.getElapsedTime(), this->threadNumber_);
   }
   return 1;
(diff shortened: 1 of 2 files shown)


core/base/ftmTree/FTMTree_MT.cpp
```diff
@@ -1006,18 +1000,7 @@ SimplexId FTMTree_MT::trunkCTSegmentation(const vector<SimplexId> &trunkVerts,
 #ifdef TTK_ENABLE_OPENMP
 #pragma omp taskwait
```
julien-tierny, line 1001: do we need to leave that here? @CharlesGueunet

core/base/ftmTree/FTMTree_MT.cpp
```diff
@@ -1078,18 +1054,12 @@ SimplexId FTMTree_MT::trunkSegmentation(const vector<SimplexId> &trunkVerts,
...
-#ifdef TTK_ENABLE_FTM_TREE_PROCESS_SPEED
-#ifdef TTK_ENABLE_OPENMP
-#pragma omp atomic update
-#endif
-      tot += acc;
-#endif
     } // end task
   }
 #ifdef TTK_ENABLE_OPENMP
 #pragma omp taskwait
```
julien-tierny, line 1060: same as above.

julien-tierny (commented):
thanks for the PR Pierre!
I have a question regarding the remaining `taskwait`.
are they really needed?
@CharlesGueunet

core/base/ftmTree/FTMTree_MT.cpp
vollia: a quick test without those lead me to a segfault, so I assume they are needed ^^

We need to ensure all tasks are finished before leaving this context, and I assume the `return` instruction is not enough in a `task` block to ensure so.

julien-tierny:
alright, thanks a lot!

PR #1002, opened 2023-12-19
[OpenMP] Parallelize memory allocation in DiscreteMorseSandwich

eve-le-guillou:
Hi all,

In this PR, the allocation of memory in DiscreteMorseSandwich has been parallelized using OpenMP tasks.

This results in a speed-up of x3.65 on 24 cores (the computation step goes from 2.45s to 0.67s on average).
The results were verified using Pierre Guillou's pdiags_bench benchmark.

The same type of parallelization was performed on the Discrete Gradient's memory allocation step.

Thanks for any feedback.

### DIFF
--- a/core/base/discreteGradient/DiscreteGradient.cpp
+++ b/core/base/discreteGradient/DiscreteGradient.cpp
@@ -26,8 +26,8 @@ void DiscreteGradient::initMemory(const AbstractTriangulation &triangulation) {
   // clear & init gradient memory
 #ifdef TTK_ENABLE_OPENMP
 #pragma omp parallel master num_threads(threadNumber_)
-  {
 #endif
+  {
     for(int i = 0; i < dimensionality_; ++i) {
 #ifdef TTK_ENABLE_OPENMP
 #pragma omp task
@@ -44,9 +44,7 @@ void DiscreteGradient::initMemory(const AbstractTriangulation &triangulation) {
         (*gradient_)[2 * i + 1].resize(numberOfCells[i + 1], -1);
       }
     }
-#ifdef TTK_ENABLE_OPENMP
   }
-#endif
 
   std::vector<std::vector<std::string>> rows{
     {"#Vertices", std::to_string(numberOfCells[0])},
--- a/core/base/discreteGradient/DiscreteGradient.cpp
+++ b/core/base/discreteGradient/DiscreteGradient.cpp
@@ -24,12 +24,31 @@ void DiscreteGradient::initMemory(const AbstractTriangulation &triangulation) {
   }
 
   // clear & init gradient memory
-  for(int i = 0; i < dimensionality_; ++i) {
-    (*gradient_)[2 * i].clear();
-    (*gradient_)[2 * i].resize(numberOfCells[i], -1);
-    (*gradient_)[2 * i + 1].clear();
-    (*gradient_)[2 * i + 1].resize(numberOfCells[i + 1], -1);
+#ifdef TTK_ENABLE_OPENMP
+#pragma omp parallel master num_threads(threadNumber_)
+  {
+#endif
+    for(int i = 0; i < dimensionality_; ++i) {
+#ifdef TTK_ENABLE_OPENMP
+#pragma omp task
+      {
+#endif
+        (*gradient_)[2 * i].clear();
+        (*gradient_)[2 * i].resize(numberOfCells[i], -1);
+#ifdef TTK_ENABLE_OPENMP
+      }
+#pragma omp task
+      {
+#endif
+        (*gradient_)[2 * i + 1].clear();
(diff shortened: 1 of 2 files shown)


core/base/discreteGradient/DiscreteGradient.cpp
```diff
@@ -24,12 +24,31 @@ void DiscreteGradient::initMemory(const AbstractTriangulation &triangulation) {
   }
 
   // clear & init gradient memory
-  for(int i = 0; i < dimensionality_; ++i) {
-    (*gradient_)[2 * i].clear();
-    (*gradient_)[2 * i].resize(numberOfCells[i], -1);
-    (*gradient_)[2 * i + 1].clear();
-    (*gradient_)[2 * i + 1].resize(numberOfCells[i + 1], -1);
+#ifdef TTK_ENABLE_OPENMP
+#pragma omp parallel master num_threads(threadNumber_)
```
julien-tierny, line 28: assuming we're dealing with a 3D dataset (i = 0, i = 1, i = 2), does it mean that the code will now use 6 threads? (or 2 threads 3 times, one dimension at a time; in other words, how will the tasks handle the for loop?)

eve-le-guillou: The code will use `threadNumber_` threads. The master thread will create the tasks for other threads to compute. It is the only thread that will go through the `for` loop. The other threads will only compute the tasks from the task pool. They may be not enough tasks to have one task for one thread, in which case some thread may stay idle.

core/base/discreteGradient/DiscreteGradient.cpp
```diff
@@ -24,12 +24,31 @@ void DiscreteGradient::initMemory(const AbstractTriangulation &triangulation) {
...
+  {
+#endif
+    for(int i = 0; i < dimensionality_; ++i) {
+#ifdef TTK_ENABLE_OPENMP
+#pragma omp task
+      {
+#endif
```
julien-tierny, line 34: I believe you can move this `#endif` before the opening bracket `{`.
then, you won't need the `#ifdef` for the corresponding closing bracket l. 39. (it's OK to have opening and closing brackets)

core/base/discreteGradient/DiscreteGradient.cpp
```diff
@@ -24,12 +24,31 @@ void DiscreteGradient::initMemory(const AbstractTriangulation &triangulation) {
...
+        (*gradient_)[2 * i].clear();
+        (*gradient_)[2 * i].resize(numberOfCells[i], -1);
+#ifdef TTK_ENABLE_OPENMP
+      }
+#pragma omp task
+      {
+#endif
```
julien-tierny, line 41: same thing.

core/base/discreteGradient/DiscreteGradient.cpp
```diff
@@ -24,12 +24,31 @@ void DiscreteGradient::initMemory(const AbstractTriangulation &triangulation) {
...
+        (*gradient_)[2 * i + 1].clear();
+        (*gradient_)[2 * i + 1].resize(numberOfCells[i + 1], -1);
+#ifdef TTK_ENABLE_OPENMP
+      }
+#endif
```
julien-tierny: same thing.

core/base/discreteGradient/DiscreteGradient.cpp
```diff
@@ -24,12 +24,31 @@ void DiscreteGradient::initMemory(const AbstractTriangulation &triangulation) {
...
+    }
+#ifdef TTK_ENABLE_OPENMP
   }
+#endif
```
julien-tierny: same thing.

core/base/discreteMorseSandwich/DiscreteMorseSandwich.h
```diff
@@ -433,25 +433,58 @@ namespace ttk {
...
-      }
-      for(int i = 0; i < dim + 1; ++i) {
-        this->pairedCritCells_[i].resize(
-          this->dg_.getNumberOfCells(i, triangulation), false);
-      }
-      for(int i = 1; i < dim + 1; ++i) {
-        this->critCellsOrder_[i].resize(
-          this->dg_.getNumberOfCells(i, triangulation), -1);
+#ifdef TTK_ENABLE_OPENMP
+#pragma omp parallel master num_threads(threadNumber_)
```
julien-tierny, line 437: I'd put the `#endif` right here. then you don't need the `#ifdef` at the very end.

julien-tierny (changes requested):
hi @eve-le-guillou ,
thanks a lot. can you make the small edits I mentioned?
thanks!

julien-tierny:
wonderful, thanks!

PR #1015, opened 2024-03-12
[PersistenceDiagram] fix vtuToDiagram for 64 bit ids

MatPont:
This PR fixes the `vtuToDiagram` function when TTK is compiled with 64 bit ids.

In the `diagramToVTU` function, the arrays `ttk::PersistencePairIdentifierName` and `ttk::VertexScalarFieldName` are of types `ttkSimplexIdTypeArray` (that is either `vtkIntArray` or `vtkIdTypeArray` if 64 bit ids is respectively disabled or not).
However, currently, the corresponding arrays in the `vtuToDiagram` function are only read with a `SafeDownCast` to `vtkIntArray` returning therefore a `nullptr` when TTK is compiled with 64 bit ids.

It can leads to problems in other methods using persistence diagrams, for instance executing `states/persistenceDiagramClustering.pvsm` does not work with 64 bit ids and the current TTK (except if I have missed something).

### DIFF
--- a/core/vtk/ttkPersistenceDiagram/ttkPersistenceDiagramUtils.cpp
+++ b/core/vtk/ttkPersistenceDiagram/ttkPersistenceDiagramUtils.cpp
@@ -35,7 +35,7 @@ int VTUToDiagram(ttk::DiagramType &diagram,
   }
 
   // cell data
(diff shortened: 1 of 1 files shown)


julien-tierny:
looks good, thanks!

PR #1016, opened 2024-03-12
[MorseSmaleComplex] Return saddle connectors and detecting cycles 

MatPont:
This PR adds methods to detect a cycle in the gradient after reversing the path between a 1-saddle and a 2-saddle paired together.
If the reversal of such a path creates a cycle then the reversal is cancelled.
The cycle is detected either in the whole gradient ("Force loop-free gradient" option in the ParaView GUI) or only in the wall of the 2-saddle.
If the search of cycles is not restricted to the wall of a 2-saddle then the saddle connectors involved in a "gradient fork" are now processed.
The pairs are now processed by order of persistence when reversing saddle connectors, from the lowest to the highest.
A pair is not skipped anymore if one of its children was not reversed.

### DIFF
--- a/core/base/morseSmaleComplex/MorseSmaleComplex.h
+++ b/core/base/morseSmaleComplex/MorseSmaleComplex.h
@@ -1761,11 +1761,13 @@ int ttk::MorseSmaleComplex::returnSaddleConnectors(
     const auto &pair{dms_pairs[i]};
     pairs.emplace_back(std::make_tuple(i, getPersistence(pair)));
   }
-  const auto comparePersistence = [](std::tuple<size_t, dataType> &pair1,
-                                     std::tuple<size_t, dataType> &pair2) {
-    return std::get<1>(pair1) < std::get<1>(pair2);
-  };
-  std::sort(pairs.begin(), pairs.end(), comparePersistence);
+  const auto comparePersistence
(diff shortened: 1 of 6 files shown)


core/base/morseSmaleComplex/MorseSmaleComplex.h
```diff
@@ -1754,50 +1755,72 @@ int ttk::MorseSmaleComplex::returnSaddleConnectors(
...
+  std::vector<std::tuple<size_t, dataType>> pairs;
   for(size_t i = firstSadSadPair; i < dms_pairs.size(); ++i) {
     const auto &pair{dms_pairs[i]};
+    pairs.emplace_back(std::make_tuple(i, getPersistence(pair)));
+  }
+  const auto comparePersistence = [](std::tuple<size_t, dataType> &pair1,
+                                     std::tuple<size_t, dataType> &pair2) {
+    return std::get<1>(pair1) < std::get<1>(pair2);
+  };
+  std::sort(pairs.begin(), pairs.end(), comparePersistence);
```
julien-tierny: please use `TTK_PSORT` instead of `std::sort()`

julien-tierny (changes requested):
thanks a lot!
can you make the quick edit regarding `TTK_PSORT`?

julien-tierny:
awesome, thanks!

PR #1017, opened 2024-03-14
[PersistenceDiagramClustering] avoid computing many distances when 2 diagrams

MatPont:
This PR changes the code of `PersistenceDiagramBarycenter` (classical Auction) in order to avoid computing many distances when 2 diagrams are provided in input.
It also adds the critical coordinates when computing the lower bound (here just for consistency).
It changes the maximum possible value of `delta_lim` (can go higher than 1 without running into problems now).
And allow a custom `delta_lim` in `PersistenceDiagramBarycenter`.


julien-tierny:
awesome, thanks!

PR #1018, opened 2024-03-16
[PersistenceDiagramClustering] fix delta lim from vtk to base layer with classical auction

MatPont:
Explanation in the name of the PR.


julien-tierny:
wonderful, thanks!

PR #1020, opened 2024-04-10
[DiscreteGradient] fix persistence computation as DMS

MatPont:
This PR fixes the persistence computation in the `DiscreteGradient` class to be the same as DMS.

It modifies the `getPersistence` function of this class (**only called** in the `returnSaddleConnectors` function of the `MorseSmaleComplex` class) to use `getCellGreaterVertex` for both critical cells as it is done in DMS (in `executeDiscreteMorseSandwich` function of the `PersistenceDiagram` class).

For example, without this modification, the persistence values of the pairs in `returnSaddleConnectors` will not be the same as the persistence in the pairs of DMS.

Moreover, this PR adds a log regarding the skipped pairs and information about them.


julien-tierny:
Thanks a lot!

PR #1027, opened 2024-05-07
[PointSetToSurface] fix allocate

MatPont:
With the migration to ParaView 5.12 the state file `mergeTreePGA.pvsm` results in a segmentation fault.

The problem seems to be related to the `pointSetToSurface` filter due to a missing `Allocate`.
With this PR the state file should work correctly.


julien-tierny:
wonderful, thanks!

PR #1030, opened 2024-05-24
fix segfault in classical auction

julien-tierny:
self explanatory


julien-tierny:
@MClemot are you OK with this?

PR #1031, opened 2024-05-29
[PersistenceDiagramClustering] fix uninitialized value

MClemot:
This PR aims to fix the uninitialized persistence value of the (0,0) point (used by the diagonal cell) in the PersistenceDiagramClustering TTK filter.


julien-tierny:
thanks a lot!

PR #1037, opened 2024-06-26
[MPI] Fix getCellRank in RegularGrid

eve-le-guillou:
Hi all,

This PR fixes the getCellRank method in the regular grid triangulation.
The former method would assign the rank of a cell by checking if all its vertices are either owned by the process or within the discrete bounding box of the process. This is a problem as a process may own a cell and own none of its vertices.

The new getCellRank method will check if the barycenter of the cell is within the floating point bounding box of the process.

In the ExplicitTriangulation, getCellRank has been also modified, so that the process that owns the cell is the process of lowest rank value.

This was tested by computing the ranks for vertices and cells on 8 processes and checking that all processes return the same rank for a given simplex on a $50^3$ data set.

Thanks for any feedback.


julien-tierny:
looks all good, thanks!

Issue #1036, opened 2024-06-11
nonstandard __int128 usage


petersteneteg:
**Describe the bug**

in #1024 __int128 is used without any guards, __int128 is a gnu extension and not part of c++ at this time hence this breaks compilation on other platforms such as MSVC. 

So far this is the only usage of __int128 in ttk that I can find.

Please revert this or at least guard it.

PR #1039, opened 2024-06-28
[RipsPersistenceDiagram] 128 bits indices option

MClemot:
This PR aims to address issue #1036 and adds an advanced cmake option for using 128 bits indices in Ripser (only available with GCC or clang).


julien-tierny:
Looks great. Thanks for the quick fix.
(I'll wait for the CI to complete)

PR #1040, opened 2024-06-28
trying to add torch to the CI

julien-tierny:
self explanatory.


julien-tierny:
hi @MatPont 
now that the CI supports TTK's build with Torch, some new errors are reported in the MT-WAE.
could you address them please?
thanks!

julien-tierny:
It works, fantastic!

PR #1043, opened 2024-07-04
add Topological Optimization code

MohamedKISSI:
Integration of the code for topological simplification based on persistence diagram optimization.

### DIFF
--- a/core/base/topologicalOptimization/TopologicalOptimization.h
+++ b/core/base/topologicalOptimization/TopologicalOptimization.h
@@ -69,19 +69,6 @@ namespace ttk {
       std::vector<std::vector<SimplexId>> &currentVertex2PairsCurrentDiagram,
       std::vector<int> &vertexInHowManyPairs) const;
 
-    /*
-      Find all neighbors of a vertex i.
-      Variable :
-        -   triangulation : domain triangulation
-        -   i : vertex for which we want to find his neighbors
-        -   neighborsIndices : vector which contains the neighboring vertices of
-      vertex i
-    */
-    template <typename triangulationType>
-    int getNeighborsIndices(triangulationType &triangulation,
-                            const SimplexId &i,
-                            std::vector<SimplexId> &neighborsIndices) const;
-
 /*
   This function allows you to copy the values of a pytorch tensor
   to a vector in an optimized way.
@@ -232,29 +219,6 @@ public:
 
 #endif
 
-/*
-  Find all neighbors of a vertex i.
-  Variable :
-    -   triangulation : domain triangulation
-    -   i : vertex for which we want to find his neighbors
-    -   neighborsIndices : vector which contains the neighboring vertices of
-  vertex i
-*/
-template <typename triangulationType>
-int ttk::TopologicalOptimization::getNeighborsIndices(
-  triangulationType &triangulation,
-  const SimplexId &i,
-  std::vector<SimplexId> &neighborsIndices) const {
-
-  size_t nNeighbors = triangulation->getVertexNeighborNumber(i);
-  ttk::SimplexId neighborId{-1};
-  for(size_t j = 0; j < nNeighbors; j++) {
-    triangulation->getVertexNeighbor(static_cast<SimplexId>(i), j, neighborId);
-    neighborsIndices.push_back(static_cast<SimplexId>(neighborId));
-  }
-
-  return 0;
-}
 
 /*
   This function allows us to retrieve the indices of the critical points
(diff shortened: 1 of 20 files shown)


julien-tierny:
Hi Mohamed,
thanks a lot for this PR.

At this point, the PR doesn't pass the continuous integration (CI), which stopped on a third-party dependency issue (here, torch). As you will read in this documentation (https://github.com/topology-tool-kit/ttk/wiki/Guidelines-for-Developing-a-New-TTK-Module), TTK must build and run even when third-party dependencies haven't been found.

This means that your code should build and run, even if torch is not installed. In particular, if the optimization backend selected by the user was Adam and that Torch was not installed, you should display a warning (`printWrn`) explaining that Adam cannot be used because Torch hasn't been found and that the code will now default to direct gradient descent.

To know in your C++ code if Torch has been found or not, use the preprocessor instruction  `#ifdef TTK_ENABLE_TORCH` (see https://github.com/topology-tool-kit/ttk/blob/dev/core/base/mergeTreeAutoencoder/MergeTreeAutoencoder.h for an example). 

Please proceed to this change and fix any issue until the CI succeeds.
Thanks

julien-tierny:
- Code tested on a simple example (with and without Torch support, with multiple options)
- PV GUI edited
- base code refactored
- console messages edited

core/base/topologicalOptimization/TopologicalOptimization.h
```diff
@@ -0,0 +1,1674 @@
...
+
+    /*
+      Find all neighbors of a vertex i.
+      Variable :
+        -   triangulation : domain triangulation
+        -   i : vertex for which we want to find his neighbors
+        -   neighborsIndices : vector which contains the neighboring vertices of
+      vertex i
+    */
+    template <typename triangulationType>
```
julien-tierny: Please remove that function (see my comments below when it is used).

core/base/topologicalOptimization/TopologicalOptimization.h
```diff
@@ -0,0 +1,1674 @@
...
+    if(not(epoch == 0 || epoch < 0)) {
+#ifdef TTK_ENABLE_OPENMP
+#pragma omp parallel for num_threads(threadNumber_)
+#endif
+      for(size_t indice = 0; indice < listAllIndicesToChange.size(); indice++) {
+        if(listAllIndicesToChange[indice] == 1) {
+          needUpdate[indice] = true;
+          // Find all the neighbors of the vertex
+          std::vector<SimplexId> neighborsIndices;
+          getNeighborsIndices(triangulation, indice, neighborsIndices);
```
julien-tierny: please remove that call.

core/base/topologicalOptimization/TopologicalOptimization.h
```diff
@@ -0,0 +1,1674 @@
...
+
+          for(SimplexId neighborsIndice : neighborsIndices) {
```
julien-tierny: replace that loop by:
```
int vertexNumber = triangulation->getVertexNeighborNumber(index);
for(int i = 0; i < vertexNumber; i++){
   int vertexNeighborId = -1;
   triangulation->getVertexNeighbor(index, i, vertexNeighborId);
   needUpdate[vertexNeighborId] = true;
}
```

core/base/topologicalOptimization/TopologicalOptimization.h
```diff
@@ -0,0 +1,1674 @@
...
+/*
+  Find all neighbors of a vertex i.
+  Variable :
+    -   triangulation : domain triangulation
+    -   i : vertex for which we want to find his neighbors
+    -   neighborsIndices : vector which contains the neighboring vertices of
+  vertex i
+*/
+template <typename triangulationType>
+int ttk::TopologicalOptimization::getNeighborsIndices(
```
julien-tierny: to remove.

core/base/topologicalOptimization/TopologicalOptimization.h
```diff
@@ -0,0 +1,1674 @@
...
+    this->printWrn("Adam unavailable (Torch not found).");
+    this->printWrn("Using direct gradient descent.");
+    enableTorch = false;
+#endif
+  }
+
+  //=======================
+  //    Copy input data
+  //=======================
+  std::vector<double> dataVector(vertexNumber_);
```
julien-tierny, line 1066: Why is the data temporarily converted to double for the computation?
Is that a requirement from Torch?

julien-tierny (commented):
Hi Mohamed, 
thanks again for this PR.
I did quite some modifications to it.

I left a question in my review as well as a request for removing yet another function.
thanks!

MohamedKISSI:
> Hi Mohamed, thanks again for this PR. I did quite some modifications to it.
> 
> I left a question in my review as well as a request for removing yet another function. thanks!

Hi Julien, 

Thank you for your feedback. It's duly noted; I have removed the getNeighborsIndices function and changed the type of the dataVector variable to dataType.
 Thanks

julien-tierny:
Hi Mohamed,

thanks a lot for your edits.
We need to discuss a couple of things though:
- necessity of double format for the data vector (see `kFloat64`), as a first effort, please use double for dataVectors and make all the computation in double.
- check consistency between `kFloat64` and `kDouble`
- multiple unnecessary copies of the ouput data `dataVector` `smoothedScalars`

julien-tierny:
Moreover, the input data must be first normalized in a pre-process, and "de-normalized" in a post-process.

julien-tierny:
Also, please report a status message every ten iterations (and not every iterations). Thanks!
EDIT: make the value 10 editable from the GUI (with 10 being a default value)

julien-tierny:
hi mohamed,
thanks for your quick fix.

it seems that your last commit is breaking things though.

please consider the statefile I sent you in slack, in the TopologicalSimplification node, switch from LTS to TopologicalOptimization and you'll see that the direct gradient descent method generates incorrect results (the adam method looks OK).

julien-tierny:
OK, looks good to me. Thanks a lot!

PR #1044, opened 2024-07-17
[Torch] fix external project linkage

MatPont:
This PR fixes the link with Torch when TTK is used as a library for another project such that the c++ example in TTK.

It fixes a typo `TORCH_FOUND` to `@TORCH_FOUND@`.
Inspired by the ZFP dependency, it sets the `Torch_DIR` variable for external projects as the value used to make TTK, in order to avoid to set this variable in external projects.
It also changes the c++ version of the c++ example from 14 to 17 because 17 is needed for Torch.


julien-tierny:
wonderful, thanks!

PR #1045, opened 2024-07-18
Avoid crash on unstructured grid with invalid cells, when using Contour / Merge tree

CharlesGueunet:
Dear Julien,

This PR fixes the Contour & Merge tree module:
before this commit, invalid unstructured grid lead to crash. The only removed checks are completely trivials and should not impact compute time.


julien-tierny:
Thanks a lot!

PR #1046, opened 2024-08-23
[SignedDistanceField] not expanded box precision

MatPont:
When disabling the `expandBox` parameter of the signed distance field module it is possible for a triangle to lie on the boundary of the box.
However in practice it can results in the triangle not being detected by a ray due to precision problems.
In the end, this can result in the algorithm being unable to detect the interior and exterior of the surface.

To fix this, this PR shift a little bit the rays that are cast on the boundary by 0.1% of the size of the spacing in the corresponding dimensions (i.e the extent of a dimension divided by its resolution).


julien-tierny:
Looks good, thanks!

PR #1048, opened 2024-08-29
Fix Merge Tree (and Persistence Diagram) barycenter precision

MatPont:
This PR fixes the precision problem of the functions updating the scalars of a barycenter of merge trees (or the special case of persistence diagrams).

The problem was that the new scalar values were computed using double variable, but then returned as a template type variable.
This cast can lead close but different values depending on the computers and this small difference quickly snowballs into bigger and more differences as iterations progress.
C-cast or static_cast can not fix this, so the fix consists in computing the new values directly in the template type instead of double.


julien-tierny:
looks good, thanks!

PR #1065, opened 2024-09-18
[MTNN] merge tree neural network refactor

MatPont:
This PR refactor the MergeTreeAutoencoder class into different classes.

First, it separates the layer implementation to its own class.
Then, it creates a MergeTreeNeuralNetwork abstract class (that is derived by the MergeTreeAutoencoder class), serving as a template to create other kind of neural networks processing topological abstractions with the layer mentioned previously.
Finally, it adds some documentation, to all the functions of the new MergeTreeNeuralLayer and MergeTreeNeuralNetwork classes, and some explanation to how to use theses classes in other modules.


julien-tierny:
[NOTES]
TODO
[x] check all mt/pd-ae examples
[x] check code

julien-tierny:
alright, thanks! let's go!

PR #1066, opened 2024-09-18
[MergeTree] saddle-saddle pairs processing

MatPont:
This PR adds the possibility to the merge tree filters to process saddle-saddle pairs when the input corresponds to persistence diagrams.

Until now, only min-saddle and saddle-max pairs could be processed and used at the same time (with a parameter allowing to weight the type of pairs in the final distance).

With this PR, the user can now also choose to process min-saddle and saddle-saddle pairs at the same time, or saddle-saddle and saddle-max pairs at the same time.
Choosing one of the above two options and setting a weight to respectively 0 and 1 allows to process only saddle-saddle pairs.


julien-tierny:
involved examples tested against the latest version of that PR.

julien-tierny:
all good! thanks a lot!

PR #1070, opened 2024-09-30
[MPI] Fix int to SimplexId

eve-le-guillou:
Hi all,

In `edgeToPosition` of both Implicit and Periodic Triangulations, an intermediary variable (`e`) was set and was an integer. This worked only up until a certain number of edges. 
For a grid of $2048$ x $2048$ x $1024$, a problem would arise, as the computed number needed a `long int` type to be expressed and was expressed as an `int`. The result was negative integer positions (impossible in our grids).

Setting the`int` as a `ttk::SimplexId` solves the problem.

For `triangleToPosition`, the variable (here `t`) is already a `ttk::SimplexId`.

Thanks for any feedback.


julien-tierny:
looks good to me, thanks a lot!

PR #1074, opened 2024-12-04
[MPI] fix image spacing implicit triangulation in 1d and 2d cases

SylvainGerbaud:
In case of 1D or 2D vtkImageData, the function getSpacing might return 0 as spacing[1] and/or spacing[2].

Here is a liittle fix to ensure that spacing is at least 1 (it's an integer).

### DIFF
--- a/core/vtk/ttkAlgorithm/ttkTriangulationFactory.cpp
+++ b/core/vtk/ttkAlgorithm/ttkTriangulationFactory.cpp
@@ -185,10 +185,10 @@ RegistryTriangulation
   double spacing[3];
   image->GetSpacing(spacing);
 
-  // 1D
-  if(!spacing[1])
+  // 1D (not tested)
+  if(!spacing[1] && !spacing[2])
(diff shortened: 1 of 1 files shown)


core/vtk/ttkAlgorithm/ttkTriangulationFactory.cpp
```diff
@@ -185,6 +185,12 @@ RegistryTriangulation
   double spacing[3];
   image->GetSpacing(spacing);
 
+  // 1D
+  if(!spacing[1])
```
julien-tierny: `if((!spacing[1])&&(!spacing[2]))` looks safer to me.

julien-tierny:
@SylvainGerbaud thanks for your PR.
can you fix the formatting CI error with clang?
also can you take my remark into consideration?
thanks a lot!

julien-tierny:
alright, wonderful!
may that be the first of many contributions to TTK :)

PR #1079, opened 2024-12-12
fix morse smale complex for multiblock input

thom-dani:

The function setArray is changed so that the output of the ttk layer is correctly copied in the vtkArrays (especially when the input is a vtkMutliBlockDataSet). The points coordinates of the 1-separatrices and 2-separatrices are copied manually. The output of the ttk layerare no longer attributes of the class but local variables in the function dispatch.


julien-tierny:
Tested on Morse-Smale complex states from ttk-data. No regression observed.

julien-tierny:
Alright, looks good to me. Thanks!

PR #1080, opened 2024-12-13
[Ci] Partial updates for recent macOS image upgrades

julien-tierny:
This PR updates the `test` workflow given the recent upgrades of the Github macOS image upgrades.

Notes:
- The `package` workflow is only partially updated (see below)
- The python binaries are simply segfaulting under the new macOS images (the corresponding tests are currently disabled). This is problematic. I will try next to go back to the native python provided with the image (instead of brew's version, this will be test soon). @pierre-guillou do you have any idea of why python is segfaulting right of the bat under macos?

### DIFF
--- a/.github/actions/test-ttk-unix/action.yml
+++ b/.github/actions/test-ttk-unix/action.yml
@@ -33,8 +33,10 @@ runs:
     - name: Test pvpython example
       shell: bash
       run: |
-        cd $GITHUB_WORKSPACE/examples/pvpython
-        pvpython example.py ../data/inputData.vtu
+        if [ "${{ matrix.os }}" != "macos-13" ] && [ "${{ matrix.os }}" != "macos-14" ] && [ "${{ matrix.os }}" != "macos-15" ]; then
+          cd $GITHUB_WORKSPACE/examples/pvpython
+          pvpython example.py ../data/inputData.vtu
+        fi
 
     - name: Test standalone
       shell: bash
--- a/.github/actions/test-ttk-unix/action.yml
+++ b/.github/actions/test-ttk-unix/action.yml
@@ -25,9 +25,9 @@ runs:
     - name: Test Python example
       shell: bash
       run: |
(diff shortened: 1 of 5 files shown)


pierre-guillou:
Hello,
From what I saw, `python` and `pvpython` crash when importing vtk.
This seems related to brew's Python.
Switching back to `setup-python` for building Paraview and TTK seems to work. c.f. https://github.com/pierre-guillou/paraview-ttk/actions/runs/12320285578 and https://github.com/pierre-guillou/ttk/actions/runs/12341335198.
My patches consist in 4 commits available at https://github.com/pierre-guillou/ttk/tree/ci.

julien-tierny:
Thanks a lot for your great feedback @pierre-guillou !!!
I updated ttk-paraview and ttk workflows accordingly and I confirm that now python and pvpython both run fine (without segfaults anymore).

However, I now run into other issues further down the test workflow. In particular, it seems that the run.py script uses Python features which do not seem available in Github's default python:3.12. 
Here's the log, look at the `Run ttk-data Python scripts` step
https://github.com/julien-tierny/ttk/actions/runs/12415698333/job/34662648130

Do you have any clue of what's going on precisely and/or how to fix this?
Thanks a lot for your great feedback!

pierre-guillou:
Yes, I stumbled on the same error last week(-end).
Apparently, the `pvpython` executable is expecting some `vktpython` in `/Library/Frameworks/Python.framework/Versions/3.12/vtkpython`.
I just added a symlink from `pvpython` and it seems to work.

I rebased my branch https://github.com/pierre-guillou/ttk/tree/ci on top of yours. Now it contains only 2 small commits.

Note that when enabling the `.pvsm` CI, there is a lot of noisy `omp_set_nested routine deprecated, please use omp_set_max_active_levels instead.` in TTK filters using OpenMP tasks. 

Also, you might want to skip `persistentGenerators_householdAnalysis.pvsm` since it took 1h30 to complete (the Rips complex takes a long time to render).

Pierre

PR #1082, opened 2024-12-22
[CI] Re-enable python tests for MacOS

julien-tierny:
This PR re-enables pythons tests for the new MacOS images.


julien-tierny:
thanks to @pierre-guillou !

Issue #1081, opened 2024-12-16
TTKTopologicalSimplificationByPersistence not working for certain thresholds


kamakshidasan:
I am trying to apply TTKTopologicalSimplificationByPersistence on a particular dataset. After a particular threshold, the filter seems to hang.

I have attached the dataset exhibiting this behaviour.

```python
vtk_file = 'smooth_fire_clip_tv_2270.vtk'
file_path = os.path.join(input_directory, vtk_file)

vtk_reader = LegacyVTKReader(FileNames=[file_path])
clean_to_grid = CleantoGrid(Input=vtk_reader)
tetrahedralize_data = Tetrahedralize(Input=clean_to_grid)

topological_simplification = TTKTopologicalSimplificationByPersistence(Input=tetrahedralize_data)
topological_simplification.PersistenceThreshold = 3.8
topological_simplification.ThresholdIsAbsolute = 1
topological_simplification.InputArray = ['POINTS', 'Elevation']
topological_simplification.PairType = 'Extremum-Saddle'
```

On thresholds from [0.001, 3.7], the filter seems to work as expected. However, on 3.8 the filter seems to hang. 

The same is seen on Paraview GUI also.

<img width="1496" alt="Screenshot 2024-12-16 at 15 00 32" src="https://github.com/user-attachments/assets/d920cd1b-2ab9-4313-a59d-548a3db15a52" />

OS: MacOS 14.7 (23H124), Paraview 5.12.0 was downloaded directly from Paraview's website. TTK was installed from the plugins section.

I will be unable to build TTK from scratch or change Paraview's version or use another OS at the moment, and hence this issue has not been tested on other environments.

Also, which paper is in the backend of TTKTopologicalSimplificationByPersistence?

[smooth_fire_clip_tv_2270.vtk.zip](https://github.com/user-attachments/files/18147683/smooth_fire_clip_tv_2270.vtk.zip)

JonasLukasczyk:
Thank you for raising this issue. I investigated the problem and the root cause is the topology of the mesh. Specifically, the region that needs to be simplified and runs into an infinite loop belongs to this part of the mesh:

![Image](https://github.com/user-attachments/assets/2ecf38cd-9a50-4e21-99af-88cc7582c9e1)

As you can see that region is connected only via the saddle (a single vertex) to the rest of the domain. 
In short, this causes a problem in the current implementation of the LTS backend. As a quick solution you can do topological simplification before we had the LTS backend:
[...]

PR #1093, opened 2025-02-21
LTS special case fix

JonasLukasczyk:
Hi Julien,
this is a fix for issue #1081. The technical details are documented there. In short, this PR just adds an additional check to the LTS subprocedure that computes the local order of a segment. Before this subprocedure could run into an infinite loop if the segment that needs to be simplified does not have a boundary on which minima can be placed. Now the subprocedure checks if a boundary exists in the first place.

I did not test this yet!

Thanks
Jonas


julien-tierny:
Thanks, looks good to me!
I'll need to double-check the log of the CI for all our test cases (hopefully I'll do that this weekend).

julien-tierny:
Looks all good.
Thanks a lot!!!

PR #1105, opened 2025-05-20
[DimensionReduction] New backend

MClemot:
This PR adds in the DimensionReduction module a new backend "TopologicallyConstrainedDimensionReduction" based on autoencoders with specific topological regularization terms that aim to favor the preservation of cycles.


julien-tierny:
[x] Runs decently without Torch (proper error message, no segfault)

julien-tierny:
[x] Runs decently without CGAL (proper warning, defaulting to ripser)

julien-tierny:
[x] Up to around an order of magnitude speedup with CGAL enabled

julien-tierny:
OK, looks good to me!
waiting for the CI to be finished prior to merging.

PR #1114, opened 2025-08-05
CMake: Add delayed load support

mwestphal:

Add delayed load support to the TTK ParaView plugin, which can be used in ParaView 5.13 and 6.0 to load the TTK plugin only when needed. More information here: https://www.kitware.com/reducing-the-temporal-impact-of-loading-plugins-in-paraview/

### DIFF
--- a/paraview/CMakeLists.txt
+++ b/paraview/CMakeLists.txt
@@ -46,6 +46,10 @@ paraview_plugin_build(
     "${CMAKE_INSTALL_BINDIR}/plugins"
   LIBRARY_DESTINATION
     "${CMAKE_INSTALL_BINDIR}/plugins"
+  PLUGINS_FILE_NAME
+    "ttk.plugins.xml"
+  DELAYED_LOAD
+    TopologyToolKit
   )
 
 install(


paraview/CMakeLists.txt
```diff
@@ -46,11 +46,15 @@ paraview_plugin_build(
...
   )
 
 install(
   TARGETS
     TTKPVTargets
   DESTINATION
     "${TTK_INSTALL_PLUGIN_DIR}"
-  )
\ No newline at end of file
+  )
```
mwestphal, line 60: This change is not needed but a good practices overall, let me know if I should keep the change or not.

mwestphal:
@julien-tierny please review

paraview/CMakeLists.txt
```diff
@@ -46,11 +46,15 @@ paraview_plugin_build(
     "${CMAKE_INSTALL_BINDIR}/plugins"
   LIBRARY_DESTINATION
     "${CMAKE_INSTALL_BINDIR}/plugins"
+  PLUGINS_FILE_NAME
+    "ttk.plugins.xml"
```
julien-tierny, line 50: Does that file get created at runtime?
It doesn't look like it's in the source tree right now.

mwestphal: it is generated at build time and installed indeed.

julien-tierny:
Awesome, thanks a lot!

PR #1121, opened 2025-10-09
[ci] replacing deprecated omp_set_nested

julien-tierny:
Needed to speedup the CI under macos (which currently takes hours due to tons of warnings).

@CharlesGueunet 
@MatPont 
Could you double-check my edits are OK?

@MatPont 
To double-check but it looks like the state file mergeTreeWAEDecoding.pvsm is broken (the top right view is missing the data points here).

Thanks for your feedback!

### DIFF
--- a/core/base/contourForests/ContourForestsTemplate.h
+++ b/core/base/contourForests/ContourForestsTemplate.h
@@ -287,7 +287,7 @@ namespace ttk {
 #endif
 
 #ifdef TTK_ENABLE_OPENMP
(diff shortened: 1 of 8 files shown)


julien-tierny:
Notes: mergeTreeWAEDecoding.pvsm does not seem to be broken with this PR on my laptop.

julien-tierny:
mergeTreeWAEDecoding.pvsm works great on my home desktop as well. 
Merging!

PR #1126, opened 2025-12-10
iostream: use std:: for {cout,cerr,endl}

mathstuf:
VTK has removed the `using` for these symbols.

VTK has removed `<iostream>` and `using std::{cerr,cin,cout}` from its headers. Also did `endl` because they were nearby, but can remove that if wanted.

### DIFF
--- a/core/base/common/Debug.cpp
+++ b/core/base/common/Debug.cpp
@@ -25,7 +25,7 @@ Debug::~Debug() {
   if((lastObject_) && (ttk::goodbyeMsg_)) {
 
     printMsg(
-      "Goodbye :)", debug::Priority::PERFORMANCE, debug::LineMode::NEW, cout);
+      "Goodbye :)", debug::Priority::PERFORMANCE, debug::LineMode::NEW, std::cout);
 
     ttk::goodbyeMsg_ = false;
   }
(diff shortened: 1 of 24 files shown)


julien-tierny:
hi there, thanks for your suggestion.
removing `using` seems like a bigger project than this PR though (TTK uses many STL classes).

mathstuf:
For sure, there are many more instances here. This just aims to handle the cases that were relying on VTK's headers to do it for TTK's code.

pierre-guillou:
The `using` keyword is used in different C++ constructs (c.f. https://en.cppreference.com/w/cpp/keyword/using.html) but as library developpers, what we want to avoid IMO is `using` or `using namespace` directives in the global namespace in public headers.

In the current case, public VTK headers did contain `using std:cout` (and a few other) but they were removed recently, causing potential compilation errors in downstream libraries like TTK (which should be fixed by this very PR).

There is already a rule in `.github/workflows/check.yml` to prevent the use of `using namespace` directives in TTK's header files.

However, `using` in the global namespace might have some legitimacy in TTK, in particular in `core/vtk` where code is not namespaced.

Back to this PR, I think this is a good thing to make explicit calls (with `std::` prefixes) to `std` classes and functions. We might want to remove the remaining `using namespace std` still in `.cpp` files. But that is just coding style.

PR #1141, opened 2026-08-13
Support embree 4

### LINKED ISSUE #1122: Support embree 4.4
Arch Linux has currently [embree 4.4](https://archlinux.org/packages/extra/x86_64/embree/), but TTK requires embree version 3. Can you add support for the current embree version?

lahwaacz:
This adds support for embree 4 - fixes https://github.com/topology-tool-kit/ttk/issues/1122

I've kept support for embree 3 which is pinned in [scripts/docker/pkg/embree.sh](https://github.com/topology-tool-kit/ttk/blob/dev/scripts/docker/pkg/embree.sh). The Docker image uses outdated versions across the entire stack (embree, ospray, openvkl, ispc, paraview, ...) and should be updated, but I will leave this to somebody else.

### DIFF
--- a/CMake/config.cmake
+++ b/CMake/config.cmake
@@ -180,7 +180,11 @@ else()
   message(STATUS "Zlib not found, disabling Zlib support in TTK.")
 endif()
 
-find_package(EMBREE 3.4 QUIET)
+# TTK supports embree 3 (>= 3.4) and embree 4, prefer embree 4 if available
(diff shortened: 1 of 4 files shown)


julien-tierny:
Looks great, thanks a lot!

julien-tierny:
(in principle, the CI stack should be updated to embree 4, but that can wait too)

